PR #32, opened 2020-12-02
optimized build action for dedicated runner

lumir-mrkva:


jak-pan (approved):

jak-pan:
Can merge when CI finishes.

PR #35, opened 2020-12-09
Better setup for testnet | testing

jak-pan:
- sync balances
- add balances to all tokens we have

- bootstrap pools (we probably needs external script to do this)


jak-pan:
This was already incorporated

PR #40, opened 2020-12-14
Math : add/remove liquidity extracted

enthusiastmartin:


jak-pan (commented):
Did we also make some different update here? It looks like it's on the same revision but the cargo.lock changed.

enthusiastmartin:
> Did we also make some different update here? It looks like it's on the same revision but the cargo.lock changed.

hm.. Cargo lock should have changed only forthe math package.  

Now looking at this - it updated all the package where there is a new version. I dont recall what i did exactly to update them. 

So now question is : do we keep it like this or revert to previous cargo.lock ?

jak-pan:
Depends on how possible is to merge this, I think that lock should not have been updated. Maybe there is a mismatched version of something somewhere? (Math package?)

enthusiastmartin:
> Depends on how possible is to merge this, I think that lock should not have been updated. Maybe there is a mismatched version of something somewhere? (Math package?)

I investigated. 

I did cargo update to update the math package - but this obviously updated all the other packages. I should have only ran the update for the math package.

I can do exactly this and revert to the lock file as it is on master but with the latest math package.

enthusiastmartin:
Lock file reverted - only math package updated.

jak-pan (approved):

PR #43, opened 2020-12-17
Extrinsic base weight update

enthusiastmartin:
This was actually not that bad as i originally expected. 

Extrinsic base weight includes payment swap extra weight which is now taken from the swap currency benchmarked weight.


jak-pan:
Nice, it would be interesting to know how many balance transfers we can fit into a block now

jak-pan (approved):

PR #34, opened 2020-12-07
WIP price-fetch from oracle

martinfridrich:
!!!Draft(mvp) price fetch pallet from diadata

PR #47, opened 2020-12-21
Multiple runtime support WIP

enthusiastmartin:
Initial support to easily support multiple runtimes

Within this initial support- there are 2 runtimes prepared - parachain and standalone

However, at the moment - there are moreorless equal. Need to decided what parachain needs and what does not, and it can be easily adjusted. 

By default - parachain node is built.

Also GH workflow might  need to be adjusted


jak-pan:
First thing I'd do is switch defaults right now as we're still in heavy development and parachain will probably not be default way people will test this. 

Then we can merge in stuff from cumulus branch when ready by @lumir-mrkva and split the logic so we DRY.

We can then try to add third staking runtime to aid development of #34. This could be done directly in #34 but it'll need update from this branch or master when this is merged. 

We'll need sync around this to not merge fight each other.

lumir-mrkva:
we'll possibly revisit this in the future

PR #53, opened 2021-01-03
feat: parachain

lumir-mrkva:


jak-pan (approved):

jak-pan (changes requested):
We would need to

runtime/src/lib.rs
```diff
@@ -424,7 +424,8 @@ parameter_types! {
 	pub NativeTokens: BTreeSet<(Vec<u8>, MultiLocation)> = {
 		let mut t = BTreeSet::new();
 		t.insert(("HDT".into(), MultiLocation::X2(Junction::Parent, Junction::Parachain { id: 200})));
```
jak-pan, line 426: We sure this is right?

enthusiastmartin: It does not have to be there. Already removed.

lumir-mrkva:
moved to the https://github.com/galacticcouncil/basilisk-node

PR #60, opened 2021-01-07
Fix benchmark build

enthusiastmartin:
Issue introduced when new payment pallet was introduced. 

Only present when building node for benchmarking.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -57,7 +57,9 @@ sp-transaction-pool = '2.0.0'
 substrate-frame-rpc-system = '2.0.0'
 
 [features]
-default = []
+default = [
+    "frame-benchmarking-cli",
+]
 runtime-benchmarks = ['hack-hydra-dx-runtime/runtime-benchmarks',
     "frame-benchmarking-cli",
 ]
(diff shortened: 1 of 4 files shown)


node/Cargo.toml
```diff
@@ -30,7 +30,7 @@ module-amm-rpc = { path = '../pallets/amm/rpc', version = '2.0.0' }
 
 # Substrate dependencies
 frame-benchmarking = '2.0.0'
-frame-benchmarking-cli = '2.0.0'
+frame-benchmarking-cli = {version = '2.0.0', optional = true }
```
jak-pan, line 33: Is this formatted?

enthusiastmartin: done. looks better, consistent now.

jak-pan: Was it actually formatted or did you fix this manually? 

Maybe we just need to set some settings. I've noticed similar stuff in the code a few times.

jak-pan (approved):

node/Cargo.toml
enthusiastmartin: I did it manually. Is there something which can format these type of files ? i guess there is but apparently i dont use it yet.

PR #61, opened 2021-01-08
Benchmarking and weights update

enthusiastmartin:
Refactored weights in hydra's pallets ( to follow similar approach as it is done in Substrate itself )

Regenerated weights ( on my machine!  ) for all pallets so same hbs template is used. 

Also this is needed for the new auto bench tool - so the structure and templates are correct.


pallets/transaction-multi-payment/src/weights.rs
```diff
@@ -0,0 +1,82 @@
+// This file is part of hack.HydraDX-node.
+
+// Copyright (C) 2021 Parity Technologies (UK) Ltd.
```
jak-pan, line 3: Can we regen this to intergalactic or public domain or something? 😄

enthusiastmartin: Oh..i did not notice..  sure.

enthusiastmartin: done.

jak-pan (approved):

PR #67, opened 2021-01-09
Added config to faucet (max limit, mintable assets and rampage mode)

jak-pan:
Also added tests. Rampage mode should be set to on only when on testnet as everybody will be able to mint any asset with this. (old behavior)

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -1,8 +1,8 @@
 #![allow(clippy::or_fun_call)]
 
 use hack_hydra_dx_runtime::{
-	AccountId, AssetRegistryConfig, AuraConfig, BalancesConfig, GenesisConfig, GrandpaConfig, Signature, SudoConfig,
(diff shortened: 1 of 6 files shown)


node/src/chain_spec.rs
```diff
@@ -189,5 +189,10 @@ fn testnet_genesis(
 				})
 				.collect(),
 		}),
+		pallet_faucet: Some(FaucetConfig {
+			rampage: true,
```
lumir-mrkva, line 193: rampage mint always allowed?

lumir-mrkva (approved):

node/src/chain_spec.rs
lumir-mrkva: i disabled in parachain spec

jak-pan: Yes the idea was to disable it in parachain. We'll probably need different chainspec build functions for multi runtime env anyway as we don't want to pre-mint exogenous tokens on mainnet/roccocco (maybe different ones) and we want to do that on testnets.

PR #81, opened 2021-01-18
Benchmark update

enthusiastmartin:
PR to benchmark all pallets on reference machine.


enthusiastmartin:
Benchmarks done for all pallets on the reference machine as part of this PR.

pallets/amm/src/weights.rs
```diff
@@ -27,15 +27,13 @@
...
-// --pallet
-// amm
-// --extrinsic
-// *
+// --pallet=amm
+// --extrinsic=*
 // --execution=wasm
 // --wasm-execution=compiled
 // --heap-pages=4096
-// --output=./pallets/amm/src/weights.rs
```
jak-pan, line 37: This OK?

pallets/exchange/src/weights.rs
```diff
@@ -62,63 +60,63 @@ pub trait WeightInfo {
...
-		(138_000_000 as Weight)
+		(85_048_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(6 as Weight))
 			.saturating_add(T::DbWeight::get().writes(2 as Weight))
 	}
 	fn on_finalize(t: u32, ) -> Weight {
-		(407_248_000 as Weight)
-			// Standard Error: 394_000
-			.saturating_add((206_463_000 as Weight).saturating_mul(t as Weight))
+		(21_475_000 as Weight)
```
jak-pan, line 77: WTF?

pallets/amm/src/weights.rs
enthusiastmartin: yes. output is a destination where new weights are written. In case of amm - it is exactly there.

pallets/exchange/src/weights.rs
enthusiastmartin: this is actually correct now. because if you look at the other method - known_overhead_for_on_finalize - it is in the same ballpark - around 20_000_000.  Not sure why original benchmark had this huge value.  And similar values were present in lumir's benchmark on this machine ( around 20 )

jak-pan (approved):

PR #87, opened 2021-01-20
Builder's Program Milestone Runtime Review

apopiak:
Main points:
+ You probably want to make sure swap/buy/sell functions are atomic. My recommendation: use ORML `with_transaction_result`.
+ Some of the error cases are not tested.
+ The way you set up your buy/sell distinction leads to lots of complexity that makes it hard to verify the code. The 4-way match you have looks decent, but from my PoV you will want to simplify even more if you can.
+ Could use more comments in the code explaining what's going on. (Doc comments are generally good.)
+ Check your math. More `saturating` and `checked`, less `-` and `+` (or more comments explaining exactly why it's safe).

Some code I changed leads to a failing test. I think my change is what you want, but you will need to decide yourselves.


jak-pan:
As far as the overview goes, I agree with all of the points, some of them are on the radar. New set of eyes greatly helps us fix stuff. Thanks for this 🙏 we're on it.

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -110,13 +110,16 @@ decl_module! {
 		/// Set currency in which transaction fees are paid.
 		/// This is feeless transaction.
 		/// Selected currency must have non-zero balance otherwise is not allowed to be set.
-		#[weight = (<T as Config>::WeightInfo::set_currency(), Pays::No)]
+		// REVIEW: You might want to only make the transaction free if it goes through successfully.
+		// Makes it harder to DoS the chain.
+		#[weight = (<T as Config>::WeightInfo::set_currency())]
```
lumir-mrkva, line 115: we want this to be usable for user with 0 balance in default currency, so the currency it actually has can be set. we need a different way to throttle this.

apopiak: Yeah, you will probably want to find some way to throttle it.

pallets/exchange/src/direct.rs
```diff
@@ -28,6 +28,8 @@ impl<'a, T: Config> DirectTradeData<'a, T> {
 	/// 1. Validate balances
 	/// 2. Calculate fees
 	/// 3. Reserve amounts for each transfer ( including fee transfers )
+	// REVIEW: I would investigate using storage transactions instead of keeping track of all the
+	// transfers here.
```
enthusiastmartin, line 32: We will definitely look at the atomic and use of storage transactions as suggested. However, I feel this would require lot of effort and refactor to change the exchange pallet and algorithm itself. So we will focus on thorough testing now and keeping this in mind when we will come back to improving the exchange pallet and algorithm ( track issue #94 )3

apopiak: I agree that you probably want to look this over and think about it. You can already make `execute_buy` and `execute_sell` transactional, though.

pallets/exchange/src/lib.rs
```diff
@@ -294,6 +314,7 @@ impl<T: Config> Module<T> {
 	/// satisfying  that sum( sell_b_intentions.amount_sell ) <= sell_a_intention.amount_sell
 	///
 	/// Intention A must be valid - that means that it is verified first by validating if it was possible to do AMM trade.
+	// REVIEW: this function does not have tests, but should
```
enthusiastmartin, line 317: Although not directly called in the tests - this function is tested as part of on_finalize calls in the tests

pallets/exchange/src/lib.rs
```diff
@@ -190,8 +201,16 @@ decl_module! {
 					trade_limit: min_bought
 			};
 
+			// REVIEW: This conflicts with your stated invariant `asset_a < asset_b`. Updated the doc comment above.
+			// Also the clone seems unnecessary (can be optimized by
+			// reordering the code).
+			// REVIEW: This is an unbounded append. There is nothing
+			// keeping users from inserting a lot of items here. You will want to
+			// make sure that the maximum amount of items appended in a block is ok.
+			// It's probably generally ok as the item is removed in `on_finalize`.
```
enthusiastmartin, line 210: This should be ok. This extrinsic has certain weight so it can be called only so many times to fill one block. So that would be mean, there can be only so many intentions appended. Or am i missing something ?

pallets/exchange/benchmarking/src/lib.rs
```diff
@@ -119,6 +119,8 @@ benchmarks! {
 	verify {
 	}
 
+	// REVIEW: This is not testing the worst case. The worst case is
+	// when appending to a long list of intentions.
```
enthusiastmartin, line 123: Working in improving the benchamrks now  but I hit an issue which I would appreciate your help.

This sell benchmark  - the worst case scenario is to append an intention to list where is already many intentions.

this is fine. However, I noticed that feeding those intentions is causing a performance issue. I investigated and found out that it is caused by the resizing of this vector. 

And it seems that the capacity is extended by n items - n is rather small ( it looks like 4 but i not sure 100% here ). 
So every n-th intentions the vector is resized. So creating 10_000 intentions is taking therefore lot of time.

I tried to create the initial list  with larger capacity, so the resizing would happen less often. 

And here is my issue - When i stored this vec in the storage, the capacity does not seem to be preserved. And appending next intentions i am back to square one.

Any hint/advice what I can do here? thanks in advance.

pallets/transaction-multi-payment/src/lib.rs
jak-pan: We might want to subtract small amount of the target currency on success to discourage infinite free flipping between the currencies. Not sure if we should calculate the spot price from the pool though as this will bring another dependency on an Oracle. Fail should not get in the block if I'm right.

pallets/exchange/src/lib.rs
apopiak: Yeah, you just want to make sure that the maximum amount of transactions possible in a block is fine, size wise. Related to the benchmarking mentioned in the other comment.

pallets/exchange/benchmarking/src/lib.rs
apopiak: I imagine (would need to check to make 100% sure) that the vec is serialized according to number of elements, not the capacity.
But if it's just about the benchmark you should be able to directly set the storage vec to whatever limit you want to test.
So create a vec with e.g. 1000 intentions and just put it into storage directly. Would that work?

pallets/exchange/src/lib.rs
apopiak: Your call to make. IMO this would make for a good target for property testing ([proptest](https://altsysrq.github.io/proptest-book/intro.html) and friends).

pallets/amm/Cargo.toml
```diff
@@ -34,6 +34,7 @@ primitives = {path = '../../primitives', default-features = false, version = '2.
 # ORML dependencies
 orml-tokens = {default-features = false, version = "0.3.3-dev"}
 orml-traits = {default-features = false, version = "0.3.3-dev"}
+orml-utilities = {default-features = false, version = "0.3.2"}
```
enthusiastmartin, line 37: we should use the same 0.3.3-dev. I believe the reason it is not found when you try to specify 0.3.3-dev - is that it is not "patched" in the Cargo.toml in the top level repo dir. If you add the orml-utilitiles to the patch section of that cargo file with the same revision as the other orml crates- you will be able to use 0.3.3-dev version.

enthusiastmartin: wrong PR.

pallets/amm/src/lib.rs
```diff
@@ -741,27 +758,30 @@ impl<T: Config> AMM<T::AccountId, AssetId, Balance> for Module<T> {
 	fn execute_buy(transfer: &AMMTransfer<T::AccountId, AssetId, Balance>) -> DispatchResult {
 		let pair_account = Self::get_pair_id(&transfer.asset_sell, &transfer.asset_buy);
 
-		if transfer.discount && transfer.discount_amount > 0 {
-			let hdx_asset = T::HDXAssetId::get();
-			T::Currency::withdraw(hdx_asset, &transfer.origin, transfer.discount_amount)?;
-		}
+		// REVIEW: same note as above re atomicity
+		with_transaction_result(|| {
```
enthusiastmartin, line 762: Just curious . Does using with_transaction_result have any performance impact vs previous iplmetation without it?

apopiak: It does come with some performance impact, yes, though it should not be very significant if used judiciously.

jak-pan:
It seems to me that all of the review comments have been addressed as a part of #91 . Right? @enthusiastmartin can we safely close this or are there some open questions still? 

Maybe we need one more issue for transaction-multi-payment antispam measures?

jak-pan:
cc @enthusiastmartin once more.

One more thing. I'd like to keep a list of contributors to the code. We're all moving this forward and this is probably the first time that we close somebody's PR. @apopiak do you have any solution on how you deal with this? I'd like to credit the suggestions and fixes where they should be credited. Maybe keeping a list of contributors at the header of a file?

enthusiastmartin:
> cc @enthusiastmartin once more. ^^

sorry, missed that. 

yes, review comments have been addressed or corresponding issues created. We can close this now.

apopiak:
If this is about crediting my contribution, I'm honored and happy to be listed in a contributors listing of the project, but I don't think it's necessary to do that per-file. That's what git blame is for :shrug: .

jak-pan:
Closing.

> If this is about crediting my contribution, I'm honored and happy to be listed in a contributors listing of the project, but I don't think it's necessary to do that per-file. That's what git blame is for 🤷 .

I hope this works going forward in similar cases dbc48b9f7559b819a200e405c794949a64f186e5 if you have better idea we're open! I searched through all of the current tooling that works with contributions but everything was too much of a hassle or duplication of github contributing system.

Issue #28, opened 2020-11-10
Add minimal pool limit, minimal trade limit and pool existential limit


jak-pan:
These should be setup so that users cannot abuse the system. We should also have minimal pool creation limit / existential limit because the integer math starts to break at some point. Also we don't want to hold very small pools that are blocking other pools because they are very volatile. (something in range of 100-1000 should be ok for math)

Inspiration here:
https://github.com/galacticcouncil/HydraDX-node/pull/27/files

Some progress was also made in #91

RoboRambo:
**1. Minimal Pool Creation Limit (MP):**

The minimal amount of assets needed to create a pool.

One should not be able to create a pool of size 0. Also one should not be able to create a pool with such a small balance that the ability to trade is restricted to a minimum.

Example:
amount of asset_a, asset_b = 1
net worth of assets = $2

If a requested trade exceeds $2, the trade cannot be executed.

Suggestion:
MP e [100, 101, ..., 999, 1000]


**2. Pool Existential Limit (P):**

The minimal amount of assets needed for a pool to exist.

A pool with a balance of 0 should not exist. Also, small pools should not exist (same reasoning as above).

Suggestion:
P e [100, 101, ..., 999, 1000]


**3. Minimal Trade Limit (M):**

The minimal amount of one asset needed to trade.

Suggestion:
M = P / x
x e [10, 100]


Acceptance Criteria:

All of these limits are experimental and:
- should be thought over (especially in combination with the [known limits](https://github.com/galacticcouncil/HydraDX-math/pull/2) of the math crate)
- should be easily tweakable during different testing phases
- prop tests should be expanded/adjusted

jak-pan:
I'd go with 1000 for existential / minimal pool creation limit and 10 as a minimal trade? I think we'll get some more tuning from edge test cases.

enthusiastmartin:
This has been done on Basislisk . All pallets move there. 

When HydraDX repo uses pallets from Basilisk (#287 ) - the limits will come along. 

@jak-pan  Can we therefore close this issue?

jak-pan:
yes

PR #91, opened 2021-01-22
Builder's review refactor

enthusiastmartin:
improvements based on the builder's review

Exchange Pallet
Multi Payment pallet
AMM pallet


pallets/exchange/src/lib.rs
```diff
@@ -84,41 +84,32 @@ decl_event!(
...
 		IntentionResolvedDirectTradeFees(AccountId, AccountId, AssetId, Balance),
 
+		/// Error event - insuficient balance of specified asset
+		/// who, asset, intention type, intention id, error detail
 		InsufficientAssetBalanceEvent(AccountId, AssetId, IntentionType, IntentionID, dispatch::DispatchError),
 
-		//Note: This event can be used instead of AMMSellErrorEvent, AMMBuyErrorEvent
+		/// Intetion Error Event
+		/// who, assets, sell or buy, intention id, error detail
 		IntentionResolveErrorEvent(
```
jak-pan, line 103: We now miss info about the stage at which the intention failed.

enthusiastmartin: Well, not really.  Each error event has 'error detail' which is the actual error which happened. So based on this, it is possible to tell what exactly happened. We just need to figure out how to map this error on the FE side to a meaningful/user friendly info.

primitives/src/asset.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+
+/// Asset Pair representation
+/// ( asset_a, asset_b ) combination where asset_a is meant to be exchanged for asset_b
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Debug, Encode, Decode, Copy, Clone, PartialEq, Eq, Default)]
+pub struct AssetPair {
+	pub asset_in: AssetId,
```
jak-pan, line 13: Maybe we should specify that this is asset pair in case of trades

primitives/src/asset.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub type AssetPairType = AssetPair;
+
+impl AssetPair {
+	pub fn new(asset_in: AssetId, asset_out: AssetId) -> Self {
+		Self { asset_in, asset_out }
+	}
+	/// Return ordered asset tuple (A,B) where A < B
+	/// Used in storage
+	pub fn pair(&self) -> (AssetId, AssetId) {
```
jak-pan, line 25: also this should be called so that we know it's sorted.

enthusiastmartin: renamed

primitives/src/asset.rs
enthusiastmartin: added bit more detailed explanation

pallets/exchange/src/weights.rs
```diff
@@ -60,63 +60,63 @@ pub trait WeightInfo {
...
-		(85_048_000 as Weight)
+		(120_097_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(6 as Weight))
 			.saturating_add(T::DbWeight::get().writes(2 as Weight))
 	}
 	fn on_finalize(t: u32, ) -> Weight {
-		(21_475_000 as Weight)
-			// Standard Error: 36_000
-			.saturating_add((156_991_000 as Weight).saturating_mul(t as Weight))
+		(49_006_000 as Weight)
```
jak-pan, line 77: Is this on the reference machine?

pallets/exchange/benchmarking/src/lib.rs
```diff
@@ -144,33 +160,55 @@ benchmarks! {
...
+		let t in 0 .. MAX_INTENTIONS_IN_BLOCK; // Intention component
 		let caller = funded_account::<T>("caller", 1);
 
 		let asset_a: AssetId = 1;
 		let asset_b: AssetId = 2;
 		let amount : Balance = 100_000_000_000_000;
 
+		// First generate random amounts
+		// This is basically used to generate intentions with different amounts
+		// it is because algorithm does sort the intention by amount, so we need something not sorted./
```
jak-pan, line 193: Is it possible to assume worst case scenario here better than randomness?

pallets/exchange/src/weights.rs
enthusiastmartin: yes. always as it is generated by the benchbot

pallets/exchange/benchmarking/src/lib.rs
enthusiastmartin: Worst case would be here the "unsorted list of amounts". That's what the test case does. It is just creates list of random amounts, this would be presumably unsorted. 
CAn you of other way ?

jak-pan: I feel like random is not the best way to benchmark as it can create different results with different levels of unsortedness. We should generate an arbitrary unsorted sequence that we will input every time? Maybe to external file if big.

pallets/exchange/src/lib.rs
```diff
@@ -36,7 +36,10 @@ mod tests;
 
 /// Intention alias
 type IntentionId<T> = <T as system::Config>::Hash;
-pub type Intention<T> = ExchangeIntention<<T as system::Config>::AccountId, AssetId, Balance, IntentionId<T>>;
+pub type Intention<T> = ExchangeIntention<<T as system::Config>::AccountId, Balance, IntentionId<T>>;
+
+/// Trading limit
+const MIN_TRADING_LIMIT: Balance = 1000;
```
jak-pan, line 42: Extract this to primitive? (we'll probably also use this in other pallets)

pallets/exchange/src/lib.rs
```diff
@@ -287,6 +267,61 @@ decl_module! {
...
+		discount: bool,
+	) -> dispatch::DispatchResult {
+		let intention_count = ExchangeAssetsIntentionCount::get(assets.ordered_pair());
+
+		let intention_id = Self::generate_intention_id(who, intention_count, &assets);
+
+		let intention = Intention::<T> {
+			who: who.clone(),
+			assets,
+			amount_sell: amount_in,
```
jak-pan, line 287: do we want to unify these to in/out?

jak-pan: If we do we should find every use here and change it

pallets/exchange/src/lib.rs
```diff
@@ -449,24 +480,24 @@ impl<T: Config> Resolver<T::AccountId, Intention<T>, Error<T>> for Module<T> {
 		let amm_transfer = match intention.sell_or_buy {
 			IntentionType::SELL => T::AMMPool::validate_sell(
 				&intention.who,
-				intention.asset_sell,
-				intention.asset_buy,
+				intention.assets.asset_in,
+				intention.assets.asset_out,
 				intention.amount_sell,
```
jak-pan, line 485: amount_in?

jak-pan (commented):
Some minor stuff.

pallets/exchange/src/lib.rs
jak-pan: Can also do as a part of #28

pallets/exchange/benchmarking/src/lib.rs
enthusiastmartin: Ok. Make sense. It should be ok - the size is same as max intention.

pallets/exchange/src/lib.rs
enthusiastmartin: Yes, I agree. will do.

runtime/src/lib.rs
```diff
@@ -165,7 +165,7 @@ impl Filter<Call> for BaseFilter {
 			| Call::Offences(_)
 			| Call::AMM(_)
 			| Call::MultiTransactionPayment(_)
-			| Call::Exchange(_) => false,
+			| Call::Exchange(_) => true,
```
jak-pan, line 168: false if we merge now

primitives/src/lib.rs
```diff
@@ -33,15 +34,18 @@ pub type Price = FixedU128;
...
+/// Trading limit
+pub const MIN_TRADING_LIMIT: Balance = 1000;
+
 /// Scaled Unsigned of Balance
 pub type HighPrecisionBalance = U256;
 pub type LowPrecisionBalance = u128;
 
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
+#[derive(Debug, Encode, Decode, Clone, Copy, PartialEq, Eq)]
```
jak-pan, line 48: did we use this in the end?

enthusiastmartin: Yes,  It reduces the clones in the code as you don't need explicitly call it.  So it helps here with readability mainly too. In theory,  may be less expensive to perform.

runtime/src/lib.rs
enthusiastmartin: oh.missed that. done

jak-pan (approved):
LGTM but I suggest another review by one more person since it's so big

pallets/amm/src/lib.rs
```diff
@@ -220,32 +227,27 @@ decl_module! {
...
 
-			// Create pool only if amounts don't overflow
-			let pair_account = Self::get_pair_id(&asset_a, &asset_b);
+			let pair_account = Self::get_pair_id(asset_pair);
 
-			let share_token = match Self::exists(asset_a, asset_b) {
-				true => Self::share_token(&pair_account),
-				false => {
-					let token_name = Self::get_token_name(asset_a, asset_b);
+			let token_name = Self::get_token_name(asset_a, asset_b);
```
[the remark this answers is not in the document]
lumir-mrkva: `ordered_pair` function could be utilized

pallets/amm/src/weights.rs
```diff
@@ -36,79 +37,78 @@
...
-use frame_support::{
-	traits::Get,
-	weights::{constants::RocksDbWeight, Weight},
-};
+use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
 use sp_std::marker::PhantomData;
 
 /// Weight functions needed for amm.
 pub trait WeightInfo {
-	fn create_pool() -> Weight;
```
lumir-mrkva, line 50: sidenote, these generated files should be formatted by rustfmt before committed

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -202,15 +260,33 @@ impl<T: Config> Module<T> {
 
 		// If not native currency, let's buy CORE asset first and then pay with that.
 		if fee_currency != CORE_ASSET_ID {
-			T::AMMPool::buy(&who, CORE_ASSET_ID, fee_currency, fee, 2u128 * fee, false)?;
+			T::AMMPool::buy(&who, AssetPair{asset_out: CORE_ASSET_ID, asset_in: fee_currency}, fee, 2u128 * fee, false)?;
```
lumir-mrkva, line 263: `2u128 * fee` magic swap limit, this should be mby explained in some function and preferably it would be lower than twice the expected amount... mby this should be configurable by user in future

primitives/src/asset.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Asset Pair representation for AMM trades
+/// ( asset_a, asset_b ) combination where asset_a is meant to be exchanged for asset_b
+///
+/// asset_in represents asset coming into the pool
+/// asset_out represents asset coming out of the pool
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Debug, Encode, Decode, Copy, Clone, PartialEq, Eq, Default)]
+pub struct AssetPair {
+	pub asset_in: AssetId,
```
lumir-mrkva, line 16: wouldn't just `in` be enough instead of repeating asset? this leads to awkward paths like `self.intention_b.assets.asset_in`

lumir-mrkva (approved):

pallets/amm/src/weights.rs
enthusiastmartin: yes. i noticed that too.  There is hbs template for this file. Probably would be enough to reformat the template. There is also a bug in this template ( but was fixed on some branches ). I will add separate issue for easy tracking.

pallets/transaction-multi-payment/src/lib.rs
enthusiastmartin: I think configurable option and can be set together with set_currency, so user  can set the limit  alongside chosen currency, is quite a good approach.

@jak-pan what do you think ?

primitives/src/asset.rs
enthusiastmartin: True. But `in` is a keyword in rust. So, any other suggestion for the name ?

pallets/amm/src/lib.rs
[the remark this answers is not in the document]
enthusiastmartin: done.

primitives/src/asset.rs
jak-pan: I'd rather be verbose and clear than short and confusing. What we could have is intention_b.asset_in instead but then we wouldn't have assets as AssetPair (which I don't mind).

pallets/transaction-multi-payment/src/lib.rs
jak-pan: What I am thinking is that we might want to simplify this sooner than expected since this brings so many questions and probably performance implications. What we wanted to do with tx fees is to burn them instead of accumulating. Similar to EIP1559, but this brings some tokenomics questions which we didn't clear yet. 

Right now, we're doing User -> poolSwap (+token,-HDX) -> treasury. We could do just the burn and send on some very basic math really. 

Soo... The solution for me is to make it as simple as possible and create new issue that targets this.

jak-pan: Question is can this be done in normal pools as well?

PR #98, opened 2021-02-05
Claims pallet

green-jay:


pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl<T: Config> Module<T> {
+	fn process_claim(signer: EthereumAddress, dest: T::AccountId) -> DispatchResult {
+		// TODO: Fix multicurrency support and separate checks for not matching addresses and already claimed
+		let balance_due = HDXClaims::get(&signer);
+
+		ensure!(balance_due != Zero::zero(), Error::<T>::NoClaimOrAlreadyClaimed);
+
+		with_transaction_result(|| {
+			HDXClaims::insert(signer, 0);
```
enthusiastmartin, line 106: Nice. Would be cool to come up with a test that tests it somehow.

pallets/claims/src/tests.rs
```diff
@@ -0,0 +1,23 @@
...
+
+#[test]
+fn claiming_works() {
+	new_test_ext().execute_with(|| {
+		// Account 42 signs a msg:
+		// "I hereby claim all my xHDX tokens to wallet:2a00000000000000"
+		let signature = hex!["5b2b46b0162f4b4431f154c4b9fc5ba923690b98b0c2063720799da54cb35a354304102ede62977ba556f0b03e67710522d4b7523547c62fcdc5acea59c99aa41b"];
+
+		assert_noop!(Claims::claim(Origin::signed(142), EcdsaSignature(signature)), Error::<Test>::NoClaimOrAlreadyClaimed);
+		assert_ok!(Claims::claim(Origin::signed(42), EcdsaSignature(signature)));
```
enthusiastmartin, line 20: In the tests - we need to verify the balance after the claim as well.

enthusiastmartin (changes requested):

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+			T::Currency::deposit(CORE_ASSET_ID, &dest, balance_due)?;
+			Ok(())
+		})?;
+
+		Self::deposit_event(RawEvent::HDXClaimed(dest, balance_due));
```
enthusiastmartin, line 111: actually, you can move the event and OK inside the with transaction and use that to return the result directly.

pallets/claims/src/tests.rs
```diff
@@ -0,0 +1,25 @@
...
+	new_test_ext().execute_with(|| {
+		// Alice (account id = 42) signs a msg:
+		// "I hereby claim all my xHDX tokens to wallet:2a00000000000000"
+		let signature = hex!["5b2b46b0162f4b4431f154c4b9fc5ba923690b98b0c2063720799da54cb35a354304102ede62977ba556f0b03e67710522d4b7523547c62fcdc5acea59c99aa41b"];
+
+		assert_eq!(Currency::free_balance(0, &ALICE), 0);
+		assert_noop!(Claims::claim(Origin::signed(BOB), EcdsaSignature(signature)), Error::<Test>::NoClaimOrAlreadyClaimed);
+		assert_ok!(Claims::claim(Origin::signed(ALICE), EcdsaSignature(signature)));
+		assert_eq!(Currency::free_balance(0, &ALICE), 50_000);
+		assert_noop!(Claims::claim(Origin::signed(ALICE), EcdsaSignature(signature)), Error::<Test>::NoClaimOrAlreadyClaimed);
```
enthusiastmartin, line 23: I would also verify that the balance has not changed in any way if error occurs.

pallets/claims/src/benchmarking.rs
```diff
@@ -0,0 +1,34 @@
...
+benchmarks! {
+	_ {}
+
+	claim {
+		let alice_id = hex!["d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"];
+		let caller = T::AccountId::decode(&mut &alice_id[..]).unwrap_or_default();
+		let eth_address = EthereumAddress(hex!["8202c0af5962b750123ce1a9b12e1c30a4973557"]);
+		let signature = hex!["ef9816023122208983c11e596446874df3d400d2f9e380a831206d0e91bfb96d54db352fbd62d3cfa8d8674cf63e6a32052ef3cab038e1e7398eac3d048ed5181c"];
+		HDXClaims::<T>::insert(eth_address, 50_000_000);
+	}: _(RawOrigin::Signed(caller), EcdsaSignature(signature))
```
enthusiastmartin, line 19: Benchmark can have verify section where you can check if the claim has been successful. Similar to test - i would check the balance here to make sure that it is correct.

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+		AccountId = <T as frame_system::Config>::AccountId,
+		Balance = Balance,
+	{
+		HDXClaimed(AccountId, Balance),
+	}
+);
+
+decl_error! {
+	pub enum Error for Module<T: Config> {
+		InvalidEthereumSignature,
```
enthusiastmartin, line 75: This error does not have tests

pallets/claims/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+hex-literal = '0.3.1'
+lazy_static = {features = ['spin_no_std'], version = "1.4.0"}
+
+# Local dependencies
+primitives = {path = '../../primitives', default-features = false, version = '2.0.0'}
+
+# ORML dependencies
+orml-traits = { default-features = false, version = '0.3.3-dev' }
+orml-tokens = { default-features = false, version = '0.3.3-dev' }
+orml-utilities = { default-features = false, version = '0.3.2' }
```
enthusiastmartin, line 41: We should use the same version - 0.3.3-dev. I believe the reason why it is not found when you try to use it - is that it is not patched in the Cargo file in the top level repo dir. If you add the orml-utilies to patch section of that cargo file with the same revision as the other orml crates - you will be able to use 0.3.3-dev version.

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+mod tests;
+
+mod benchmarking;
+
+use weights::WeightInfo;
+pub mod weights;
+
+pub trait Config: frame_system::Config {
+	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
+	type Currency: MultiCurrencyExtended<Self::AccountId, CurrencyId = AssetId, Balance = Balance, Amount = Amount>;
```
enthusiastmartin, line 38: I was just thinking that there is no need to use multi currency in this pallet. Claims only deal with native/core asset. So we could just use the Currency from the frame support. And then use balances pallet to handle native currency. 

just for info: multi currency depends on the native currency "balances" anyway. And if you deposit core asset like you do here in this pallet - the multi currency just calls the balances deposit. So we can get rid of the multi currency layer and call balances directly.

green-jay: I was thinking similarly and originally used the currency type. But when talking about DEX, it just seemed more natural and correct to use the multicurrency type and explicitly define that it is the core asset we want to deposit even though it is not required here. It is also aligned with other pallets. Shall we modify it regardless?

enthusiastmartin: Yes, this makes sense too. But i wonder if using multi currency support here would require also adding additioan tests to make sure that we do not deposit to any other currency ( might sound farfetched?!, not sure). I will discuss this with @jak-pan and we'll see.

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+	type Prefix: Get<&'static [u8]>;
+	type WeightInfo: WeightInfo;
+}
+
+type BalanceOf<T> = <<T as Config>::Currency as MultiCurrency<<T as frame_system::Config>::AccountId>>::Balance;
+
+decl_storage! {
+	trait Store for Module<T: Config> as Claims {
+		HDXClaims get(fn hdxclaims): map hasher(blake2_128_concat) EthereumAddress => BalanceOf<T>;
```
jak-pan, line 47: Pls rename this to ClaimData or something more indicative of what it is.

pallets/claims/src/benchmarking.rs
```diff
@@ -0,0 +1,37 @@
...
+
+benchmarks! {
+	_ {}
+
+	claim {
+		let alice_id = hex!["d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"];
+		let caller = T::AccountId::decode(&mut &alice_id[..]).unwrap_or_default();
+		let eth_address = EthereumAddress(hex!["8202c0af5962b750123ce1a9b12e1c30a4973557"]);
+		let signature = hex!["ef9816023122208983c11e596446874df3d400d2f9e380a831206d0e91bfb96d54db352fbd62d3cfa8d8674cf63e6a32052ef3cab038e1e7398eac3d048ed5181c"];
+		HDXClaims::<T>::insert(eth_address, 1_000_000_000_000_000_000);
```
jak-pan, line 18: Try to not use HDX as we might want to open-source this as a generic pallet.

pallets/claims/src/lib.rs
green-jay: IMHO, the "Data" does not provide more clarity but I can refactor the storage name and remove HDX from it. "Claims" as the name of storage is used for multiple projects including Polkadot, so I would stick with that.

pallets/claims/src/lib.rs
jak-pan: Your call. 

We can simplify to be quick and safe. 
If we actually do this in a generic way. We can probably pr this to orml? It would need more testing tho.

First approach is more general. Second approach is more advanced and we're limiting users so that they need to use orml. 

Q to think about: Can this be a configurable trait? Can we do without id if no asset_id is present and have it if it is present?

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+	pub struct Module<T: Config> for enum Call where origin: T::Origin {
+		type Error = Error<T>;
+
+		fn deposit_event() = default;
+
+		/// The Prefix that is used in signed Ethereum messages for this network
+		const Prefix: &[u8] = T::Prefix::get();
+
+		#[weight = (<T as Config>::WeightInfo::claim(), DispatchClass::Normal, Pays::No)]
+		fn claim(origin, ethereum_signature: EcdsaSignature)  {
```
jak-pan, line 90: https://github.com/paritytech/substrate/blob/171ad2322d7561934b82f0b7e41c58d1710f8c24/primitives/runtime/src/traits.rs#L777

From: Alexander Popiak
Answering the questions about transaction propagation:
Txs that return error when passed to validate_transaction here are not propagated.

jak-pan: cc @enthusiastmartin

runtime/src/lib.rs
```diff
@@ -345,6 +346,17 @@ impl pallet_amm::Config for Runtime {
 	type GetExchangeFee = ExchangeFee;
 }
 
+parameter_types! {
+	pub Prefix: &'static [u8] = b"I hereby claim all my xHDX tokens to wallet:";
```
jak-pan, line 350: ClaimMessagePrefix

pallets/claims/src/mock.rs
```diff
@@ -0,0 +1,129 @@
...
+	type CurrencyId = AssetId;
+	type WeightInfo = ();
+	type ExistentialDeposits = ExistentialDeposits;
+	type OnDust = ();
+}
+
+pub type Currency = orml_tokens::Module<Test>;
+
+parameter_types! {
+	pub Prefix: &'static [u8] = b"I hereby claim all my xHDX tokens to wallet:";
```
jak-pan, line 68: "HDX tokens to wallet: "?

green-jay: The whitespace could do more harm than good. I agree with change to HDX.

pallets/claims/src/lib.rs
green-jay: As a fan of the Occam's razor, I would choose quick and safe. I should be ready with a commit tomorrow.

pallets/claims/src/lib.rs
```diff
@@ -37,6 +36,7 @@ pub trait Config: frame_system::Config {
 	type Currency: Currency<Self::AccountId>;
 	type Prefix: Get<&'static [u8]>;
 	type WeightInfo: WeightInfo;
+	type IntoBalance: From<u128> + Into<<Self::Currency as Currency<<Self as frame_system::Config>::AccountId>>::Balance>;
```
enthusiastmartin: I would suggest to use like this :

type CurrencyBalance: Into<BalanceOf<Self>> + From<Balance>;

and order it so Currency and CurrencyBalance is together , and make a comment that this is the currency used in Currency and needs to support conversion to Balance (as we use it directly for the actual data )

pallets/claims/src/lib.rs
green-jay: can be resolved or still a concern?

green-jay: @jak-pan @enthusiastmartin

pallets/claims/src/mock.rs
lumir-mrkva: any specific reason not to add space after : ? it looks like typo when you are signing

pallets/claims/src/mock.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl ExtBuilder {
+	// builds genesis config
+	pub fn build(self) -> sp_io::TestExternalities {
+		let mut t = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
+
+		GenesisConfig::<Test> {
+			claims: vec![(
+				EthereumAddress([
+					130, 2, 192, 175, 89, 98, 183, 80, 18, 60, 225, 169, 177, 46, 28, 48, 164, 151, 53, 87,
```
lumir-mrkva, line 93: please add hex representation and attach private key of the test address

lumir-mrkva (approved):

node/src/chain_spec.rs
```diff
@@ -468,5 +471,35 @@ fn lerna_genesis(
 		}),
 		pallet_elections_phragmen: Some(ElectionsConfig { members: vec![] }),
 		pallet_collective_Instance1: Some(CouncilConfig::default()),
+		pallet_claims: Some(ClaimsConfig { claims: vec![] }),
 	}
 }
+
+fn create_testnet_claims() -> Vec<(EthereumAddress, Balance)> {
```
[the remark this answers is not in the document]
Roznovjak: I am not sure what you mean. Claim balances are tied to eth addresses, not to our accounts. Do you mean to create claim balances and sign messages for all `endowed_accounts`?

pallets/claims/src/mock.rs
green-jay: It is a string that user will copy&paste. If e.g. was copied as two parts, prefix and later address, the whitespace could cause some confusion. It is imho more visible and clear not to have it than have it. Comes down to personal preference I guess. Don't mind either way

pallets/claims/src/mock.rs
green-jay: done

node/src/chain_spec.rs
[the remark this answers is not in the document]
lumir-mrkva: ah, sorry for the confusion, you are correct. is just important to have claims pre-generated to addresses for local dev chain with known private keys to be usable

pallets/claims/src/lib.rs
green-jay: txs validated by SignedExtension

node/src/chain_spec.rs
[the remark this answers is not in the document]
green-jay: test seed added

pallets/claims/src/migration.rs
```diff
@@ -0,0 +1,61 @@
+use super::*;
+use hex::FromHex;
+use primitives::Balance;
+
+pub fn migrate_to_v2<T: Config>(claims_data: &[(&'static str, Balance)]) -> frame_support::weights::Weight {
```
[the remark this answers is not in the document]
green-jay: agreed, changed to import_initial_claims

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+	}
+}
+
+decl_event!(
+	pub enum Event<T>
+	where
+		AccountId = <T as frame_system::Config>::AccountId,
+		Balance = BalanceOf<T>,
+	{
+		Claimed(AccountId, Balance),
```
lumir-mrkva, line 77: Events shouldn't be in past tense `Claimed` -> `Claim` and this particular event should contain ETH address from which was this balance claimed.

lumir-mrkva (changes requested):

pallets/claims/src/lib.rs
lumir-mrkva: Currently, the same information will be laso emitted in the `Endowed` event so the ETH claim address is the most useful information transmitted by this event.

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+
+mod benchmarking;
+
+use weights::WeightInfo;
+
+pub mod weights;
+
+pub trait Config: frame_system::Config {
+	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
+	type Prefix: Get<&'static [u8]>;
```
lumir-mrkva, line 43: is there any reason for the prefix being configurable?

pallets/claims/src/lib.rs
enthusiastmartin: handled by the signed extension to validate claim.

pallets/claims/src/lib.rs
green-jay: In substrate recipe describing events there is the present tense so we can stick with that.
ETH address added

pallets/claims/src/lib.rs
green-jay: yes, it was designed bearing in mind we might want to publish the claims as a standalone pallet

jak-pan: It can be used by anybody else then as it's pretty generic

node/src/chain_spec.rs
```diff
@@ -468,5 +471,35 @@ fn lerna_genesis(
...
+	let mut claims = Vec::<(EthereumAddress, Balance)>::new();
+
+	// Alice's claim
+	// Signature: 0xbcae7d4f96f71cf974c173ae936a1a79083af7f76232efbf8a568b7f990eceed73c2465bba769de959b7f6ac5690162b61eb90949901464d0fa158a83022a0741c
+	// Message: "I hereby claim all my HDX tokens to wallet:d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"
+	let claim_address_1 = (
+		EthereumAddress(hex!["8202c0af5962b750123ce1a9b12e1c30a4973557"]),
```
jak-pan, line 485: Add private key to comment.

node/src/chain_spec.rs
```diff
@@ -468,5 +471,35 @@ fn lerna_genesis(
...
+		50_000_000_000_000_000,
+	);
+
+	// Another claim
+	let claim_address_2 = (
```
jak-pan, line 490: Remove these before merging

pallets/claims/src/claims_data.rs
```diff
@@ -0,0 +1,10 @@
+use primitives::Balance;
+use sp_std::vec;
+lazy_static::lazy_static! {
+pub static ref CLAIMS_DATA: vec::Vec<(&'static str, Balance)> = vec![
+	("0x8202c0af5962b750123ce1a9b12e1c30a4973557", 555),
```
jak-pan, line 5: We'll replace this with real data right?

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+use sp_runtime::traits::Zero;
+use sp_std::prelude::*;
+use sp_std::vec::Vec;
+
+use frame_support::sp_runtime::traits::{DispatchInfoOf, SignedExtension};
+use frame_support::sp_runtime::transaction_validity::{
+	InvalidTransaction, TransactionValidity, TransactionValidityError, ValidTransaction,
+};
+use frame_support::traits::IsSubType;
+use sp_std::marker::PhantomData;
```
jak-pan, line 21: Consolidate imports. Got multiple lines of the same package. You can use the ::{} syntax

jak-pan: Also reorder pls

pallets/claims/src/claims_data.rs
green-jay: yes, if those can be published (otherwise we could restrict file permissions on github or other solution in mind?)

pallets/claims/src/migration.rs
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::mock::Test;
+
+	#[test]
+	fn data_migration_should_work() {
```
jak-pan, line 33: We need more tests.

- negative balances
- strings
- floats
- duplicated address
- zero balance
- overflowed balance

Also we don't handle these in migration and we probably should.

pallets/claims/src/claims_data.rs
jak-pan: Of course they need to be published :)

jak-pan (changes requested):
Overall looks very nice just some tidying up and maybe a few more migration tests and we're good

node/src/chain_spec.rs
green-jay: Richard just added some additional info like private keys to claim2 & claim3, remove anyway?

pallets/claims/src/migration.rs
Roznovjak: I think it's too late to do any tests at runtime (in this case, since data migration is a one-time operation handled by responsible devs/chain maintainers).

node/src/chain_spec.rs
Roznovjak: The question is What data do we want to use for claims in local testnets?

jak-pan: Ok if we have priv keys we're good. But add some reasonable amounts to them then. 
0.001, 1, and 1000 * 10^12 tokens should be good?

Roznovjak: Is 1 equal to 10^12 ?

pallets/claims/src/migration.rs
jak-pan: How do we generate the data into .rs file? @lumir-mrkva

node/src/chain_spec.rs
jak-pan: Yes.

Here:
https://github.com/galacticcouncil/HydraDX-node/blob/496178440b9ed980d58ab07cecfc3c0432a7e877/runtime/src/constants.rs#L4

And here:
https://github.com/galacticcouncil/HydraDX-node/blob/fadfe28ff04abf504caa0adea475d2a3e04c82d8/node/src/chain_spec.rs#L87

pallets/claims/src/migration.rs
lumir-mrkva: i agree with @Roznovjak, migration doesn't need to be tested so thoroughly, besides we can test it with real data that we'll do real migration with

jak-pan: OK but then we need to definitely test the source script that generates it.

Roznovjak: We can create public subrepo for the script, or add the script to the `claims pallet` folder (or some `script` folder?).

green-jay: Will be resolved separately [here](https://github.com/galacticcouncil/xhdx) I suppose?

node/src/chain_spec.rs
```diff
@@ -485,7 +485,7 @@ fn create_testnet_claims() -> Vec<(EthereumAddress, Balance)> {
 		// Test seed: "image stomach entry drink rice hen abstract moment nature broken gadget flash"
 		// private key (m/44'/60'/0'/0/0) : 0xdd75dd5f4a9e964d1c4cc929768947859a98ae2c08100744878a4b6b6d853cc0
 		EthereumAddress(hex!["8202C0aF5962B750123CE1A9B12e1C30A4973557"]),
```
jak-pan, line 487: Will this actually work? (upper/lowercase)

Roznovjak: Tests pass. AFAIK same wallets use capitalization to compute a checksum and validate an address. Hex conversion algorithms should be clever enough to handle both lower and uppercase input. ('a' and 'A' are the same in hex).

jak-pan: We should probably test this with metamask as I know that there were some problems with this and also we're lowercasing addresses in UI. I'll do that when I have time.

Roznovjak: I've generated all signatures using mycrypto + metamask. It's also easy to convert 3 public keys to lowercase and update the source code.

lumir-mrkva (approved):

node/src/chain_spec.rs
jak-pan: It looks like it works anyway so all good.

pallets/claims/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+		+ Into<<Self::Currency as Currency<<Self as frame_system::Config>::AccountId>>::Balance>;
+}
+
+pub type BalanceOf<T> = <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
+
+decl_storage! {
+	trait Store for Module<T: Config> as Claims {
+		Claims get(fn claims): map hasher(blake2_128_concat) EthereumAddress => BalanceOf<T>;
+
+		PalletVersion: StorageVersion = StorageVersion::V1EmptyBalances;
```
lumir-mrkva, line 49: consider using common `PalletVersion` storage for this instead

lumir-mrkva: here is the change https://github.com/galacticcouncil/HydraDX-node/pull/132

jak-pan (approved):
LGTM. Tested. Let's go 🤘

PR #101, opened 2021-02-10
Benchmarking performance check update

enthusiastmartin:
Includes all pallets weights update.

And updated to performance check tool ( increased margin to 15% and other improvements).


enthusiastmartin:
Updated all pallet weights and made some little improvements to the perf check tool.

Also increased to 15%.

jak-pan (approved):

jak-pan:
Went back to 10% as 15% could really impact performance.

PR #107, opened 2021-02-15
TX Multi payment pallet - review changes

enthusiastmartin:


pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -114,28 +125,34 @@ decl_module! {
 		pub fn set_currency(
 			origin,
 			currency: AssetId,
-		)  -> DispatchResult {
+		)  -> DispatchResult{
```
jak-pan, line 128: https://github.com/paritytech/substrate/blob/171ad2322d7561934b82f0b7e41c58d1710f8c24/primitives/runtime/src/traits.rs#L777

From: Alexander Popiak
Answering the questions about transaction propagation:
Txs that return error when passed to validate_transaction here are not propagated.

jak-pan (commented):
Can we try to validate tx so that it's not spammable? Otherwise LGTM.

jak-pan (approved):

PR #114, opened 2021-02-23
feat(amm)!: add pool and trade limits

### LINKED ISSUE #28: Add minimal pool limit, minimal trade limit and pool existential limit

RoboRambo:
Resolves #28


pallets/amm/src/lib.rs
```diff
@@ -189,8 +189,8 @@ decl_module! {
 			let who = ensure_signed(origin)?;
 
 			ensure!(
-				!amount.is_zero(),
-				Error::<T>::CannotCreatePoolWithZeroLiquidity
+				amount >= MIN_POOL_LIMIT,
+				Error::<T>::InsufficientPoolAssetBalance
```
RoboRambo, line 193: Any Suggestions for a more descriptive Error?

jak-pan: ~ MinimalPoolLiquidityRequirementNotMet ?
We'll have something more verbose in the UI anyways but you should be able to know what happened with just error msg.

pallets/amm/src/lib.rs
```diff
@@ -399,7 +401,7 @@ decl_module! {
 
 			Self::deposit_event(RawEvent::RemoveLiquidity(who.clone(), asset_a, asset_b, liquidity_amount));
 
-			if liquidity_left == 0 {
+			if liquidity_left < MIN_POOL_LIQUIDITY_LIMIT {
```
enthusiastmartin, line 404: Add also a test for this scenario where liquidity left is > 0 but less < the limit.

jak-pan: We also need to do something with the remaining tokens. We shouldn't just leave them sitting there I guess?

RoboRambo: Agree.
Would be grateful for suggestions.

jak-pan: I would check if when exiting pool you left more than minimal pool amount or exactly 0. We'd make sure you can't leave less in the pool and therefore this will never happen (would need comment/doc tho)

jak-pan: * of your tokens.

RoboRambo: > I would check if when exiting pool you left more than minimal pool amount or exactly 0. We'd make sure you can't leave less in the pool and therefore this will never happen (would need comment/doc tho)

If I understand correctly:

- if liquidity_left >= MIN_POOL_LIQ_LIMIT: **remove liquidity**
- if liquidity_left == 0: **remove liquidity & destroy pool**
- if liquidity_left < MIN_POOL_LIQ_LIMIT: **report Error to User?**

jak-pan: Yes! 
but 1 and 2 are pool_liquidity_left 
3 is users_outstanding_liquidity_left (share tokens) < MIN_POOL_LIQ_LIMIT

jak-pan: Because If we only checked pool_liquidity_left in all cases, we could arrive to a state where there are two liq providers that combined have more than MIN_POOL_LIQ_LIMIT but can't withdraw liquidity because their own liquidity left is smaller than < MIN_POOL_LIQ_LIMIT.

pallets/amm/src/lib.rs
```diff
@@ -399,13 +401,20 @@ decl_module! {
 
 			Self::deposit_event(RawEvent::RemoveLiquidity(who.clone(), asset_a, asset_b, liquidity_amount));
 
-			if liquidity_left == 0 {
+			ensure!(
```
jak-pan, line 404: This check MUST be done before any transfers. You have changed the state here and after the fact failing the transaction but with already changed state.

jak-pan: You'll need to check if balance of share tokens of the holder after the trade is higher than minimal limit before doing that operation.

pallets/amm/src/lib.rs
```diff
@@ -399,13 +401,20 @@ decl_module! {
...
+				u128::from(share_token) < MIN_POOL_LIQUIDITY_LIMIT,
+				Error::<T>::MinimalPoolLiquidityRequirementNotMet
+			);
+
+			if liquidity_left >= MIN_POOL_LIQUIDITY_LIMIT || liquidity_left.is_zero() {
```
jak-pan, line 409: This doesn't make much sense to me. Why do you remove token if liquidity_left is more than limit?

pallets/amm/src/lib.rs
```diff
@@ -328,19 +333,21 @@ pub mod pallet {
...
-			ensure!(
-				T::Currency::free_balance(asset_b, &pair_account) >= remove_amount_b,
-				Error::<T>::InsufficientPoolAssetBalance
-			);
-
 			let liquidity_left = total_shares
 				.checked_sub(liquidity_amount)
 				.ok_or(Error::<T>::InvalidLiquidityAmount)?;
 
+			let pool_liquidity_left = user_shares
```
jak-pan, line 340:
```suggestion
			let user_liquidity_left = user_shares
```

pallets/amm/src/lib.rs
```diff
@@ -328,19 +333,21 @@ pub mod pallet {
...
+				.checked_sub(liquidity_amount)
+				.ok_or(Error::<T>::InvalidLiquidityAmount)?;
+
+			if total_shares != liquidity_amount {
+				ensure!(
+					pool_liquidity_left >= MIN_POOL_LIQUIDITY_LIMIT,
```
jak-pan, line 346:
```suggestion
					user_liquidity_left >= MIN_POOL_LIQUIDITY_LIMIT,
```

jak-pan:
bump amm, runtime and spec_version pls

jak-pan:
LGTM will wait for @enthusiastmartin review

pallets/amm/src/lib.rs
```diff
@@ -328,19 +333,21 @@ pub mod pallet {
...
-
 			let liquidity_left = total_shares
 				.checked_sub(liquidity_amount)
 				.ok_or(Error::<T>::InvalidLiquidityAmount)?;
 
+			let user_liquidity_left = user_shares
+				.checked_sub(liquidity_amount)
+				.ok_or(Error::<T>::InvalidLiquidityAmount)?;
+
+			if total_shares != liquidity_amount {
```
enthusiastmartin, line 344: this condition is probably redundant ?  can total shares be less than min pool limit ?

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
 	});
 }
 
+#[test]
+fn remove_liquidity_should_not_work() {
```
enthusiastmartin, line 287: remove liquidity should work;-)

it should not work if requirements are not met.

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
...
+			asset_b,
+			100_000_000,
+			Price::from(10_000)
+		));
+
+		let asset_pair = AssetPair {
+			asset_in: asset_a,
+			asset_out: asset_b,
+		};
+		assert_eq!(AMM::exists(asset_pair), true);
```
enthusiastmartin, line 305: you can just use assert.

```suggestion
		assert!(AMM::exists(asset_pair));
```

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
...
```
enthusiastmartin, line 334: same here. assert would do it

```suggestion
		assert!(AMM::exists(asset_pair));
```

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
...
+
+		assert_ok!(AMM::add_liquidity(
+			Origin::signed(user_two),
+			asset_a,
+			asset_b,
+			MIN_TRADING_LIMIT - 1,
```
enthusiastmartin, line 340: there is not the same min limit for adding liquidity ? just wondering

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
...
+			asset_b,
+			100_000_000,
+			Price::from(10_000)
+		));
+		let asset_pair = AssetPair {
+			asset_in: asset_a,
+			asset_out: asset_b,
+		};
+		let pair_account = AMM::get_pair_id(asset_pair);
+		assert_eq!(AMM::exists(asset_pair), true);
```
enthusiastmartin, line 370: assert
```suggestion
		assert!(AMM::exists(asset_pair));
```

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
...
+
+		assert_ok!(AMM::remove_liquidity(
+			Origin::signed(user),
+			asset_a,
+			asset_b,
+			100_000_000
+		));
+
+		assert_eq!(AMM::total_liquidity(&pair_account), 0);
+		assert_eq!(AMM::exists(asset_pair), false);
```
enthusiastmartin, line 380:
```suggestion
		assert!(!AMM::exists(asset_pair), "Pool not destroyed after removing all liquidity");
```

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
...
+			100_000_000,
+			Price::from(10_000)
+		));
+
+		let asset_pair = AssetPair {
+			asset_in: asset_a,
+			asset_out: asset_b,
+		};
+		let pair_account = AMM::get_pair_id(asset_pair);
+		assert_eq!(AMM::exists(asset_pair), true);
```
enthusiastmartin, line 409:
```suggestion
		assert!(AMM::exists(asset_pair));
```

pallets/amm/src/tests.rs
```diff
@@ -283,6 +283,149 @@ fn remove_zero_liquidity_should_not_work() {
...
+
+		assert_ok!(AMM::remove_liquidity(
+			Origin::signed(user),
+			asset_a,
+			asset_b,
+			100_000_000 - MIN_POOL_LIQUIDITY_LIMIT
+		));
+
+		assert_eq!(AMM::total_liquidity(&pair_account), MIN_POOL_LIQUIDITY_LIMIT);
+		assert_eq!(AMM::exists(asset_pair), true);
```
enthusiastmartin, line 420:
```suggestion
		assert!(AMM::exists(asset_pair));
```

pallets/amm/src/lib.rs
```diff
@@ -328,19 +333,21 @@ pub mod pallet {
...
+				ensure!(
+					user_liquidity_left >= MIN_POOL_LIQUIDITY_LIMIT,
```
enthusiastmartin, line 346: Is this case tested in tests ? I could not figure out which one it is.

enthusiastmartin (commented):
Minor suggestions. Few questions. Otherwise looks ok.

pallets/amm/src/lib.rs
jak-pan: If it was not there you wouldn't be able to withdraw everything since the ensure inside makes sure that you can't leave amount smaller than min liquidity limit left in the pool. We should probably add documentation here stating why we are checking it against users liquidity (that's because users could interlock each other if they left smaller amount than min liquidity limit)

jak-pan: It could also be `liquidity_left == zero` which is basically the same thing.

lumir-mrkva:
@enthusiastmartin @jak-pan please LGTM asap

pallets/amm/src/tests.rs
jak-pan: should be

RoboRambo:
[CannotAddZeroLiquidity](https://github.com/RoboRambo/HydraDX-node/blob/f51d31d760e02f0d45e30a757cc9633667a1a677/pallets/amm/src/lib.rs#L113) & [CannotRemoveLiquidityWithZero](https://github.com/RoboRambo/HydraDX-node/blob/f51d31d760e02f0d45e30a757cc9633667a1a677/pallets/amm/src/lib.rs#L110) & [Test](https://github.com/RoboRambo/HydraDX-node/blob/f51d31d760e02f0d45e30a757cc9633667a1a677/pallets/amm/src/tests.rs#L300) might be redundant now

jak-pan:
@RoboRambo do you have time to check this again after the comments?

jak-pan:
Closes #273

enthusiastmartin:
All pallets have been moved to Basilisk repo.

Therefore, the issue was re-implemented on Basilisk as this has changed since originally implemented.

PR #130, opened 2021-03-06
fix!: Multi Payment storage migration

lumir-mrkva:


pallets/transaction-multi-payment/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = '2018'
 homepage = 'https://github.com/galacticcouncil/hydra-dx'
 license = 'Unlicense'
 repository = 'https://github.com/galacticcouncil/hydra-dx'
-version = '2.0.0'
```
jak-pan, line 9: Semver?

pallets/transaction-multi-payment/src/migrations.rs
```diff
@@ -0,0 +1,13 @@
+use frame_support::traits::PalletVersion;
+
+pub fn migrate_to_2_0_1() -> frame_support::weights::Weight {
```
jak-pan, line 3: SemVer :)

pallets/transaction-multi-payment/Cargo.toml
lumir-mrkva: not sure if we should even use this approach for migrations... we already use a special storage key in the claims pallet

jak-pan: Ideally migration will be a part of PR that introduces the breaking change so it would be bumped there. The question is what to do if it slips? (shouldn't happen) But if yes, we're fixing a breaking change so it should also increase the main version? We can ignore this now but after migrating to the released version we should definitely make it a rule.

lumir-mrkva: we have to change the version if we want to use an embedded storage key for migrations

pallets/transaction-multi-payment/src/migrations.rs
lumir-mrkva: should be named for change not version

pallets/transaction-multi-payment/Cargo.toml
jak-pan: Now the question is... Do we also increase version of runtime? I'd say, yes...? (We already bumped spec version once to "quickfix" but it should have been bumped as a part of this really, we can either bump again to keep runtime version synced with spec_version as it is the same thing, or we can ignore bumping runtime version for this one case to do the same thing.

lumir-mrkva: yeah, we should probably bump everything that is dependent on this pallet

jak-pan: Ok let's not bump runtime this time, but we should think about automating this for the next versions.

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -248,6 +250,26 @@ decl_module! {
...
+				pub struct Module<T>(sp_std::marker::PhantomData<T>);
+				frame_support::decl_storage! {
+					trait Store for Module<T: super::Config> as TransactionPayment {
+						pub AcceptedCurrencies get(fn currencies) config(): super::Vec<super::AssetId>;
+					}
+				}
+			}
+			let version = <Self as GetPalletVersion>::storage_version();
+
+			if version == None || version == Some(PalletVersion::new(3, 0, 0)) {
```
enthusiastmartin, line 265: Shouldn't this be compared against the previous one (2,0,0)?

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -248,6 +250,26 @@ decl_module! {
...
+				previous::AcceptedCurrencies::kill();
+				AcceptedCurrencies::put(OrderedSet::<AssetId>::new());
+				T::BlockWeights::get().max_block
```
enthusiastmartin, line 268: Is the PalletVersion value updated automatically to 3.0.0, resp the current crate version ?

enthusiastmartin (commented):
Looks good to me.

pallets/transaction-multi-payment/src/lib.rs
lumir-mrkva: yes

pallets/transaction-multi-payment/src/lib.rs
lumir-mrkva: it should be

jak-pan (approved):
Ok if tested

PR #132, opened 2021-03-07
fix: claims migration based on pallet version

lumir-mrkva:
parent #98


green-jay:
Removing debug info - Ok. But what was wrong with declarative approach (V1/V2)? Seems inline with docs.

lumir-mrkva:
@green-jay using special storage for migrations was effectively deprecated with this feature https://github.com/paritytech/substrate/pull/7208

PR #136, opened 2021-03-08
fix: claims benchmarking build fix

enthusiastmartin:
Fixed claims pallet build with runtime benchmarks features.

### DIFF
--- a/pallets/claims/src/lib.rs
+++ b/pallets/claims/src/lib.rs
@@ -1,5 +1,6 @@
 #![cfg_attr(not(feature = "std"), no_std)]
-use codec::{Decode, Encode};
+use codec::{Decode, Encode, EncodeLike};
+use core::fmt::Debug;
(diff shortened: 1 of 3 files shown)


pallets/claims/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![cfg_attr(not(feature = "std"), no_std)]
-use codec::{Decode, Encode};
+use codec::{Decode, Encode, EncodeLike};
+use core::fmt::Debug;
```
jak-pan, line 3: Is this needed?

enthusiastmartin: yes as t is currently implemented - we need specify debug and encodelike trait for the currency balance which we use.

green-jay (approved):

jak-pan (approved):

PR #142, opened 2021-03-11
fix(claims): update Alice's signature in claims bench test

green-jay:


enthusiastmartin:
Still getting the same  error unfortunately.

```
---- benchmarking::tests::test_benchmarks stdout ----
thread 'benchmarking::tests::test_benchmarks' panicked at 'Expected Ok(_). Got Err(
    "NoClaimOrAlreadyClaimed",
)', pallets/claims/src/benchmarking.rs:35:13
```

green-jay:
@jak-pan why the commit message rename? AFAIK conventional commits should be in imperative form. Btw. where can I find the guidelines?

green-jay:
@enthusiastmartin this is strange, I get test result OK. What command do you use?

enthusiastmartin:
> @enthusiastmartin this is strange, I get test result OK. What command do you use?

to run benchmark tests - you have to buld with runtime-benchmarks features.

 `cargo test  --features runtime-benchmarks  -p pallet-claims`

green-jay:
@enthusiastmartin hmm, that way I get

```
error: --features is not allowed in the root of a virtual workspace
note: while this was previously accepted, it didn't actually do anything
```

jak-pan:
> @jak-pan why the commit message rename? AFAIK conventional commits should be in imperative form... although thinking about it, it is indeed more natural like "fix: noun". Btw. where can I find the guidelines?

Imperative is just a good practice but you are right we should use it if possible, guidelines are being worked on. The update was more of a :bug: removal than about update word so we can remove that.

enthusiastmartin:
> @enthusiastmartin hmm, that way I get
> 
> ```
> error: --features is not allowed in the root of a virtual workspace
> note: while this was previously accepted, it didn't actually do anything
> ```

run it inside node directory.

green-jay:
thanks, now I get the error too, will check it

pallets/claims/src/benchmarking.rs
```diff
@@ -12,13 +12,21 @@ benchmarks! {
...
+
 		let caller = T::AccountId::decode(&mut &alice_id[..]).unwrap_or_default();
 		let eth_address = EthereumAddress(hex!["8202c0af5962b750123ce1a9b12e1c30a4973557"]);
-		let signature = hex!["ef9816023122208983c11e596446874df3d400d2f9e380a831206d0e91bfb96d54db352fbd62d3cfa8d8674cf63e6a32052ef3cab038e1e7398eac3d048ed5181c"];
 		Claims::<T>::insert(eth_address, T::CurrencyBalance::from(1_000_000_000_000_000_000_u128));
 	}: _(RawOrigin::Signed(caller.clone()), EcdsaSignature(signature))
 	verify {
-		assert_eq!(T::Currency::free_balance(&caller), T::CurrencyBalance::from(2_152_921_504_606_846_975_u128).into());
+		#[cfg(test)]
+		assert_eq!(T::Currency::free_balance(&caller), T::CurrencyBalance::from(1_000_000_000_000_000_000_u128).into());
```
enthusiastmartin, line 28: It would be probably good to check the balance when running the benchmark - as it is , it is only verified in the test run.
```suggestion
	        let expected_balance = T::CurrencyBalance::from(2_152_921_504_606_846_975_u128);
+
+              #[cfg(test)]
+              let expected_balance = T::CurrencyBalance::from(1_000_000_000_000_000_000_u128);
+
-              assert_eq!(T::Currency::free_balance(&caller), T::CurrencyBalance::from(1_000_000_000_000_000_000_u128).into());
+              assert_eq!(T::Currency::free_balance(&caller), expected_balance.into());
```

enthusiastmartin (approved):
One last minor issue. but then lgtm.

pallets/claims/src/benchmarking.rs
green-jay: added

jak-pan:
bump cargo.toml pls :)

jak-pan (approved):

runtime/Cargo.toml
```diff
@@ -29,7 +29,7 @@ tracing-core = {optional = true, version = '0.1.17'}
 module-amm-rpc-runtime-api = {path = '../pallets/amm/rpc/runtime-api', default-features = false, version = '2.0.0'}
 pallet-amm = {path = '../pallets/amm', default-features = false, version = '2.0.0'}
 pallet-asset-registry = {path = '../pallets/asset-registry', default-features = false, version = '2.0.0'}
-pallet-claims = {path = '../pallets/claims', default-features = false, version = '2.0.0'}
+pallet-claims = {path = '../pallets/claims', default-features = false, version = '2.0.1'}
```
jak-pan, line 32: We probably don't need a version here right now since we have just one runtime and (planning for two) but it should always be compatible. One less bump? (We can do this in separate issue)

enthusiastmartin: the version is not needed in this case, as it is local depedency where you'll have always only 1 version. I already did in another  PR for amm pallet, without version.

Issue #147, opened 2021-03-12
Turn on identity pallet


jak-pan:

Roznovjak:
I will do some tests once we decide on the Config values.

PR #163, opened 2021-03-17
feat!: add identity pallet

Roznovjak:
## Description
This PR adds the identity pallet. 
Extrinsic calls are not filtered out.

## Related Issue: 
#147


Roznovjak:
I will do some tests once we decide on the Config values.

Roznovjak:
I tested adding a new registrar, set a fee for judgements, set identity of an account, asked for a judgement and subsequently approved the identity by the registrar.

runtime/src/lib.rs
```diff
@@ -351,6 +352,36 @@ parameter_type_with_key! {
...
 }
 
+type EnsureRootOrHalfCouncil = EnsureOneOf<
+	AccountId,
+	EnsureRoot<AccountId>,
+	pallet_collective::EnsureProportionMoreThan<_1, _2, AccountId, CouncilCollective>,
+>;
+
+parameter_types! {
+	pub const BasicDeposit: Balance = 20 * DOLLARS + 25_800 * MILLICENTS;
```
jak-pan, line 362: WTF :D why it's not just 20 DOLLARS?

jak-pan: Is this some kind of byte fee?

Roznovjak: I used the exact numbers. Ok. I will round the deposits to Dollars.

Roznovjak: I get the numbers from the `deposit()` method.

jak-pan: I'm trying to understand where did the number come from. If it's something that was actually setup by weights then we should state it? If it's by using deposit method why is it not used here?

jak-pan: I found two different implementations. 
One is using dollar amount: https://github.com/paritytech/polkadot/blob/a5c5512efb27edd149d3ae2a21a8b39adab7521a/runtime/kusama/src/lib.rs#L742
One is using deposit function: https://github.com/paritytech/polkadot/blob/a5c5512efb27edd149d3ae2a21a8b39adab7521a/runtime/polkadot/src/lib.rs#L420

It probably depends on what we want to achieve?

Roznovjak: We don't use this function so I substituted the function call with the result.

jak-pan: So we decided to go with only dollar amount in the end.

jak-pan:
Bump runtime cargo.toml version and spec_version pls

runtime/src/lib.rs
jak-pan: I got a question from our community managers about what's good amount to leave non-bound and unfortunately in a rush, I said that 100hdx will be ok not thinking about identity. Can we make this so that it fits that bill? I guess the space cost is 25.8 cents only, so it's still very safe to have 5 * DOLLARS here. It should then fit 100HDX with enough for some tx fees.

jak-pan (commented):
bump runtime and spec_version pls

jak-pan (approved):

PR #170, opened 2021-03-22
chore(perf-check): install bench wizard as part of the check perf script

enthusiastmartin:
## Description
Install bench-wizard cli tool as part of the check_performance script. 

## Related Issue
#167 

## Motivation and Context
Currently, the script exists and user has to manually install the tool and rerun the script afterwards. This change helps with that ,so no need to exit, install and rerun again.

## How Has This Been Tested?
Manually on benchbot machine.

### DIFF
--- a/scripts/check_performance.sh
+++ b/scripts/check_performance.sh
@@ -28,7 +28,7 @@ fi
 echo "OK"
 
 echo -n "Toolchain ...... "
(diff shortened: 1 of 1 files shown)


jak-pan:
Can you also make it update? :)

jak-pan (approved):

Issue #178, opened 2021-03-24
Bug: performanche check script install/upgrade benchwizard problem


enthusiastmartin:
### Description
Installation or upgrade of bench-wizard tool does not work if python3.8 exists in the system and it is not 'default' python used.

### Expected Behavior
install or upgrade of bench-wizard should work correctly. 

### Actual Behavior
if python3.8 exists and it is not the default python - install and upgrade of bench-wizard are done for the default python version.

### Possible Fix
Either handle pip3 similarly to what is done for python3.8 - or remove explicit handling of python3.8 in the script.

### Steps to Reproduce
1. python3.8 present
2. python3 is different python version
3.execute check_performance script without bench-wizard installed.

PR #179, opened 2021-03-24
bug(perf-check): fix bench-wizard install/upgrade

enthusiastmartin:
## Description
Remove python3.8 explicit handling

## Related Issue
#178

### DIFF
--- a/scripts/check_performance.sh
+++ b/scripts/check_performance.sh
@@ -34,7 +34,7 @@ else
         exit 1
 fi
 
-EXPECTED_BENCHWIZARD_VERSION="0.2.1"
+EXPECTED_BENCHWIZARD_VERSION="0.4.1"
 
 echo -n "benchwizard >= $EXPECTED_BENCHWIZARD_VERSION ..... "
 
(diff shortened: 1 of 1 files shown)


scripts/check_performance.sh
```diff
@@ -11,81 +14,71 @@ echo "Prerequisites"
...
-echo "OK"
+echo "OK ($($PYTHON --version))"
 
 echo -n "Toolchain ...... "
 TOOLCHAIN=$(rustup default)
 
-if [[ $TOOLCHAIN = "nightly"* ]]
-then
-        echo "OK"
+if [[ $TOOLCHAIN == "nightly"* ]]; then
```
jak-pan, line 34: we should also check version since you can have 1.47 nightly

enthusiastmartin: this check is actually not really required - because we have pinned the toolchain version. So if you run it,it installs it automatically ( unless you have something else explicitly set but in that case, you should know what you are doing )

so the toolchain check here should pick up the correct pinned version.

enthusiastmartin: Updated to check the actual active toolchain correctly. Still checks only nightly. I can add exact pinned version - but it would have to be updated everytime toolchain is updated too ( not sure about that - exact or not? ).

So it picks it up from the pinned file ( unless user explicitly set to something else - but as mentioned before that's not officially recommended. ).

PR #183, opened 2021-03-25
Feat!: governance

green-jay:
## Checklist:

### DIFF
--- a/runtime/src/constants.rs
+++ b/runtime/src/constants.rs
@@ -2,7 +2,7 @@ pub mod currency {
 	pub use primitives::Balance;
 
 	pub const HDX: Balance = 1_000_000_000_000;
-	pub const DOLLARS: Balance = HDX / 10;	// 10 HDX ~= 1 $
+	pub const DOLLARS: Balance = HDX / 10; // 10 HDX ~= 1 $
 	pub const CENTS: Balance = DOLLARS / 100;
 	pub const MILLICENTS: Balance = CENTS / 1_000;
 }
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -33,7 +33,7 @@ use sp_std::prelude::*;
 use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 
-use frame_system::{limits};
+use frame_system::limits;
 // A few exports that help ease life for downstream crates.
 pub use frame_support::{
 	construct_runtime, parameter_types,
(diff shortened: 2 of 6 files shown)


runtime/src/constants.rs
```diff
@@ -2,9 +2,11 @@ pub mod currency {
 	pub use primitives::Balance;
 
 	pub const HDX: Balance = 1_000_000_000_000;
-	pub const DOLLARS: Balance = HDX / 10;	// 10 HDX ~= 1 $
+	pub const DOLLARS: Balance = HDX / 10; // 10 HDX ~= 1 $
 	pub const CENTS: Balance = DOLLARS / 100;
 	pub const MILLICENTS: Balance = CENTS / 1_000;
+
+	pub const FORTUNE: Balance = u128::MAX;
```
jak-pan, line 9: Why do we have this here? (Didn't find it used anywhere)

jak-pan:
Pls update merge master

runtime/src/constants.rs
green-jay: Meant to be used for disabling certain features (e.g. set for minimum candidacy bond to council)

runtime/src/lib.rs
```diff
@@ -663,11 +682,84 @@ impl pallet_collective::Config<CouncilCollective> for Runtime {
...
+	type ExternalOrigin = pallet_collective::EnsureMember<AccountId, CouncilCollective>;
+	/// A 60% super-majority can have the next scheduled referendum be a straight majority-carries vote.
+	type ExternalMajorityOrigin = pallet_collective::EnsureMember<AccountId, CouncilCollective>;
+	/// A unanimous council can have the next scheduled referendum be a straight default-carries
+	/// (NTB) vote.
+	type ExternalDefaultOrigin = pallet_collective::EnsureMember<AccountId, CouncilCollective>;
+	/// Two thirds of the technical committee can have an ExternalMajority/ExternalDefault vote
+	/// be tabled immediately and with a shorter voting/enactment period.
+	type FastTrackOrigin = pallet_collective::EnsureMember<AccountId, CouncilCollective>;
+	type InstantOrigin = pallet_collective::EnsureMember<AccountId, CouncilCollective>;
```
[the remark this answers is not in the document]
green-jay: Probably not worth discussing at this point (these fast tracks should be handled by techcomm i guess). Democracy might even be disabled for testnet, as we might want to avoid public proposals for systemic changes and leave it only for tech.comm proposals?

runtime/src/constants.rs
green-jay: What I meant as an example, is a prevention for other accounts to submit their candidacy to council

runtime/src/lib.rs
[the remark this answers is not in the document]
green-jay: Have to dig deeper into it if that is even possible

Issue #145, opened 2021-03-11
Postpone financial slashing


jak-pan:
Postpone financial slashing to maximum viable time < bond duration

lumir-mrkva:
slashing shouldn't be disabled completely, just not applied #185

jak-pan:
Yes this is the case. It's unclear from this issue, but we discussed this on a call. And it's part of #184 Slashes are still applied, nominators get kicked from the nomination set and offences are logged. We just don't want somebody to lose money in the first parts of testnet. I think that the punishment by removing nominations is good enough for people to behave correctly.

PR #184, opened 2021-03-25
feat!: disable slashing

martinfridrich:
## Description
Delay slashing to infinity so slashes can be canceled with sudo

## Related Issue
#145


runtime/src/lib.rs
```diff
@@ -456,6 +459,12 @@ parameter_types! {
 	pub MinSolutionScoreBump: Perbill = Perbill::from_rational_approximation(5u32, 10_000);
 }
 
+type SlashCancelOrigin = EnsureOneOf<
+	AccountId,
+	EnsureRoot<AccountId>,
+	pallet_collective::EnsureProportionAtLeast<_3, _4, AccountId, CouncilCollective>,
```
green-jay, line 465: as there will be only once council member, EnsureMember makes a bit more sense

green-jay (commented):
just ooc, is the equivocation fix somehow related to slashing?

runtime/src/lib.rs
martinfridrich: I have to fix babe/grandpa equivocations at first to slashing start to works. I can make it in 2 issues/PRs if we want

green-jay: Ok so the slashing wasn't working in the first place, got it. Probably no need to split it at this point.

jak-pan:
Bump runtime and spec_version pls

jak-pan:
Oops. I'll fix this.

jak-pan (approved):

PR #197, opened 2021-03-30
feat: create testing runtime

### LINKED ISSUE #150: Create new runtime build for testing
Currently we're using call filter to allow or disable certain functions. We should have a build in which this is disabled.
At the same time we should look at these for test runtime:

- [x] Faster epochs
- [x] Faster eras
- [x] Faster elections
- [ ] ??

Roznovjak:
## Description
Create new testing node and runtime

## Related Issue
fixes #150 

## Motivation and Context
Current implementation filters certain extrinsic calls. Testing runtime does not block any calls and makes testing easier.


green-jay:
Had an idea, not entirely sure about it tho. Once the governance is prepared, could we also "supercharge" the testing runtime so that the block production is faster and all the parameters e.g. for governance are tweaked to run faster for testing?

Example param. TEST vs PROD for democracy pallet:
pub const LaunchPeriod: BlockNumber = MINUTES;
vs
pub const LaunchPeriod: BlockNumber = 7 * DAYS;

Roznovjak:
We need to clarify what are our plans with this node. We also need to ensure that testing on this node is enough (no need to repeat the same tests on the 'main' node). I don't think the block production needs to be faster. But I agree that we can tweak some settings.

jak-pan:
> Had an idea, not entirely sure about it tho. Once the governance is prepared, could we also "supercharge" the testing runtime so that the block production is faster and all the parameters e.g. for governance are tweaked to run faster for testing?
> 
> Example param. TEST vs PROD for democracy pallet:
> pub const LaunchPeriod: BlockNumber = MINUTES;
> vs
> pub const LaunchPeriod: BlockNumber = 7 * DAYS;

Yes this is exactly the reason why we're doing this. Let's compile a list here #197

jak-pan:
What's different that we needed to copy all the node files twice?

jak-pan:
We'll also need to update the baseline tokens we're adding to the test node (I.E. We want DAI to be #1 since it will be first token in the pool)

Roznovjak:
> What's different that we needed to copy all the node files twice?

There is no difference right now, but will be in the future. Hence, we need to create a separate directory for it. There is an option to use/include the original node/directory as a crate, but it does not solve this issue (because it would be necessary to make additional updates to the project structure in the future). We can also include files from the original folder/crate, but I don't like this approach because it would mean that two crates share the same source files. We also can't use a library created by the original folder/crate because not all files are included in the library. I can try to split the original folder/crate into multiple smaller parts/libs as Acala did. This way we can import unmodified parts as libraries and include modified files as modules.

runtime/src/lib.rs
```diff
@@ -158,6 +158,9 @@ pub fn native_version() -> NativeVersion {
 }
 
 pub struct BaseFilter;
+#[cfg(feature = "testing-node")]
+mod testing_runtime;
```
lumir-mrkva, line 162: no reason to really separate this to file when only this filter is overridden, either both should be extracted or neither

node/testing_node/Cargo.toml
```diff
@@ -0,0 +1,175 @@
+[patch.crates-io]
```
lumir-mrkva, line 1: testing node should be part of rest of workspace, just another binary target. we will have to build it anyway always for testing.

Roznovjak:
I have modified the following params:
- Calls are not filtered
- LaunchPeriod = MINUTES                  ( changed from 7 days )
- VotingPeriod = MINUTES                    ( changed from 7 days )
- EpochDuration = 10 * MINUTES         ( changed from 4 hours )
- SessinsPerEra = 1                                 ( changed from 6 )
Affected params:
- ReportLongevity = BondingDuration * SessionsPerEra * EpochDuration

Roznovjak:
Moved to #226

PR #200, opened 2021-04-05
fix!:  set DOLLARS constant to correct number

jak-pan:
We've set wrong DOLLARS constant the transactions are too cheap and the storage for identities is 0.05$ instead of 5$

### DIFF
--- a/runtime/src/constants.rs
+++ b/runtime/src/constants.rs
@@ -2,7 +2,7 @@ pub mod currency {
 	pub use primitives::Balance;
 
 	pub const HDX: Balance = 1_000_000_000_000;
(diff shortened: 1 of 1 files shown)


jak-pan:
Please try to re-test this. What happens with already set identities?

Roznovjak:
It's safe to update the DOLLARS constant. Identity pallet stores a deposit value for each account and this value is returned when you are no longer interested in using this feature. When you decide to update your data, the difference is reserved( newly calculated deposit - what is already deposited). The previous 'pattern" is also used in the Treasury pallet and the Elections phragmen pallet (where is the DOLLARS constant variable also used). It has been tested for the identity pallet, the updated storage price for identities is 50 HDX = 5$ ( as expected ).

Roznovjak (approved):

Issue #192, opened 2021-03-30
Prepare re-genesis script


jak-pan:
- [x] Get all valid chain data
- [x] Inject last block and genesis hash of old chain into new chain
- [x] Change protocolId

lumir-mrkva:
where we will store the block and genesis hash of the former chain? 

create special pallet with storage for it?
`map gen -> (lastBlockHash, genesisHash)`

jak-pan:
We can do a system.remark or this. Maybe this way, we could let the old chain run (blocking all txs) and insert last genesis hash after the launch>?

jak-pan:
Done.

Issue #191, opened 2021-03-30
Prepare re-genesis


jak-pan:
We decided for a short epoch time of 10 minutes in the begining of the testnet which wasn't a problem until now. 

Since multiple validators were concerned that this might lead to unnecessary kicking from the validator set and losing nominations if the node is offline for 10 minutes, and also if no blocks were produced for those 10 minutes, the chain could stall like Kusama did before. 

This could result in the need to do a time-warp and coordination effort of all of the validators. We were trying to decide the best way forward as it's unfortunately not possible to just change this setting after the start of the network. 

The current best way forward is re-genesis

We'll need to prepare a script to migrate all of the data to fresh genesis in the following order.

PHASE 1:
- [x] Release and upgrade chain to 4.0.0 + #190 (4.0.1) @jak-pan 
- [x] Prepare re-geneis scraping script probably by modifying [fork-off](https://github.com/maxsam4/fork-off-substrate/blob/master/index.js) #192
- [x] Change epoch to 4 hours #186 
- [x] Update protocolId #192 
- [x] Store genesis and last block of the gen1 in the gen2 #192 
- [x] Update chain_spec and node to connect to the new chain
- [x] Test the upgrade with new epoch time and old validators - validators need to set new keys
- [x] Add governance #183 
- [x] Soft pause gen1 #209 

PHASE 2:
- [x] Start gen2
- [x] Update Polkadot-js/common #195 🥇

PHASE 3: 
- [ ] Kill the gen1 paritytech/substrate#7458
- [ ] Release 5.0.0
- [ ] Coordinate upgrade with validators

PHASE X (Anytime):
- [x] ~Check if it's possible to keep block data and change epoch retroactively during re-genesis~
- [x] ~Check if it's possible to do migration https://github.com/paritytech/substrate/pull/8491#pullrequestreview-624108828~
This won't be possible - See comments below

andresilva:
> Check if it's possible to do migration

A runtime migration alone would not be enough unfortunately, as the client-side is also keeping track of epoch information and it needs to line up with what the runtime expects, and this is made more difficult by the fact that epochs are announced in advance (i.e. when epoch N starts, the client already knows about the details of epoch N+1). Also any on-chain calculations that use `Babe::epoch_start(epoch_index)` would start giving bad results (since it would assume that all epochs since genesis would have the new length).

> Check if it's possible to keep block data and change epoch retroactively during re-genesis

[...]

PR #202, opened 2021-04-07
feat: genesis history pallet

lumir-mrkva:
Pallet provides simple storage linking to the previous chain genesis and last block hash.

required by #191 #192

## TODO
- [x] basic pallet with storage
- [x] sort out defaults issue (fix build)
- [x] tests
- [x] docs

----

### Note
*feel free to change PR target, this targets master for workflow to run (also is based on master)*

Types are generic bytes for maximal compatibility, alternatively, it could be Hash type. Also `PreviousChain` could be posibly unwrapped into two storage values in pallet.


pallets/genesis-history/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+		pub fn assimilate_storage<T: Config>(&self, storage: &mut sp_runtime::Storage) -> Result<(), String> {
+			<Self as frame_support::traits::GenesisBuild<T>>::assimilate_storage(self, storage)
+		}
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+	#[pallet::call]
+	impl<T:Config> Pallet<T> {}
```
jak-pan, line 89: Is this actually required? (I can see it could be.)

lumir-mrkva: yes, macro requires it, same as hooks, but it isn't actually used in runtime. interestingly, events are not required.

PR #205, opened 2021-04-12
[WIP] Builder's Program Milestone 2 Review

apopiak:


pallets/faucet/src/lib.rs
```diff
@@ -115,6 +116,8 @@ pub mod pallet {
 	impl<T: Config> Pallet<T> {
 		#[pallet::weight((0, DispatchClass::Normal, Pays::No))]
 		pub fn rampage_mint(origin: OriginFor<T>, asset: AssetId, amount: Balance) -> DispatchResultWithPostInfo {
+			// REVIEW: Are you sure you want to allow anyone to mint any asset when rampage is
```
jak-pan, line 119: This is just for testing. I don't expect this to be ever turned on on a real chain. Maybe docs would help? Or make it build conditionally?

apopiak: Yeah sure, a quick note on the place where the pallet is included in the runtime would be enough.
I just cannot always tell which parts are testnet and which you intend to stay the same. :shrug:

PR #209, opened 2021-04-12
feat!: soft pause the gen1 chain

jak-pan:


jak-pan:
Closing. Already served it's purpose.

Issue #203, opened 2021-04-09
Bug: Re-benchmark every pallet in runtime


jak-pan:
As noted by @apopiak here: https://github.com/galacticcouncil/HydraDX-node/pull/188#discussion_r610595114 we should re-weight every pallet in our runtime (with our settings) and use that instead of default.

apopiak:
Just saw that Kusama also uses default weights in some places. So I guess you don't have to literally re-weight everything (that was just my gut reaction), but *check* all the pallets on whether they need to be weighted.

Re-weighting all of them is just the safe default. If you are confident that the benchmark will give the same result (because you use the same config on the same benchmarking hardware) you can skip it of course.

jak-pan:
> Just saw that Kusama also uses default weights in some places. So I guess you don't have to literally re-weight everything (that was just my gut reaction), but _check_ all the pallets on whether they need to be weighted.
> 
> Re-weighting all of them is just the safe default. If you are confident that the benchmark will give the same result (because you use the same config on the same benchmarking hardware) you can skip it of course.

It's not that much work / time to do all of them on top of existing benchmarks so we might just add everything to be safe.

jak-pan:
Also I've noticed that every transaction currently costs the same amount to execute. We should have a look at why this happens since exchange:buy should be more expensive than balance transfers

jak-pan:
We should have a look here https://github.com/paritytech/polkadot/blob/5b755428f73345c0015d499482c48c37ce7b9e21/runtime/polkadot/src/constants.rs#L57

PR #210, opened 2021-04-14
weights update - all pallets

enthusiastmartin:
## Description
Use weights from hydra benchmark machine for all pallets the node uses.
Perform benchmarks of each pallet on hydra bench machine.

Benchmarks of some of the pallet cannot complete successfully with the parameters set in the runtime:
Democracy - fails with `MaxVotesReached` error
Tips - fails with `ReasonTooBig` error
Election phragmen - fails with `Failed to submit candidacy`
Balances - fails because Existential deposit is 0.

And the following fails to compile with `runtime-benchmarks` features : collective, election_provider_multi_phase.

Babe and grandpa generate different WeightInfo that the one in the pallets - so not sure what to do about that.

For those pallets mentioned above - it uses  `SubstrateWeights` for now. 

## Related Issue
#203


jak-pan:
Closing needs re-do

PR #220, opened 2021-04-22
ci: automatically creates an EC2 instance for builds

fakirAyoub:
### Each build has its own EC2 instance


.github/workflows/workflow.yml
```diff
@@ -0,0 +1,131 @@
...
+      - name: checkout code
+        uses: actions/checkout@v2.1.0
+      - name: setup node  
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - uses: hashicorp/setup-terraform@v1
+        with:
+          terraform_version: 0.12.25
+      #- name: Declare some variables
```
lumir-mrkva, line 27: remove this, if not useful

.github/workflows/tests.yml
```diff
@@ -1,44 +0,0 @@
-name: Build
```
lumir-mrkva, line 1: would be nice to have this file renamed to `workflow.yml` and then changed to keep history

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,131 @@
...
+      - uses: actions/checkout@v2
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly-2021-02-12
+          target: wasm32-unknown-unknown
+          override: true
+          default: true
+      - run: ./scripts/init.sh
```
lumir-mrkva: scripts init installs rust again, either remove this or install rust task

lumir-mrkva:
hmm, when i triggered multiple builds simultaneosly, second one got provisioned, but got stuck starting build task. i guess that culprit is naming of the runner, I'll chose something random, not actor name so that one actor can trigger multiple builds

lumir-mrkva (changes requested):

.github/workflows/workflow.yml
lumir-mrkva: i already did

lumir-mrkva (changes requested):
hmm, when i triggered multiple builds simultaneosly, second one got provisioned, but got queued on build task (queued). then got stuck because the runner was removed in previous workflow in middle of build. not sure what got provisioned and if it was removed from aws as the last stepn was skipped.

https://github.com/galacticcouncil/HydraDX-node/runs/2406319418

 i guess that culprit is naming of the runner, I'll chose something random, not actor name so that one actor can trigger multiple builds.

lumir-mrkva:
i have tried to run another build, but the action got queued again waiting on worker that never came online
https://github.com/galacticcouncil/HydraDX-node/actions/runs/772807396

lumir-mrkva:
@fakirAyoub it is still in queue, maybe it will help you debug the issue

lumir-mrkva (changes requested):
@fakirAyoub the runner should be removed from repository after instance destroy

here's API reference https://docs.github.com/en/rest/reference/actions#delete-a-self-hosted-runner-from-a-repository

seems like you cannot rerun action of the same commit without removing runner 1st

lumir-mrkva (approved):

lumir-mrkva (changes requested):
git has to be part of the runner environment so that action can clone repo by using it

https://github.com/galacticcouncil/HydraDX-node/runs/2431407553

lumir-mrkva (approved):

PR #225, opened 2021-04-27
fix(node): configure justification import for full node

andresilva:
As part of my investigation in https://github.com/paritytech/substrate/issues/4678#issuecomment-827709562, I also found that this was wrong (although it doesn't fix the original issue).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,6 @@
 name: build-and-deploy
 #${GITHUB_REF##*/}
-on:
-  push:
+on: [push, pull_request]
 
 jobs:
   pop-runner-instance:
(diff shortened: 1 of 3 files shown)


andresilva:
Moving the discussion from https://github.com/paritytech/substrate/issues/4678#issuecomment-827709562 to here. I believe a possible solution to this could be as follows:
- Resync all (or a >2/3 majority) validators from scratch. They should be able to sync to the latest block (#161136) but should all get stuck at best finalized block #104960.
- Once that's done trigger a forced change through the `note_stalled` API using block #104960 as the best finalized block number.
- Any existing client will need to resync as this will be reverting finality.

⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️
I am not sure whether the above will work. It seems that you have sudo on this network so please just test this locally before trying it on the actual network. This will be easier to do if you have a small validator set that you control completely. Since you have sudo you could temporarily reduce the number of validators and make sure only your nodes are in the set. Even if it does work I think people that sync from scratch from then on will be seeing some temporary `Invalid justification` errors (that are expected).

lumir-mrkva (approved):

PR #226, opened 2021-04-27
feat: create testing runtime

Roznovjak:
## Description
Create new testing node and runtime

## Related Issue
Fixes: #150
Local copy of #197 

## Motivation and Context
Current implementation filters certain extrinsic calls. Testing runtime does not block any calls and makes testing easier.


Roznovjak:
The following params have been modified:

Calls are not filtered
LaunchPeriod = MINUTES ( changed from 7 days )
VotingPeriod = MINUTES ( changed from 7 days )
EpochDuration = 10 * MINUTES ( changed from 4 hours )
SessionsPerEra = 1 ( changed from 6 )
Affected params:

ReportLongevity = BondingDuration * SessionsPerEra * EpochDuration

primitives/src/constants.rs
```diff
@@ -50,3 +50,35 @@ pub mod time {
...
+	/// Max fraction of pool to buy in single transaction
+	pub const MAX_OUT_RATIO: u128 = 3;
+
+	/// Max fraction of pool to sell in single transaction
+	pub const MAX_IN_RATIO: u128 = 3;
+
+	/// Trading limit
+	pub const MIN_TRADING_LIMIT: Balance = 1000;
+
+	pub const RUNTIME_AUTHORING_VERSION: u32 = 1;
```
lumir-mrkva, line 71: runtime variables should be probably in the runtime common instead and its cargo version in sync with spec version

lumir-mrkva: or mby it shouldn't be extracted at all

lumir-mrkva: also the spec version should be bumped to 10 already

primitives/src/lib.rs
```diff
@@ -8,19 +7,51 @@ use frame_support::sp_runtime::FixedU128;
...
+
+/// Index of a transaction in the chain.
+pub type Index = u32;
+
+/// A hash of some data used by the chain.
+pub type Hash = sp_core::H256;
+
+/// Digest item type.
+pub type DigestItem = generic::DigestItem<Hash>;
+
```
enthusiastmartin, line 51: I am not very convinced about these types to be moved from runtime into the primitives crate. 

I believe that the runtime common would be better. This makes primitives depends on sp-runtime and sp-core. 

And these types are not going to be used anywhere else other runtime (i think). 

what do you think @lumir-mrkva ?

lumir-mrkva: yeah, if is not needed on higher, there is no reason to push it there

Roznovjak:
Do we want to move AMM related traits and structs (AMM trait, Resolver trait, AssetPair struct, Intention, Fee, ...) to _common_runtime_ dir?

primitives/src/lib.rs
Roznovjak: done

primitives/src/constants.rs
Roznovjak: I have moved runtime variables to common-runtime, synced cargo and spec versions and bumped the spec version to 10.

enthusiastmartin:
> 
> 
> Do we want to move AMM related traits and structs (AMM trait, Resolver trait, AssetPair struct, Intention, Fee, ...) to _common_runtime_ dir?

Nope. These are used in multiple pallets. These should remain in the primitives crate.

enthusiastmartin (commented):
looks alright to me. However i would suggest to try it out by someone who works with both environments more often to see how easy and clear is to use it.

jak-pan:
cc @galacticcouncil/hydradxweb

lumir-mrkva (approved):

lumir-mrkva (changes requested):
needs to be merged with master and versions bumped 

still not sure about spec version being in common instead of in particular runtime, seems like well have to always update spec version and then bump common and runtime crate version. with spec version directly in runtime we're just dealing with that runtime. also there no need to have spec version synced between both runtimes

jak-pan:
How do I purge chain?

this doesn't work
```target/release/hydra-dx purge-chain --testing```
nor
```target/release/hydra-dx purge-chain --dev --testing```

In the current state it's quite confusing to use because you don't really know what --testing and --dev means. We should explicitly state which combinations do what. Maybe figure out better naming for these. (or docs)

**Comment by jak-pan** [comment]
`runtime=testing` flag might be slightly more explicit but we still need more docs on this.

We should also change pallet balances in the testing runtime to polkadot default pallet balances instead of our local custom one so that you can send tokens normally.

**Comment by jak-pan** [comment]
Can we resolve the conflicts pls?

File: node/src/chain_spec.rs [hunk 3]
```diff
@@ -595,3 +590,251 @@ fn create_testnet_claims() -> Vec<(EthereumAddress, Balance)> {
 	claims.push(claim_address_3);
 	claims
 }
+
+pub mod testing_node {
```
jak-pan, line 594: Can we split this into separate files? Starts to get big

node/src/client.rs
```diff
@@ -0,0 +1,352 @@
+//! HydraDX Client abstractions
```
jak-pan, line 1: Do we need this?

node/src/service.rs
```diff
@@ -191,21 +270,41 @@ pub fn new_partial(
...
+		new_full::<testing_hydra_dx_runtime::RuntimeApi, TestingHydraExecutor>(config)
+			.map(|full| full.with_client(Client::TestingHydraDX))
+	} else {
+		new_full(config).map(|full| full.with_client(Client::HydraDX))
+	}
 }
 
 /// Creates a full service from the configuration.
-pub fn new_full_base(
+pub fn new_full<RuntimeApi, Executor>(
```
jak-pan, line 300: can we use this pattern in client.rs?

pallets/xyk/src/lib.rs
```diff
@@ -34,7 +34,8 @@ use frame_support::sp_runtime::{
 };
 use frame_support::{dispatch::DispatchResult, ensure, traits::Get, transactional};
 use frame_system::ensure_signed;
-use primitives::{asset::AssetPair, fee, traits::AMM, AssetId, Balance, Price, MAX_IN_RATIO, MAX_OUT_RATIO};
+use primitives::{asset::AssetPair, fee, traits::AMM, AssetId, Balance, Price};
+use common_runtime::constants::chain::{MAX_IN_RATIO, MAX_OUT_RATIO};
```
jak-pan, line 38: is this a wanted change?

jak-pan:
Just some general questions about possible improvements but other than that LGTM.

runtime/testing-hydra-dx/src/lib.rs
```diff
@@ -0,0 +1,1140 @@
...
+		ideal_stake: 0_160_000,
+		falloff: 1_000_000,
+		max_piece_count: 40,
+		test_precision: 0_005_000,
+	);
+}
+
+parameter_types! {
+	pub const SessionsPerEra: sp_staking::SessionIndex = 6;
+	// Funds are bonded for 28 Days
```
jak-pan, line 453: nope :P

node/src/client.rs
Roznovjak: I believe it is needed because of the abstraction and generics used in [`service.rs`](https://github.com/galacticcouncil/HydraDX-node/blob/cbb81fb14c5959729a0a22e3a45f1a02c9558757/node/src/service.rs#L305) file.

pallets/xyk/src/lib.rs
Roznovjak: I think it makes sense to store common constants in `common_runtime`. There is also an option to create a new submodule for constants in `primitives` module.

runtime/testing-hydra-dx/src/lib.rs
Roznovjak: what values do you suggest?
```
SessionsPerEra = 24
BondingDuration = 2
SlashDeferDuration = 1          // SlashDeferDuration should be less than BondingDuration
```

node/src/service.rs
Roznovjak: Could you be please more specific where to use the pattern? I'm not sure.

jak-pan: I was thinking about making client configurable instead of copying stuff but when looking at it it might not be an option.

runtime/testing-hydra-dx/src/lib.rs
jak-pan: I meant the comment about 28 days. It's 60 minutes right now. Maybe era = session is not perfect also. We might want to do multiple sessions per era. Can you source some sane settings? Like 4 sessions = era.

Roznovjak: what about the following:

pub const EpochDuration: u64 = 10 * MINUTES as u64;
pub const SessionsPerEra: sp_staking::SessionIndex = 4;
// Funds are bonded for 32 hours
pub const BondingDuration: pallet_staking::EraIndex = 2;
pub const SlashDeferDuration: pallet_staking::EraIndex =  1;

jak-pan: Sounds good

Roznovjak: done

jak-pan (approved):

jak-pan:
Update docs pls and we can merge

lumir-mrkva (approved):

PR #230, opened 2021-04-28
ci: Build workflow from fork PR

### LINKED ISSUE #227: Test pull requests from forks
Our current setup requires secrets to run that are unavailable for the build from forks. We need an alternative workflow for those.

lumir-mrkva:
fixes #227 
closes #232

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,6 +57,12 @@ jobs:
         uses: jakejarvis/wait-action@master
         with:
           time: '20s'
+          
+  test-output:
+    runs-on: ubuntu-latest
+    needs: pop-runner-instance
+    steps:
+      - run: echo ${{ needs.pop-runner-instance.outputs.runner }}
 
   build:
     needs: pop-runner-instance
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,7 @@ jobs:
     outputs:
       runner: ${{ steps.determine-runner.outputs.runner }}
     steps:
+      - run: echo "${{ toJSON(github) }}"
       - id: determine-runner
         name: Determine runner
         run: if $IS_LOCAL; then echo "::set-output name=runner::${{ github.sha }}"; else echo "::set-output name=runner::ubuntu-latest"; fi;
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,39 +21,39 @@ jobs:
       - id: determine-runner
         run: if [ -z $AWS_ACCESS_KEY_ID ]; then echo "::set-output name=runner::ubuntu-latest"; else echo "::set-output name=runner::${{ github.sha }}"; fi;
       - name: checkout code
-        if: ${{ $AWS_ACCESS_KEY_ID }}
+        if: ${{ secrets.AWS_ACCESS_KEY_ID }}
         uses: actions/checkout@v2.1.0
       - name: setup node
-        if: ${{ $AWS_ACCESS_KEY_ID }}
+        if: ${{ secrets.AWS_ACCESS_KEY_ID }}
         uses: actions/setup-node@v2
         with:
           node-version: '14'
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -14,50 +15,53 @@ jobs:
...
+      runner: ${{ steps.determine-runner.outputs.runner }}
     steps:
+      - id: determine-runner
+        name: Determine runner
+        run: if [ -z $AWS_ACCESS_KEY_ID ]; then echo "::set-output name=runner::ubuntu-latest"; else echo "::set-output name=runner::${{ github.sha }}"; fi;
       - name: checkout code
+        if: ${{ env.AWS_ACCESS_KEY_ID }}
         uses: actions/checkout@v2.1.0
       - name: setup node
+        if: ${{ env.AWS_ACCESS_KEY_ID }}
```
jak-pan, line 28: Can't we move this one level higher so that it's not repeated? It should be possible according to https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif

jak-pan: Example here https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

lumir-mrkva: i have already tried that, but i wasn;t accepted as valid syntax https://github.com/galacticcouncil/HydraDX-node/actions/runs/793368736

.github/workflows/tests.yml
```diff
@@ -110,27 +114,33 @@ jobs:
...
         id: init
         working-directory: ./infrastructure
         run: terraform init
         continue-on-error: false
 
       - name: Terraform apply
+        if: ${{ env.AWS_ACCESS_KEY_ID }}
         id: apply
         working-directory: ./infrastructure
         run: terraform destroy -var="ec2_pwd=$EC2_SECRET" -var="branch_name=${{ github.sha }}" -auto-approve
```
fakirAyoub, line 146: given the fact that the "runs-on" in the build phase has changed, the branch_name should have ${{ needs.pop-runner-instance.outputs.runner }} as a value;

the name of this variable is to be changed in the next PR

.github/workflows/tests.yml
```diff
@@ -14,50 +15,53 @@ jobs:
...
         id: tf-state-key
+        if: ${{ env.AWS_ACCESS_KEY_ID }}
         working-directory: ./infrastructure
         run: sed -i 's/BRANCH_ACTOR_KEY/${{ env.S3_BACKEND_KEY_NAME }}/g' infra.tf
       - name: Terraform Init
+        if: ${{ env.AWS_ACCESS_KEY_ID }}
         id: init
         working-directory: ./infrastructure
         run: terraform init
-        continue-on-error: false
```
fakirAyoub, line 41: We should keep this unless it's false by default

fakirAyoub (changes requested):

.github/workflows/tests.yml
lumir-mrkva: i removed it because github reported it as an unexpected input https://github.com/galacticcouncil/HydraDX-node/actions/runs/793403551. failing is default behavior anyway

.github/workflows/tests.yml
lumir-mrkva:
```suggestion
        run: terraform destroy -var="ec2_pwd=$EC2_SECRET" -var="branch_name=${{ needs.pop-runner-instance.outputs.runner }}" -auto-approve
```

.github/workflows/tests.yml
jak-pan: Let's try the owner method and then LGTM? We can get rid of all these ifs.

lumir-mrkva: repository owner is no/go, it is the one that fork is pull request against, so always galacticcouncil... I'm trying some other variable that could work

jak-pan (approved):

fakirAyoub (approved):
lgtm

Issue #234, opened 2021-04-29
Change Price type 


martinfridrich:
### Detailed Description
Change `Price` type form `frame_support::sp_runtime::FixedU128` to `substrate_fixed::U64F64`. 
https://github.com/encointer/substrate-fixed 

### Context
`substrate-fixed` have better support for parsing from different types and implement useful functionality.

PR #235, opened 2021-04-29
feat: change Price type

martinfridrich:
#234


Roznovjak:
Is [this part](https://github.com/galacticcouncil/HydraDX-node#interaction-with-the-node) of the docs still valid?

lumir-mrkva (approved):
looks good, i would recommend re-benchmarking after this change

PR #239, opened 2021-05-03
fix(multi-payment): move balances dependency to dev-dependency

enthusiastmartin:
## Description
transaction-multi-payment pallet has dependency on local balances pallet implementation. This is only dev dependency used in test.

This PR moves the dependency to dev dependencies section and it changes to use substrate balances instead of local balances.


jak-pan (approved):

jak-pan:
bump version

PR #241, opened 2021-05-04
fix!: set babe epoch config at genesis

enthusiastmartin:
## Description
Set epoch_config for babe pallet at genesis.

## Motivation and Context
Well, it does not work without it!


jak-pan (approved):

runtime/Cargo.toml
```diff
@@ -5,7 +5,7 @@ homepage = 'https://github.com/galacticcouncil/hydradx-node'
 license = 'Apache 2.0'
 name = 'hydra-dx-runtime'
 repository = 'https://github.com/galacticcouncil/hydradx-node'
-version = '11.0.0'
+version = '11.1.0'
```
jak-pan, line 8:
```suggestion
version = '12.0.0'
```

PR #262, opened 2021-05-25
feat(exchange)!: Update IntentionResolvedDirectTradeFees event

### LINKED ISSUE #248: `IntentionId` should be added to event's data
https://github.com/galacticcouncil/HydraDX-node/blob/6692079e84784eab8915be1fe5f42847b4f8a7fd/pallets/exchange/src/lib.rs#L151

unordered-set:
Add intention_id to the IntentionResolvedDirectTradeFees event.
This commit resolves #248.

This change changes the signature of the IntentionResolvedDirectTradeFees event.
I think it does it for good - to be similar to `IntentionResolvedDirectTrade`

## Description
The change is pretty straightforward, on each direct match we have access to two intentions and by the `fee: true` flag we already know all of the tranansactions which are taking part in taking the fees. Also those transactions has a field `from`. So, by matching that `from` field with intention data we are computing the intention IDs. 

In my opinion it is a pretty robust way of doing things. Because I didn't touch the matching rules, and also, the change will adopt itself to any further changes in the fees rules.

## Related Issue
Fixes: #248 

## Motivation and Context
That missing field is pretty important to any tools which want to drill down into the transaction costs. Otherwise they all will have the same snippet of decyphering other emitted events.

## How Has This Been Tested?
Run tests locally. Worked. Unbelivable. Broke tests, verified they got broken. Fixed back.

## Checklist:

- [v] I have updated the documentation if necessary.
- [v] This is a breaking change.

### DIFF
--- a/pallets/exchange/src/direct.rs
+++ b/pallets/exchange/src/direct.rs
@@ -234,7 +234,8 @@ impl<'a, T: Config> DirectTradeData<'a, T> {
 			.expect("Cannot fail. Checks should have been done prior to this.");
 
 			if transfer.fee_transfer {
-				Self::send_trade_fee_event(transfer.from, transfer.to, transfer.asset, transfer.amount);
+				let intention = if transfer.from == &self.intention_a.who { self.intention_a } else { self.intention_b };
+				Self::send_trade_fee_event(transfer.from, intention, transfer.to, transfer.asset, transfer.amount);
 			}
 		}
 		true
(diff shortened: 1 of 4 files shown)


enthusiastmartin (approved):
looks ok to me.

pallets/exchange/src/direct.rs
```diff
@@ -260,9 +261,10 @@ impl<'a, T: Config> DirectTradeData<'a, T> {
 	}
 
 	/// Send pallet event after a fee is transferred.
-	fn send_trade_fee_event(from: &T::AccountId, to: &T::AccountId, asset: AssetId, amount: Balance) {
+	fn send_trade_fee_event(from: &T::AccountId, intention: &Intention<T>, to: &T::AccountId, asset: AssetId, amount: Balance) {
```
jak-pan, line 264: Do we only send fee from one side?

jak-pan: @enthusiastmartin

enthusiastmartin: it is sent for each intention/transaction. you can see it in the tests.

PR #263, opened 2021-05-26
fix!: WithFee::with_fee method

Roznovjak:
## Description
fix WithFee::with_fee() method and also add without_fee() method.

https://github.com/galacticcouncil/HydraDX-node/blob/694cce69ee44890dcbd05c9ef1cd83079e3ce1ac/primitives/src/lib.rs#L121-L124

The current implementation computes `value * (100% - fee_percentage)` instead of `value + fee`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2357,7 +2357,7 @@ dependencies = [
 
 [[package]]
 name = "hydra-dx"
-version = "6.0.0"
+version = "5.1.0"
 dependencies = [
  "frame-benchmarking",
  "frame-benchmarking-cli",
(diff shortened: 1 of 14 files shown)


enthusiastmartin (commented):
looks ok to me. but i am not very convinced that we need to update all versions of all pallets and runtime too. 

I would just update the primitives crate version.

Roznovjak:
What do you think @jak-pan ? It is a breaking change, but the affected method was never used before in the code( at least not in this repository).

jak-pan:
Just runtime and impl_version in runtime. Where else is this used?

Roznovjak:
We don't use `with_fee()` in our node repositories. We also need `without_fee` method in LBP so I have implemented it. I am not sure if I answered your question.

jak-pan:
Only runtime then.

Roznovjak:
I have reverted impl_runtime version to the previous one. The only version changes are now in `primitives` and `runtime` package.

enthusiastmartin (approved):

PR #264, opened 2021-05-29
refactor: remove custom balances pallet

enthusiastmartin:
## Description
Removed custom balances pallet and restricted balances calls in the call filter.

## How Has This Been Tested?
locally


jak-pan:
I think that this change has made this possible, when looking at this though, we should look elsewhere if we can use this or if it's not breaking something we use https://github.com/paritytech/substrate/pull/6318

Roznovjak (approved):
LGTM

PR #269, opened 2021-05-31
fix(amm): buy_price calculation

Roznovjak:
## Description
Fee in buy_price calculation is wrongly assigned to the `amount_out`. The correct operation order is to compute `amount_in` first, and then apply fee to this value in the next step.

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -617,8 +617,12 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
 			Self::calculate_fee(amount)?
 		};
 
+		let amount_without_fee = amount
+			.checked_sub(transfer_fee)
+			.ok_or(Error::<T>::SellAssetAmountInvalid)?;
+
 		let sale_price =
-			hydra_dx_math::calculate_out_given_in(asset_in_reserve, asset_out_reserve, amount - transfer_fee)
+			hydra_dx_math::calculate_out_given_in(asset_in_reserve, asset_out_reserve, amount_without_fee)
 				.map_err(|_| Error::<T>::SellAssetAmountInvalid)?;
 
 		ensure!(asset_out_reserve > sale_price, Error::<T>::InsufficientAssetBalance);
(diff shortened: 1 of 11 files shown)


pallets/xyk/src/lib.rs
```diff
@@ -604,21 +603,25 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
 		let transfer_fee = if discount {
 			Self::calculate_discounted_fee(amount)?
 		} else {
 			Self::calculate_fee(amount)?
 		};
 
-		let sale_price = hydra_dx_math::calculate_out_given_in(asset_in_total, asset_out_total, amount - transfer_fee)
-			.map_err(|_| Error::<T>::SellAssetAmountInvalid)?;
+		let sale_price =
+			hydra_dx_math::calculate_out_given_in(asset_in_reserve, asset_out_reserve, amount - transfer_fee)
```
enthusiastmartin, line 621: Let's fix this as well and use checked math for fee subtraction

pallets/xyk/src/lib.rs
```diff
@@ -604,21 +603,25 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
+				.map_err(|_| Error::<T>::SellAssetAmountInvalid)?;
 
-		ensure!(asset_out_total >= sale_price, Error::<T>::InsufficientAssetBalance);
+		ensure!(asset_out_reserve > sale_price, Error::<T>::InsufficientAssetBalance);
```
enthusiastmartin, line 624: why not >=?

pallets/xyk/src/lib.rs
```diff
@@ -731,27 +734,25 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
+			.ok_or(Error::<T>::BuyAssetAmountInvalid)?;
 
 		ensure!(
-			T::Currency::free_balance(assets.asset_in, who) >= buy_price,
+			T::Currency::free_balance(assets.asset_in, who) >= buy_price_with_fee,
 			Error::<T>::InsufficientAssetBalance
 		);
 
-		ensure!(max_limit >= buy_price, Error::<T>::AssetBalanceLimitExceeded);
+		ensure!(max_limit >= buy_price_with_fee, Error::<T>::AssetBalanceLimitExceeded);
```
enthusiastmartin, line 755: Perhaps you can move this check before the free balance check.

enthusiastmartin (commented):
Just few minor comments. Otherwise looks good to me.

enthusiastmartin:
Btw i dont think this is a breaking change. IT is just implementation change but it does not break anything.

pallets/xyk/src/lib.rs
Roznovjak: just to be sure that there is always some liquidity left in the pool

Roznovjak:
You are right. A change in behavior to fix a bug is not considered as a breaking change.

pallets/xyk/src/lib.rs
Roznovjak: done

pallets/xyk/src/lib.rs
Roznovjak: done

enthusiastmartin (approved):

Issue #268, opened 2021-05-31
bug: Exchange pallet error incorrect and inacurate


enthusiastmartin:
### Description
Change Exchange pallet's Error `AssetBalanceLimitExceeded` and split it into two errors - one for sell case and one for buy.

### Actual status
Currently there is only one error for two different cases and error name itself is not accurate and the description is incorrect as well. The documentation description for these cases is inverted

And this error is also used for another case ( search for TODO in exchange pallet ) - it might be worth to fix this too. 

### Expected status
There should be 2 separate errors - Exceeded and NotReached ( or something similar ) to better indicate what's the problem when it actually occurs.

PR #271, opened 2021-06-02
fix(exchange): change pallet errors

nasqn:
## Description

 `AssetBalanceLimitExceeded`  is splitted into two errors, one for sell case and one for buy. 

Fixes: #268  

## Motivation and Context

There was only one error for two different cases and error name itself was not accurate. Two different errors helps to better indicate what's the problem and when it actually occurs.

## How Has This Been Tested?
Run tests locally, fixed the broken one.

## Checklist:

- [ v ] I have updated the documentation if necessary.
- [v ] This is a breaking change.

### DIFF
--- a/pallets/exchange/src/lib.rs
+++ b/pallets/exchange/src/lib.rs
@@ -181,8 +181,11 @@ pub mod pallet {
 		/// Insufficient asset balance.
 		InsufficientAssetBalance,
 
-		/// Given trading limit has been exceeded (Sell) or has Not been reached (buy).
-		AssetBalanceLimitExceeded,
+		/// Given trading limit has been exceeded (buy).
+		TradeAmountExceededLimit,
+
+		/// Given trading limit has not been reached (sell).
+		TradeAmountNotReachedLimit,
 
 		/// Overflow
 		ZeroSpotPrice,
@@ -580,7 +583,7 @@ impl<T: Config> Resolver<T::AccountId, Intention<T>, Error<T>> for Pallet<T> {
 						if dt.amount_from_a < matched_intention.trade_limit {
 							Self::send_intention_error_event(
 								&matched_intention,
-								Error::<T>::AssetBalanceLimitExceeded.into(),
+								Error::<T>::TradeAmountNotReachedLimit.into(),
 							);
 							continue;
 						}
@@ -589,7 +592,7 @@ impl<T: Config> Resolver<T::AccountId, Intention<T>, Error<T>> for Pallet<T> {
 						if dt.amount_from_b > matched_intention.trade_limit {
 							Self::send_intention_error_event(
 								&matched_intention,
-								Error::<T>::AssetBalanceLimitExceeded.into(),
+								Error::<T>::TradeAmountExceededLimit.into(),
 							);
 							continue;
 						}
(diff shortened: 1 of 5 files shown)


nasqn:
This issue looked very simple, but when it's about namings of public interfaces - you should understand all code around it pretty well to give good names.

I was trying to figure out how exchange pallete works, but it's not too easy :) so I'm not sure that I've solved the issue correctly.

nasqn:
Also I don't know is changing of errors a breaking change?

pallets/exchange/src/lib.rs
```diff
@@ -181,8 +181,14 @@ pub mod pallet {
...
-		/// Given trading limit has been exceeded (Sell) or has Not been reached (buy).
-		AssetBalanceLimitExceeded,
+		/// Given trading limit has been exceeded (buy).
+		TradeAmountExceededLimit,
+
+		/// Given trading limit has not been reached (sell).
+		TradeAmountNotReachedLimit,
+
+		/// Intentions for direct trading are matched wrong.
+		IncorrectIntentionsMatch,
```
nasqn, line 191: If I understand the scenario right, when this error occurs it means that main intention and array of matched intentions was choosen incorrectly as parameters of `resolve_matched_intentions` func. That's why i picked up this name.

pallets/exchange/src/lib.rs
```diff
@@ -627,7 +633,7 @@ impl<T: Config> Resolver<T::AccountId, Intention<T>, Error<T>> for Pallet<T> {
 				if rest_in_diff.is_none() || rest_out_diff.is_none() {
```
nasqn, line 633: by the way, do we need `rest_out_diff.is_none()` check? Seems it unnecessary, because there is already a check `else if amount_a_in < amount_b_out` for this scope

pallets/exchange/src/tests.rs
```diff
@@ -832,7 +832,7 @@ fn sell_trade_limits_respected_for_matched_intention() {
 				user_3_sell_intention_id,
 				DispatchError::Module {
 					index: 1,
-					error: 2,
+					error: 3,
```
nasqn, line 835: is this just index of error in pallet's `enum Error<T> `, right?

enthusiastmartin: yes.

enthusiastmartin:
This looks ok , except the new IncorrectMatch error. 

However, there is a bug in this area #274 ( fix coming soon ) which will likely to make some changes in this area.

So, let's first fix that bug and then make changes in this PR.

unordered-set:
> Also I don't know is changing of errors a breaking change?

Yes, it is. Ui has some expectations about error names to display valid translations etc. One you change it, fronted needs to be updated accordingly. So you also need to update the palette version

jak-pan:
> > Also I don't know is changing of errors a breaking change?
> 
> Yes, it is. Ui has some expectations about error names to display valid translations etc. One you change it, fronted needs to be updated accordingly. So you also need to update the palette version

Actually, UI shows errors based on `/// documentation for error` it gets from the node. But it's indeed a breaking change since the deposited events need to be deserialized differently now and that means we need to increase pallet version and also spec_version and runtime version.

PR #282, opened 2021-06-17
feat: vested claim

green-jay:
## Description
Add vesting capability to ETH-based claims.

### DIFF
--- a/pallets/claims/src/lib.rs
+++ b/pallets/claims/src/lib.rs
@@ -79,7 +79,7 @@ pub mod pallet {
 		type Prefix: Get<&'static [u8]>;
 		type WeightInfo: WeightInfo;
 		type Currency: Currency<Self::AccountId>;
-		type VestingSchedule: VestingSchedule<Self::AccountId, Moment=Self::BlockNumber, Currency=Self::Currency>;
+		type VestingSchedule: VestingSchedule<Self::AccountId, Moment = Self::BlockNumber, Currency = Self::Currency>;
 		// This type is needed to convert from Currency to Balance
 		type CurrencyBalance: From<Balance>
 			+ Into<<Self::Currency as Currency<<Self as frame_system::Config>::AccountId>>::Balance>;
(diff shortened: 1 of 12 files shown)


green-jay:
Agreed to migrate to ORML vesting because of the possibility of accounts having multiple vesting schedules. Also will be adjusted to calculate and distribute already vested + to be vested amounts, according to the global vesting schedule

pallets/claims/src/lib.rs
```diff
@@ -105,15 +106,26 @@ pub mod pallet {
 	#[pallet::getter(fn claims)]
 	pub type Claims<T: Config> = StorageMap<_, Blake2_128Concat, EthereumAddress, BalanceOf<T>, ValueQuery>;
 
+	/// Vesting schedule for a claim.
+	/// The balance is how much should be unlocked per block.
+	/// The block number is when the vesting should start.
+	#[pallet::storage]
+	#[pallet::getter(fn vesting)]
+	pub type Vesting<T: Config> = StorageMap<_, Identity, EthereumAddress, (BalanceOf<T>, T::BlockNumber), ValueQuery>;
```
enthusiastmartin, line 114: one suggestion here would be to use OptionQuery instead of ValueQuery.

It will simplify the "check and get" in the code - you can just do one `get` and it returns an option  and it will be None if it is does not exists.

pallets/claims/src/lib.rs
```diff
@@ -134,9 +149,17 @@ pub mod pallet {
 			let sender = ensure_signed(origin)?;
 
 			let (balance_due, address) = Self::validate_claim(&sender, &ethereum_signature)?;
-
-			Self::process_claim(sender, balance_due, address)?;
-
+			let vesting_check = Vesting::<T>::contains_key(&address);
+
+			if vesting_check {
```
enthusiastmartin, line 154: AS mentioned in previous comment, if you use OptionQuery, you can do something like :

```
If Some(vesting) = Vesting::<T>::get(&adress) {
...
}else{
...
}
```

enthusiastmartin (commented):
I will wait with the review for the change to orml vesting.

However,i have some code improvement so far.

PR #283, opened 2021-06-21
chore!: v0.9.5

enthusiastmartin:
## Description
v0.9.5


Roznovjak (commented):
LGTM, but I'm not super familiar with the changes in `service.rs`.

pallets/exchange/src/tests.rs
```diff
@@ -177,6 +177,10 @@ fn sell_test_pool_finalization_states() {
 				user_3_sell_intention_id,
 			)
 			.into(),
+			orml_tokens::Event::Reserved(asset_a, 2, 1000000000000).into(),
+			orml_tokens::Event::Reserved(asset_b, 3, 2000000000000).into(),
+			orml_tokens::Event::Reserved(asset_b, 2, 2000000000).into(),
+			orml_tokens::Event::Reserved(asset_b, 3, 4000000000).into(),
```
green-jay, line 183: was wondering if this repetitive part could be setup and reused as a function?

runtime/src/lib.rs
```diff
@@ -210,6 +210,7 @@ impl Filter<Call> for BaseFilter {
 
 			Call::XYK(_)
 			| Call::AssetRegistry(_)
+			| Call::Balances(_)
```
green-jay, line 213: will staking, governance etc. work?

pallets/exchange/src/tests.rs
enthusiastmartin: yes.possibly. but i have noticed little discrepancy in this, so i investigate that first. And when all good, we can handle this kind of improvement.

runtime/src/lib.rs
enthusiastmartin: Yes, it should. It was tested previously with treasury tip.

lumir-mrkva (changes requested):
please bump versions and test running main chain node

lumir-mrkva (approved):

Issue #275, opened 2021-06-07
change: xyk limit error name not accurate


enthusiastmartin:
### Description
In XYK pallet - there is an error called `AssetBalanceLimitExceeded`. It is used incorrectly in one case. 

### Actual status
This error is used for two different scenarios - in buy and sell. However for the sell part , it is not accurate because the limit represents the minimum amount to buy, so the limit is not exceeded here but rather not reached.

### Expected status
New error should be introduced for the sell case to be more accurate representation of the issue.

PR #284, opened 2021-06-21
fix(xyk)!: change xyk pallet errors

nasqn:
## Description

 `AssetBalanceLimitExceeded`  is splitted into two errors, one if limit exceeded and one if limit not reached. 

Fixes: #275  

## Motivation and Context
There was only one error for two different cases and error name itself was not accurate. Two different errors helps to better indicate what's the problem and when it actually occurs.

## How Has This Been Tested?
Run tests locally, fixed the broken one.

## Checklist:

- [v ] I have updated the documentation if necessary.
- [v ] This is a breaking change.


nasqn:
Why do we need specific errors even for very similar situations? 

for example, instead of:

```
		/// Overflow
		AddAssetAmountInvalid, // no tests
		/// Overflow
		RemoveAssetAmountInvalid, // no tests
		/// Overflow
		SellAssetAmountInvalid, // no tests
		/// Overflow
		BuyAssetAmountInvalid, // no tests
		/// Overflow
		FeeAmountInvalid, // no tests
```
we could have only `AssetAmountInvalid` and use it in all cases.


I'm asking, because I have the similar situation - exceeding the limit in `add_liquidity` function an in `buy` function. So I added more common error `AssetAmountExceededLimit` and use it in both places

Or should I add one more error for `add_liquidity` function and give more accurate names?

enthusiastmartin:
> 
> 
> Why do we need specific errors even for very similar situations?
> 
> for example, instead of:
> 
> ```
> 		/// Overflow
> 		AddAssetAmountInvalid, // no tests
> 		/// Overflow
> 		RemoveAssetAmountInvalid, // no tests
> 		/// Overflow
> 		SellAssetAmountInvalid, // no tests
> 		/// Overflow
> 		BuyAssetAmountInvalid, // no tests
> 		/// Overflow
> 		FeeAmountInvalid, // no tests
> ```
> 
> we could have only `AssetAmountInvalid` and use it in all cases.
> 
> I'm asking, because I have the similar situation - exceeding the limit in `add_liquidity` and `buy` functions. So I added more common error `AssetAmountExceededLimit` and use it in both places
> 
> Or should I add one more error for `add_liquidity` function and give more accurate names?


I agree that this could be simplified to have only one error. It was over-enthusiastic approach when the pallet was initially implemented. 

I believe that the specific error  wont be very helpful anyway. You can see which extrinsic it has been raised for.

These errors are always returned when there is problem in calculation, an arithmetic problem - so i would also consider adding simple 'Overflow' error for all these cases.

nasqn:
So, is current implementation OK?

enthusiastmartin (approved):

Issue #207, opened 2021-04-12
Existential deposits


lumir-mrkva:
currently is set to 0, because we allow users to pay transactions fee with 3rd party currencies. so theoretically u don't have to touch HDX ever.

ideally, we would determine the existence of accounts by the sum of value across all assets and compare it with some constant DAI minimal value.

different solutions should be explored:
- always store this sum and update it with every transfer (could be slow)
- allow off-chain audit of account of balances and reimburse bots removing accounts (possibly abusable)

enthusiastmartin:
Another possible solution to consider would be to use the selected currency in which user wants to pay as currency determining the account existence.  
Because it is not allowed to select currency with no balance. 

But does not deal with destroying account where there are assets for this account. But perhaps it could be used as first step to mark account for "destroy" - so the off chain audit can only go through these marked accounts and not all of them.

Just an idea.

jak-pan:
> Another possible solution to consider would be to use the selected currency in which user wants to pay as currency determining the account existence.
> Because it is not allowed to select currency with no balance.
> 
> But does not deal with destroying account where there are assets for this account. But perhaps it could be used as first step to mark account for "destroy" - so the off chain audit can only go through these marked accounts and not all of them.
> 
> Just an idea.

We need to be careful here as the person can have $1 worth of token he is paying the txs with but hold $10000 in value of other tokens. This would wipe his account clean. I'd go more by the sum of values of these tokens. We can make some "optimizations" to balances / currencies so that user will not be able to do this under normal conditions. 

The problem with the off-chain cleanup is that when a holder that didn't fall below existential deposit before, will fall into this range, it will wipe him so, we probably still need to sum. 

Other solution could be to only do this check when the account has transacted in a block making this much easier to handle.

jak-pan:
Let's be also careful about this https://substrate.dev/docs/en/knowledgebase/advanced/account-info (ref-counting) which might actually be the reason we've locked some accounts.

enthusiastmartin:
Using Basilisk pallets (asset registry and duster ) - Existential deposit can be set up in similar way to what has been done on basilisk.

So once #299  is merged - it'd be possible to set it up.

lumir-mrkva:
should be configured in parachain

lumir-mrkva:
already done

PR #288, opened 2021-06-29
feat: account duster/existential deposits

enthusiastmartin:
Added duster pallet and offchain duster pallet  to handle account dusting

#207


pallets/offchain-duster/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+		if !signer.can_sign() {
+			return Err("No local accounts available. Consider adding one via `author_insertKey` RPC.");
+		}
+
+		let results = signer.send_signed_transaction(|_account| DusterCall::dust_account(from.clone(), currency_id));
+
+		for (acc, res) in &results {
+			match res {
+				Ok(()) => {
+					frame_support::log::info!("Dust moved successfully to [{:?}]", acc.id)
```
jak-pan, line 137: We should refund the tx fee here

pallets/offchain-duster/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+use sp_core::crypto::KeyTypeId;
+use sp_runtime::traits::Saturating;
+
+pub use pallet_duster::Call as DusterCall;
+
+// Re-export pallet items so that they can be accessed from the crate namespace.
+pub use pallet::*;
+
+pub const OFFCHAIN_WORKER_LOCK: &[u8] = b"hydradx/offchain-duster/lock/";
+pub const LOCK_DURATION: u64 = 100;
```
jak-pan, line 43: Is this ms? If yes isn't it too short? We should probably do this only every x blocks

enthusiastmartin: yes. but it is for one dust_account call. Lock guard is extended after each call. 

It is because it is hard to determine correct duration as it may vary.

pallets/offchain-duster/src/lib.rs
enthusiastmartin: hm. Not sure how we can do that here. I'll investigate.

There is a reward for successful account dusting. Isn't that enough ?  It is possible to configure it so it'd cover the tx fee + some additiional reward.

jak-pan:
Closing in favor of Basilisk

PR #289, opened 2021-06-29
feat!: update fee calculation

Roznovjak:
## Description
Current fees for a TX are set too low. Tune the costs.


Roznovjak:
Transfer costs 0.0047 $ (the same as in Basilisk)
Full block: 395 $ (393 in Basilisk)
Fee multiplier after one hour of operation with fully loaded blocks: 1.0135 (the same as in Basilisk)
After one day: 1.3826 (the same as in Basilisk)

jak-pan (approved):

Issue #290, opened 2021-06-29
Add orml vesting


green-jay:
Currently there is no possibility to vest balances. Vesting should be supported because of upcoming Crowdloan/LBP/other types of claims.

Implementation:

- [x] add orml vesting

green-jay:
@jak-pan linked PR has not been merged yet, sure about closing this issue?

PR #291, opened 2021-06-29
feat: vesting

green-jay:
## Related Issue
Fixes: #290


runtime/hydra-dx/src/lib.rs
```diff
@@ -820,6 +813,39 @@ impl pallet_scheduler::Config for Runtime {
...
+	fn successful_origin() -> Origin {
+		Origin::from(RawOrigin::Signed(Default::default()))
+	}
+}
+
+impl orml_vesting::Config for Runtime {
+	type Event = Event;
+	type Currency = Balances;
+	type MinVestedTransfer = MinVestedTransfer;
+	type VestedTransferOrigin = OnlyGalacticCouncil;
```
jak-pan, line 843: root or galactic council

runtime/common/src/lib.rs
```diff
@@ -277,6 +277,12 @@ parameter_types! {
 	pub const TechnicalMaxMembers: u32 = 10;
 }
 
+// pallet vesting
+parameter_types! {
+	pub MinVestedTransfer: Balance = 1_000 * HDX;
+	pub const MaxVestingSchedules: u32 = 100;
```
jak-pan, line 283: Is this for one person or total?

runtime/common/src/lib.rs
```diff
@@ -277,6 +277,12 @@ parameter_types! {
 	pub const TechnicalMaxMembers: u32 = 10;
 }
 
+// pallet vesting
+parameter_types! {
+	pub MinVestedTransfer: Balance = 1_000 * HDX;
```
enthusiastmartin, line 282: Is this per account ? Isnt it too high ? 

Eg. recent crowdloan bonus rewards are lower than this if contributed 1KSM for example.

Roznovjak: It is per `vesting_transfer`, so you can't vest less than this amount.

runtime/common/src/lib.rs
Roznovjak: for one account

runtime/hydra-dx/src/lib.rs
Roznovjak: AFAIK we can't use the `root` account here because the implementation of `VestedTransferOrigin` requires `AccountId` as a return type and it is something what the `root` account doesn't have https://docs.substrate.io/rustdocs/latest/frame_system/enum.RawOrigin.html.

green-jay: You could call it with Galactic account in case origin is Root.

Roznovjak: Do you mean something like this?
```
impl EnsureOrigin<Origin> for GalacticCouncilOrRoot {
	type Success = AccountId;

	fn try_origin(o: Origin) -> Result<Self::Success, Origin> {
		Into::<Result<RawOrigin<AccountId>, Origin>>::into(o).and_then(|o| match o {
			RawOrigin::Signed(caller) => {
				if caller == hex!["0abad795adcb5dee45d29528005b1f78d55fc170844babde88df84016c6cd14d"].into() {
					Ok(caller)
				} else {
					Err(Origin::from(Some(caller)))
				}
			}
			RawOrigin::Root => Ok(hex!["0abad795adcb5dee45d29528005b1f78d55fc170844babde88df84016c6cd14d"].into()),
			r => Err(Origin::from(r)),
		})
	}
```

green-jay: yup

runtime/common/src/lib.rs
green-jay: why don't we just set it to 0?

Roznovjak: what about setting it to `NATIVE_EXISTENTIAL_DEPOSIT`?

green-jay: yes makes sense too

runtime/hydra-dx/src/lib.rs
green-jay: i would also rather use constant everywhere than accountid value directly

runtime/common/src/lib.rs
Roznovjak: Oh, we can't use it because we don't have the duster pallet there.

lumir-mrkva: set to 0 for now, it will have to be increased if we allowed vested transfer to anyone

jak-pan: We should definitely set it to existential deposit and we have that here. Just extract it to const.

Roznovjak:
LGTM

lumir-mrkva:
@Roznovjak next time if you want to reformat everything, please at least separate this to the different commit

lumir-mrkva (approved):

PR #299, opened 2021-08-19
chore: use basilisk pallets

enthusiastmartin:
Use pallets from Basilisk as they are developed there and to avoid duplication.

Note: versions of the pallets are pinned to rev from feat/asset-registry branch to include newly refactored asset-registry.


lumir-mrkva (approved):

node/src/chain_spec.rs
```diff
@@ -469,9 +468,8 @@ fn lerna_genesis(
 			key: root_key,
 		},
 		asset_registry: AssetRegistryConfig {
-			core_asset_id: CORE_ASSET_ID,
-			asset_ids: vec![],
-			next_asset_id: 1,
+			asset_names: vec![],
+			native_asset_name: b"HDX".to_vec(),
```
lumir-mrkva, line 472: @enthusiastmartin shouldn't we do some storage migration because of this? i guess the empty array is fine, but the native asset name won't be set in the live chain. alternatively, we could sudo set the storage value manually.

lumir-mrkva: did you test runtime upgrade?

lumir-mrkva (changes requested):
- [ ] test runtime upgrade

node/src/chain_spec.rs
enthusiastmartin: Hm, correct - native details wont be in the storage. 

It is actually not an issue. We (so far) dont use the native name at all nor we need the native details in the storage. We always use only the native asset id.

And we can do it manually anyway.

However, I'll look into the upgrade (so i learn one or thing as well )

enthusiastmartin:
@lumir-mrkva I'd just leave it. It is just more hassle to implement the upgrade migration.  It is possible to do it manually via update call.

enthusiastmartin:
This is obsolete.

There is a new repo for all crates and pallets which are or will be used by Basilisk and HydraDX. 

And this kind of refactor will be done as pallets and features are introduce to HydraDX parachain.

PR #300, opened 2021-08-20
fix: add pool id to xyk pool created/destroyed events

enthusiastmartin:


jak-pan:
@enthusiastmartin can we also add pair_account to the events?

mckrava:
But this commit `fix: add pool id to xyk pool created/destroyed events` adds `shared_token_id` but not `pool id` :)

pallets/xyk/src/lib.rs
```diff
@@ -169,11 +169,11 @@ pub mod pallet {
...
 
-		/// Pool was created. [who, asset a, asset b, initial shares amount]
-		PoolCreated(T::AccountId, AssetId, AssetId, Balance),
+		/// Pool was created. [who, asset a, asset b, initial shares amount, pool id]
+		PoolCreated(T::AccountId, AssetId, AssetId, Balance, AssetId),
 
-		/// Pool was destroyed. [who, asset a, asset b]
-		PoolDestroyed(T::AccountId, AssetId, AssetId),
+		/// Pool was destroyed. [who, asset a, asset b, pool id]
+		PoolDestroyed(T::AccountId, AssetId, AssetId, AssetId),
```
jak-pan, line 176: this is share token id not exactly pool id -> we should add pool address as well

enthusiastmartin: added

jak-pan (approved):

PR #304, opened 2021-09-13
feat: governance launch

jak-pan:


vgantchev:
What is `DOLLARS` and how do they translate to HDX?

Check out the parameters that we announced: https://docs.hydradx.io/democracy_intro
I spotted some inconsistencies, if we deviate from this I will update the docs:
* EnactmentPeriod (is 6 days, should be 3)
* MaxProposals (is 30, should be 100)

vgantchev:
EnactmentPeriod = 3 => 3 months lockup for highest voting conviction (32*3=96d)
EnactmentPeriod = 6 => 6 months lockup for highest voting conviction (32*6=192d)

I think 6 would be fine as well

jak-pan:
> EnactmentPeriod = 3 => 3 months lockup for highest voting conviction (32 * 3=96d)
> EnactmentPeriod = 6 => 6 months lockup for highest voting conviction (32 * 6=192d)
> 
> I think 6 would be fine as well

Yes we'll keep the 6 for now.

PR #342, opened 2021-12-14
chore: update weights and reference weights

enthusiastmartin:
Weights updated for xyk,exchange and multi payment pallet together with reference weights used in check perf script.


.maintain/bench-check/hydradx-bench-data.json
```diff
@@ -1,23 +1,23 @@
...
     },
     "exchange": {
-        "sell_intention": 109.0,
-        "buy_intention": 112.0,
-        "sell_extrinsic": 121.0,
-        "on_finalize_for_one_sell_extrinsic": 163.0,
-        "buy_extrinsic": 120.0,
-        "on_finalize_for_one_buy_extrinsic": 179.0
+        "sell_intention": 102.0,
+        "buy_intention": 142.0,
```
jak-pan, line 11: Is really buy that much more expensive?

enthusiastmartin: nope. mistake.  it is fixed. it is the same as sell.

PR #356, opened 2022-01-25
fix: update AMI image with new key

fakirAyoub:
Update AMI image with valid token

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,5 @@
 name: build-and-deploy
 on:
-  push:
   workflow_dispatch:
   pull_request:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -1,5 +1,6 @@
 name: build-and-deploy
 on:
+  push:
```
lumir-mrkva, line 3: dont trigger workflow on push

lumir-mrkva (changes requested):

lumir-mrkva:
also the pr has to semantically named https://github.com/zeke/semantic-pull-requests

.github/workflows/tests.yml
lumir-mrkva: @fakirAyoub this is still an issue

fakirAyoub: @lumir-mrkva change done

lumir-mrkva (approved):

PR #358, opened 2022-01-31
chore: upgrade snakenet to polkadot v0.9.15

### LINKED ISSUE #357: upgrade snakenet solochain to polkadot 0.9.15
storage should be as close as possible to parachain for the following data migration

enthusiastmartin:
fixes #357


enthusiastmartin:
@martinfridrich if you can have a look at the chain spec, that would be great.

node/src/command.rs
```diff
@@ -92,24 +92,19 @@ impl SubstrateCli for Cli {
...
-
 /// Parse and run command line arguments
 pub fn run() -> sc_cli::Result<()> {
 	let cli = Cli::from_args();
-	set_ss58();
 
 	match &cli.subcommand {
 		None => {
-			let run_testing_runtime = cli.run.runtime.is_testing_runtime();
+			//let runner = cli.create_runner(&cli.run.base.normalize())?;
```
lumir-mrkva, line 101: remove

node/src/command.rs
```diff
@@ -92,24 +92,19 @@ impl SubstrateCli for Cli {
...
 			let runner = cli.create_runner(&cli.run.base)?;
+
 			runner.run_node_until_exit(|config| async move {
 				match config.role {
-					Role::Light => service::build_light(config, run_testing_runtime),
-					_ => service::build_full(config, run_testing_runtime).map(|full| full.task_manager),
+					Role::Light => panic!("Light is not supported"),
+					_ => service::build_full(config, false).map(|full| full.task_manager),
```
lumir-mrkva, line 107: now, how do you run testing runtime? :)

primitives/src/constants.rs
```diff
@@ -0,0 +1,39 @@
...
+	pub const MAX_IN_RATIO: u128 = 3;
+
+	/// Trading limit
+	pub const MIN_TRADING_LIMIT: Balance = 1000;
+
+	/// Minimum pool liquidity
+	pub const MIN_POOL_LIQUIDITY: Balance = 1000;
+
+	/// We allow for
+	pub const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND / 2;
```
lumir-mrkva, line 38: 0.5 sec is limitation of parachain, solochain is fine with 2 sec processing time

```suggestion
	pub const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND * 2;
```

pallets/claims/src/lib.rs
```diff
@@ -231,7 +233,7 @@ where
 		_len: usize,
 	) -> TransactionValidity {
 		match call.is_sub_type() {
-			Some(Call::claim(signature)) => match Pallet::<T>::validate_claim(who, signature) {
+			Some(Call::claim{ethereum_signature}) => match Pallet::<T>::validate_claim(who, ethereum_signature) {
```
lumir-mrkva, line 236: why was this renamed? isn't it breaking any API? https://claim.hydradx.io

runtime/common/src/lib.rs
```diff
@@ -282,9 +284,78 @@ parameter_types! {
...
+	}
+}
+
+
+use pallet_staking::BenchmarkingConfig;
+
+pub struct StakingBenchmarkingConfig;
+
+impl BenchmarkingConfig for StakingBenchmarkingConfig{
+	type MaxValidators = ConstU32<100>;
```
lumir-mrkva, line 342: is this accurate for our case? we should have like 165 validators with 64 nominators each. anyway it requires rebench probably

runtime/hydra-dx/src/lib.rs
```diff
@@ -145,8 +160,8 @@ pub fn native_version() -> NativeVersion {
 
 pub struct BaseFilter;
 
-impl Filter<Call> for BaseFilter {
-	fn filter(call: &Call) -> bool {
+impl Contains<Call> for BaseFilter {
+	fn contains(call: &Call) -> bool {
 		match call {
 			Call::Authorship(_)
```
lumir-mrkva, line 166: note that rest of the calls dont have to be listed in match and you can just match rest (_) as true

runtime/hydra-dx/src/lib.rs
```diff
@@ -545,29 +596,34 @@ parameter_types! {
...
 	type SolutionImprovementThreshold = MinSolutionScoreBump;
-	type MinerMaxIterations = MinerMaxIterations;
 	type MinerMaxWeight = MinerMaxWeight;
 	type MinerTxPriority = MultiPhaseUnsignedPriority;
 	type DataProvider = Staking;
-	type OnChainAccuracy = Perbill;
-	type CompactSolution = NposCompactSolution16;
-	type Fallback = Fallback;
-	type BenchmarkingConfig = ();
+	type Fallback = pallet_election_provider_multi_phase::NoFallback<Self>;
```
lumir-mrkva, line 625: fallback should be same as before (on-chain)

pallets/claims/src/migration.rs
```diff
@@ -16,13 +16,17 @@
...
 
 pub fn import_initial_claims<T: Config>(claims_data: &[(&'static str, Balance)]) -> frame_support::weights::Weight {
-	let version = <Pallet<T> as GetPalletVersion>::storage_version();
-	if version == None {
+	let version = <Pallet<T> as GetStorageVersion>::current_storage_version();
+
+	//TODO: it used to return None if it did not exists, so what do now ?
+	// I believe
+
+	if version == 0 {
```
lumir-mrkva, line 29: check storage in mainnet, make sure that we don't run this migration again. theoretically, this migration could be removed already, with all data related

node/src/service.rs
```diff
@@ -124,14 +132,15 @@ pub fn new_partial<RuntimeApi, Executor>(
...
+			Option<TelemetryWorkerHandle>,
 		),
 	>,
 	ServiceError,
 >
 where
 	RuntimeApi: ConstructRuntimeApi<Block, FullClient<RuntimeApi, Executor>> + Send + Sync + 'static,
 	RuntimeApi::RuntimeApi: RuntimeApiCollection<StateBackend = sc_client_api::StateBackendFor<FullBackend, Block>>,
-	Executor: NativeExecutionDispatch + 'static,
+	Executor: NativeExecutionDispatch + 'static, //+ sp_core::traits::CodeExecutor + sc_executor::RuntimeVersionOf ,
```
lumir-mrkva, line 143: remove comment

pallets/claims/src/tests.rs
```diff
@@ -112,11 +113,12 @@ fn signed_extention_success() {
 	new_test_ext().execute_with(|| {
 		let signature = hex!["5b2b46b0162f4b4431f154c4b9fc5ba923690b98b0c2063720799da54cb35a354304102ede62977ba556f0b03e67710522d4b7523547c62fcdc5acea59c99aa41b"];
 
-		let call = <crate::Call<Test>>::claim(EcdsaSignature(signature)).into();
+		//let call = <pallet_claims::Call<Test>>::claim(EcdsaSignature(signature));//.into();
```
lumir-mrkva, line 116: remove

runtime/hydra-dx/src/tests.rs
```diff
@@ -15,9 +14,9 @@ fn full_block_cost() {
 
 	let max_weight = BlockWeights::get().get(DispatchClass::Normal).max_total.unwrap_or(1);
 	let weight_fee = WeightToFee::calc(&max_weight);
-	assert_eq!(weight_fee, 19149775500000);
+	assert_eq!(weight_fee, 120000000000000);
```
lumir-mrkva, line 17: does this mean that we're basically increasing-price 10 fold?

lumir-mrkva (changes requested):
- [ ] bump versions
- [ ] reuse or remove asset registry, xyk and exchange

pallets/claims/src/lib.rs
enthusiastmartin: It was not renamed . it is actually called ethereum_signature in the claim extrinsic.

the reason it had to be renamed in the match statement  is that it had to be changed to use  struct pattern syntax. 

so instead of using this:
```rust
Some(Call::claim{ethereum_signature: signature}) => match Pallet::<T>::validate_claim(who, signature)
```

i went with just eithereium signature.

runtime/hydra-dx/src/lib.rs
enthusiastmartin: Well, Fallback strategy where it originally took the OnChain fallback has been removed. 

any tip how to do ?

runtime/hydra-dx/src/tests.rs
enthusiastmartin: the difference is because of removal of extra weight from ExtrinsicBaseWeight.

If i  add back the extra weight - it is the same result 19149775500000.

Which is kind of surprising 😕

runtime/common/src/lib.rs
enthusiastmartin: Set 165 and 64.

runtime/hydra-dx/src/lib.rs
enthusiastmartin: leaving to NoFallback.

pallets/claims/src/migration.rs
enthusiastmartin: removed on_runtime_upgrade hook.

node/src/command.rs
Roznovjak: done

lumir-mrkva (approved):

PR #363, opened 2022-02-15
chore: upgrade snakenet to v0.9.16 

### LINKED ISSUE #362: Upgrade Snakenet to Polkadot 0.9.16

enthusiastmartin:
based on #358 
fixes #362


runtime/common/src/lib.rs
```diff
@@ -232,6 +226,7 @@ parameter_types! {
 parameter_types! {
 	pub const ProposalBond: Permill = Permill::from_percent(3);
 	pub const ProposalBondMinimum: Balance = 100 * DOLLARS;
+	pub const ProposalBondMaximum: Balance = 50 * DOLLARS;
```
lumir-mrkva, line 229: probably should be more than minimum

lumir-mrkva:
spams this message when syncing, but still proceeds to sync

```
Listener: Ignoring invalid protocol:  due to A protocol (name) is invalid.    
```

enthusiastmartin:
> spams this message when syncing, but still proceeds to sync
> 
> ```
> Listener: Ignoring invalid protocol:  due to A protocol (name) is invalid.    
> ```



> spams this message when syncing, but still proceeds to sync
> 
> ```
> Listener: Ignoring invalid protocol:  due to A protocol (name) is invalid.    
> ```

protocol name is for grandpa config. currently set to default value. What name can it be set ? any hints ?

lumir-mrkva:
> > spams this message when syncing, but still proceeds to sync
> > ```
> > Listener: Ignoring invalid protocol:  due to A protocol (name) is invalid.    
> > ```
> 
> protocol name is set for grandpa config. currently set to default value. What name can it be set to ? any hints ?

i would use "standard procotol name" construction as is used in polkadot https://github.com/paritytech/polkadot/blob/43b390f289fbe2054fbc86dd91f79dee314dc689/node/service/src/lib.rs#L777

node/src/service.rs
```diff
@@ -491,7 +496,7 @@ where
 		keystore,
 		local_role: role,
 		telemetry: telemetry.as_ref().map(|x| x.handle()),
-		protocol_name: Default::default(),
+		protocol_name: grandpa_protocol_name,
```
lumir-mrkva, line 499: hmm, this doesn't sync at all now

```
2022-02-17 13:37:37 add_set_reserved_peer with unknown protocol: /0xd2a6…86fc/grandpa/1  
```

we are probably already using some protocol name

lumir-mrkva: ok, looks like fallback kicked in and it synces like before, but still logging same message

lumir-mrkva:
pls update deps to the latest polkadot-0.9.16 / release-0.9.16 revisions, parity cherrypicked new stuff there

node/src/service.rs
enthusiastmartin: Is it Worth effort to investigate further ?

lumir-mrkva: yes, even if its just visual it will be reported by out validators

lumir-mrkva (approved):

PR #381, opened 2022-03-17
feat: added vesting account and lowered minimal vested transfer

vgantchev:
Related to https://github.com/galacticcouncil/HydraDX-node/issues/373 and https://github.com/galacticcouncil/Basilisk-crowdloan-ui/issues/30

* adds vesting account to whitelist for vested_transfer
* lowers MinVestedTransfer to allow for vestings under existential deposit

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -250,7 +250,7 @@ parameter_types! {
 
 // pallet vesting
 parameter_types! {
-	pub MinVestedTransfer: Balance = 1_000;
+	pub MinVestedTransfer: Balance = 100;
 	pub const MaxVestingSchedules: u32 = 100;
 	pub const VestingPalletId: PalletId = PalletId(*b"py/vstng");
 }
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -252,6 +252,7 @@ parameter_types! {
 parameter_types! {
 	pub MinVestedTransfer: Balance = NativeExistentialDeposit::get();
 	pub const MaxVestingSchedules: u32 = 100;
+	pub const VestingPalletId: PalletId = PalletId(*b"py/vstng");
 }
 
 // pallet proxy
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -250,7 +250,7 @@ parameter_types! {
 
 // pallet vesting
 parameter_types! {
-	pub MinVestedTransfer: Balance = NativeExistentialDeposit::get();
+	pub MinVestedTransfer: Balance = 1_000;
 	pub const MaxVestingSchedules: u32 = 100;
 	pub const VestingPalletId: PalletId = PalletId(*b"py/vstng");
 }
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -24,8 +24,8 @@
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 use codec::{Decode, Encode};
-use scale_info::TypeInfo;
 use frame_system::{EnsureRoot, RawOrigin};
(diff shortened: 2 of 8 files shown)


Roznovjak:
I believe that tests are not necessary (similar [test](https://github.com/open-web3-stack/open-runtime-module-library/blob/9e041dc9d213f843b18b3008f32f3acabb287dcb/vesting/src/tests.rs#L65) is here).

I think that we want to use `TreasuryPalletId` instead of `VestingPalletId`. ([here](https://github.com/galacticcouncil/HydraDX-crowdloan-ui/blob/9cbab46cdd9fd1666c0495d403b649dd2114c7b1/scripts/batch-vesting/batchVestingCall.js#L53))

Is there a reason why is lower `MinVestedTransfer` required?

vgantchev:
The tests are needed because we have modified in runtime which Origin of calls we accept. As you can see, we are not testing the default vesting functionality, but simply Ok() or Err()

We discussed with @lumir-mrkva that we don't want to send directly from treasury, but mint to a dedicated account instead

The lower `MinVestedTransfer` is required because some of the vestings contain sum of decimal amounts which can be under the existential deposit

Roznovjak:
> The tests are needed because we have modified in runtime which Origin of calls we accept. As you can see, we are not testing the default vesting functionality, but simply Ok() or Err(BadOrigin)
> 
> We discussed with @lumir-mrkva that we don't want to send directly from treasury, but mint to a dedicated account instead
> 
> The lower `MinVestedTransfer` is required because some of the vestings contain sum of decimal amounts which can be under the existential deposit

Thanks for the clarification.
But I still don't think that the tests are necessary :) If we want to test that it's configured correctly, testing `EnsureOrigin` implementation on `GalacticCouncilOrRoot` should be enough. But it's definitely not bad to have more tests. @lumir-mrkva what do you think?

lumir-mrkva (approved):
i like what account did you choose for vestings

@Roznovjak I think we need more tests for our runtime config, it is definitely better to config something and add a test for it than just try it once with running dev node. tests inside of the pallet itself won't test our config combination and attached types

vgantchev:
An example why we should test every line of our runtime :)

Imagine instead of this:

```
if caller == primitives::constants::chain::GALACTIC_COUNCIL_ACCOUNT.into()
					|| caller == VestingPalletId::get().into_account()
```

Someone would commit this:
```
if caller == primitives::constants::chain::GALACTIC_COUNCIL_ACCOUNT.into()
					|| VestingPalletId::get().into_account()
```

Without tests, this would pass all checks and build. 

During the next vested_transfer, the following will happen to the chain:
```
2022-03-19 17:04:54 [Parachain] panicked at '`AccountId` type is never greater than 32 bytes; qed: Error', /Users/cl0w/.cargo/git/checkouts/substrate-7e08433d4c370a21/19162e4/primitives/runtime/src/traits.rs:1286:14
```

Roznovjak:
I agree that we should test the structs we use in our configs.
But I think a better way is to test it directly, and make tests independent of pallets.
Something like this: 
```
#[test]
fn galactic_council_or_vesting_or_root_should_work() {
	assert_ok!(GalacticCouncilOrVestingOrRoot::ensure_origin(crate::Origin::signed(primitives::constants::chain::GALACTIC_COUNCIL_ACCOUNT.into())));
	assert_ok!(GalacticCouncilOrVestingOrRoot::ensure_origin(crate::Origin::signed(VestingPalletId::get().into_account())));
	assert_ok!(GalacticCouncilOrVestingOrRoot::ensure_origin(crate::Origin::root()));
	assert_eq!(GalacticCouncilOrVestingOrRoot::ensure_origin(crate::Origin::none()).is_err(),
		true);
	assert_eq!(GalacticCouncilOrVestingOrRoot::ensure_origin(crate::Origin::signed(ALICE)).is_err(),
		true);
}
```

PR #396, opened 2022-04-04
feat: Omnipool

enthusiastmartin:
Omnipool implementation. 

TODO:

- [x] Imbalances/TVL updates 
- [x] Few clarifications about math ( marked as TODO: ask colin ).
- [x] Tests
- [x] Benchmarks


pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,980 @@
...
+	#[pallet::storage]
+	/// State of an asset in the omnipool
+	pub(super) type Assets<T: Config> = StorageMap<_, Blake2_128Concat, T::AssetId, AssetState<T::Balance>>;
+
+	#[pallet::storage]
+	/// Imbalance of hub asset
+	pub(super) type HubAssetImbalance<T: Config> = StorageValue<_, SimpleImbalance<T::Balance>, ValueQuery>;
+
+	#[pallet::storage]
+	/// Total TVL
```
green-jay, line 170: No [RAS Syndrome](https://en.wikipedia.org/wiki/RAS_syndrome) please :)

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,980 @@
...
+		StorageMap<_, Blake2_128Concat, T::PositionInstanceId, Position<T::Balance, T::AssetId>>;
+
+	#[pallet::storage]
+	/// Position IDs sequencer
+	pub(super) type PositionInstanceSequencer<T: Config> = StorageValue<_, u32, ValueQuery>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		TokenAdded(T::AssetId, T::Balance, Price),
```
green-jay, line 189: Named params in events are nicer to read

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,980 @@
...
+		NotAllowed,
+		/// Signed account is not owner of position instance.
+		Forbidden,
+		/// Asset weight cap has been exceeded.
+		AssetWeightCapExceeded,
+		/// TVL cap has been exceeded
+		TVLCapExceeded,
+		/// Asset is not registered in asset registry
+		AssetNotRegistered,
+		///
```
green-jay, line 227: missing doc

pallets/omnipool/src/tests/scenario_04.rs
```diff
@@ -0,0 +1,114 @@
+use super::*;
+
+/// Auto-generated test
```
green-jay, line 3: would be helpful to have some docs over how these tests were generated and what is the scenario about

pallets/omnipool/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	 */
+
+	pub(super) fn sub<T: Config>(mut self, amount: Balance) -> Option<Self> {
+		if self.negative {
+			let result = self.value.checked_add(&amount);
+
+			if result.is_some() {
+				self.value = result.unwrap();
```
green-jay, line 82: replace unwrap with expect("comment") or line comment?

green-jay (commented):
had just a brief look, will look into the logic at later point

pallets/omnipool/src/types.rs
enthusiastmartin: refactored in different way already

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,980 @@
...
+#[macro_export]
+macro_rules! ensure_asset_in_pool {
+	( $x:expr, $y:expr $(,)? ) => {{
+		if !Assets::<T>::contains_key($x) {
+			return Err($y.into());
+		}
+	}};
+}
+
+type NFTClassIdOf<T> = <<T as Config>::NFTHandler as Inspect<<T as frame_system::Config>::AccountId>>::ClassId;
```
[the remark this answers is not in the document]
enthusiastmartin: what you mean exactly ? 

oknmipool is nft impl agnostic, you can plug in whatever implementation you'd like.

this type is just an alias to type used in given implementation.

apopiak (commented):
in-progress review

pallets/omnipool/src/types.rs
```diff
@@ -0,0 +1,628 @@
...
+}
+
+#[macro_export]
+macro_rules! update_value {
+	( $x:expr, $y:expr) => {{
+		match &$y {
+			BalanceUpdate::Increase(amount) => $x.checked_add(&amount),
+			BalanceUpdate::Decrease(amount) => $x.checked_sub(&amount),
+		}
+	}};
```
apopiak, line 255: I notice that you solved a similar problem with `SimpleBalance` and `BalanceUpdate` (potentially negative number for unsigned ints) in different ways.
I would suggest building one abstraction. My personal opinion is that enum in combination with trait impl is the better combination (as opposed to using either of struct representation or this kind of macro for calculations).
Have you tried expressing this macro as an `Add` implementation?

apopiak: playground link for the add implementation:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c8da265a0ec8b4a085a576eeb3dbdc5e

apopiak (commented):
As you're aware:
Needs docs and benchmarks :-)

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1121 @@
...
+					);
+				}
+				Self::stable_asset()?
+			} else {
+				// Trying to add preferred stable asset.
+				// This can happen only once , since it is first token to add to the pool.
+				(
+					amount,
+					initial_price.checked_mul_int(amount).ok_or(Error::<T>::Overflow)?,
+				)
```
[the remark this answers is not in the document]
enthusiastmartin: Definitely. I like the initialization in migration.

The thing here is  the assumption that the stable asset must be already transferred to Hydra before omnipool pallet is introduced. 

@jak-pan Is this safe assumption ?

pallets/omnipool/Cargo.toml
```diff
@@ -0,0 +1,75 @@
...
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false , optional = true}
+pallet-balances = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false, optional = true}
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false, optional = true}
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false , optional = true}
+sp-tracing = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false, optional = true}
+
+[dev-dependencies]
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false}
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false}
+sp-tracing = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false}
```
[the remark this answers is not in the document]
enthusiastmartin: removed for now as not used.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1121 @@
...
+		/// The units in which we handle balances.
+		type Balance: Member
+			+ Parameter
+			+ AtLeast32BitUnsigned
+			+ Default
+			+ Copy
+			+ MaybeSerializeDeserialize
+			+ MaxEncodedLen
+			+ FixedPointOperand
+			// TODO: the following from/into is due to use of FixedU128, which internally uses u128.
```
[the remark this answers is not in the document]
enthusiastmartin: Comments regarding this little inconvenience are updated and the type of price stored will be change to FixedU128 with 0.9.17 upgrade

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1121 @@
...
+			let mut position = Positions::<T>::get(position_id).ok_or(Error::<T>::PositionNotFound)?;
+
+			ensure!(position.shares >= amount, Error::<T>::InsufficientShares);
+
+			let asset_id = position.asset_id;
+
+			let mut asset_state = Assets::<T>::get(asset_id).ok_or(Error::<T>::AssetNotFound)?;
+
+			let state_changes = calculate_remove_liquidity_state_changes::<T>(&asset_state, amount, &position)
+				.ok_or(Error::<T>::Overflow)?;
```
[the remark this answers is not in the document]
enthusiastmartin: Nice, thanks. I was not aware.

pallets/omnipool/src/tests/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+		assert_eq!(Tokens::free_balance($y, &$x), $z);
+	}};
+}
+
+#[macro_export]
+macro_rules! check_balance {
+	( $x:expr, $y:expr, $z:expr) => {{
+		assert_eq!(Tokens::free_balance($y, &$x), $z);
+	}};
+}
```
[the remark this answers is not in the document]
enthusiastmartin: yes, one is not used at all. and it was removed.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1121 @@
...
+		///
+		/// Emits `LiquidityAdded` event when successful.
+		///
+		#[pallet::weight(<T as Config>::WeightInfo::add_liquidity())]
+		#[transactional]
+		pub fn add_liquidity(origin: OriginFor<T>, asset: T::AssetId, amount: T::Balance) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(
+				amount >= T::MinimumPoolLiquidity::get(),
```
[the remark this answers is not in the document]
enthusiastmartin: Yes, it would.. But the min limit is really low - for xyk pool is set to 1000.  Therefore we made it same for each asset

This would apply to min trading limit too. It is same idea, where it is set to 1000. 

We might revisit this if needed.

pallets/omnipool/src/mock.rs
```diff
@@ -0,0 +1,412 @@
...
+	pub fn with_endowed_accounts(mut self, accounts: Vec<(u64, AssetId, Balance)>) -> Self {
+		self.endowed_accounts = accounts;
+		self
+	}
+	pub fn add_endowed_accounts(mut self, account: (u64, AssetId, Balance)) -> Self {
+		self.endowed_accounts.push(account);
+		self
+	}
+	pub fn with_registered_asset(self, asset: AssetId) -> Self {
+		REGISTERED_ASSETS.with(|v| {
```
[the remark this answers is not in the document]
enthusiastmartin: Yes, this makes sense.

btw. If you have an idea, how to change pallets constants another way ( not using the global state ) - let me know.

pallets/omnipool/src/types.rs
enthusiastmartin: The issue i have with one abstraction is that `SimpleImbalance` is stored in storage.

and imbalance represents a number - it can be positive or negative. 

`BalanceUpdate` is used to represent that something is decreased/increased by that amount.

Using `BalanceUpdate` in storage  instead could be confusing to read imo, where you'd see `BalanceUpdate::Decrease(amount)`

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1121 @@
...
+		} else {
+			Ok(())
+		}
+	}
+
+	/// Update total tvl and asset tvl. Write new total tvl value to storage, and mutate asset tvl accordingly.
+	fn update_tvl(asset_state: &mut AssetState<T::Balance>) -> DispatchResult {
+		let (stable_asset_reserve, stable_asset_hub_reserve) = Self::stable_asset()?;
+
+		if stable_asset_reserve != T::Balance::zero() && stable_asset_hub_reserve != T::Balance::zero() {
```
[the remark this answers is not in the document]
enthusiastmartin: Well, 

`FixedU128::from( (100,0) )`  panics. 

So, it is also because of that. 

now that i think of it - i use this `from( (N,D) )` in other places too, without explicitly checking that the D != 0. 

What do you think ?

pallets/omnipool/src/lib.rs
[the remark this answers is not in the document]
enthusiastmartin: Refactored already. and extracted the initialize logic into separate extrinsic. 

as mentioned, init in migration would need some assumption, which cannot be guaranteed atm.

`initialize_pool `call can be called only by root. And add_token origin is configurable `AddTokenOrigin`.

pallets/omnipool/src/lib.rs
```diff
@@ -922,25 +1018,26 @@ impl<T: Config> Pallet<T> {
...
 	) -> DispatchResult {
 		<HubAssetLiquidity<T>>::try_mutate(|liquidity| -> DispatchResult {
 			match delta_amount {
 				BalanceUpdate::Increase(amount) => {
-					*liquidity = liquidity.checked_add(amount).ok_or(Error::<T>::Overflow)?;
+					*liquidity = liquidity.checked_add(amount).ok_or(ArithmeticError::Overflow)?;
 				}
 				BalanceUpdate::Decrease(amount) => {
-					*liquidity = liquidity.checked_sub(amount).ok_or(Error::<T>::Overflow)?;
+					*liquidity = liquidity.checked_sub(amount).ok_or(ArithmeticError::Overflow)?;
```
apopiak, line 1040: nit: use `Underflow`?

pallets/omnipool/src/lib.rs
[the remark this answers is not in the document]
apopiak: I think it makes sense to use [checked_from_rational](https://github.com/paritytech/substrate/blob/4d3d480d8149fbca2834d05cc050a01f9a25afbf/primitives/arithmetic/src/fixed_point.rs#L142) whenever the denominator could be zero and then return `ArithmeticError::DivisionByZero`

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1121 @@
...
+		/// Emits `BuyExecuted` event when successful.
+		///
+		#[pallet::weight(<T as Config>::WeightInfo::buy())]
+		#[transactional]
+		pub fn buy(
+			origin: OriginFor<T>,
+			asset_out: T::AssetId,
+			asset_in: T::AssetId,
+			amount: T::Balance,
+			max_limit: T::Balance,
```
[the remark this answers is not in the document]
enthusiastmartin: changed to 
- `max_sell_amount` in `buy`
- `min_buy_amount` in `sell`

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1121 @@
...
+				amount >= T::MinimumPoolLiquidity::get(),
+				Error::<T>::InsufficientLiquidity
+			);
+
+			ensure!(
+				T::Currency::free_balance(asset, &who) >= amount,
+				Error::<T>::InsufficientBalance
+			);
+
+			let mut asset_state = Assets::<T>::get(asset).ok_or(Error::<T>::AssetNotFound)?;
```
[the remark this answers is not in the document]
enthusiastmartin: refactored to use try_mutate. 

however, it does not change much right now, because the scope is basically the same, only in the try_mutate. 
this is due to the fact, that mutable reference is still needed at the very end when TVL is updated. 

tvl updates will be refactored ( made part of the math) and then, the scope of asset state can be reduced as mutable reference might be no longer needed.

pallets/omnipool/src/mock.rs
```diff
@@ -0,0 +1,409 @@
...
+
+parameter_types! {
+	pub const HDXAssetId: AssetId = HDX;
+	pub const LRNAAssetId: AssetId = LRNA;
+	pub const DAIAssetId: AssetId = DAI;
+	pub const PosiitionClassId: u32= 1000;
+
+	pub ProtocolFee: (u32,u32) = FeeProtocol::get();
+	pub AssetFee: (u32,u32) = FeeAsset::get();
+	pub AssetWeightCap: (u32,u32) = WeightCap::get();
```
apopiak, line 177: I think this is some duplication? You should be able to write 
```Rust
parameter_types! {
	pub AssetWeightCap: (u32,u32) = ASSET_WEIGHT_CAP.with(|v| *v.borrow());
}
```

pallets/omnipool/src/mock.rs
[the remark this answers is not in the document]
apopiak: I don't think I have any elegant solutions either. (Just stuff like creating storage entries.)

pallets/omnipool/src/mock.rs
apopiak: There are also storage parameter_types btw. (see [examples here](https://paritytech.github.io/substrate/master/frame_support/macro.parameter_types.html#examples))

pallets/omnipool/src/types.rs
enthusiastmartin: > playground link for the add implementation: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c8da265a0ec8b4a085a576eeb3dbdc5e

replaced the update_value macro with Add implementation as suggested.

pallets/omnipool/src/benchmarks.rs
```diff
@@ -1 +1,115 @@
...
+
+	initialize_pool{
+		let stable_amount: T::Balance = T::Balance::from(1_000_000_000_000_000u128);
+		let native_amount: T::Balance = T::Balance::from(1_000_000_000_000_000u128);
+		let stable_price: FixedU128= FixedU128::from((1,2));
+		let native_price: FixedU128= FixedU128::from(1);
+
+	}: _(RawOrigin::Root, stable_amount, native_amount, stable_price, native_price)
+	verify {
+	}
```
apopiak, line 46: nitpick: might be good to verify the stable asset and native asset "arrived" in storage?

pallets/omnipool/src/lib.rs
```diff
@@ -178,14 +178,18 @@ pub mod pallet {
 	/// Total amount of hub asset reserve. It equals to sum of hub_reserve of each asset in omnipool
 	pub(super) type HubAssetLiquidity<T: Config> = StorageValue<_, T::Balance, ValueQuery>;
 
+	#[pallet::storage]
+	/// Tradable state of hub asset.
+	pub(super) type HubAssetTradability<T: Config> = StorageValue<_, Tradable, ValueQuery>;
```
apopiak, line 183: Optimization: I suspect these are mostly accessed together and could be combined?

enthusiastmartin: I was thinking the same. but i discarded it for some reason. Now i need to remember the reason 🤔 

let me check.

enthusiastmartin: I had a look again. I don't think there is a reason why it cannot be done.

I decided to this way - because the Tradable state of hub asset is only needed when you trade Hub asset - which is currently not allowed to buy, only to sell. And to obtain LRNA, you can only obtain hub asset after remove_liquidity and only if the price changes in specific way. 

But HubAssetLiquidity is used almost in all calls. so the state would be read alongside the total hub liquidity every time even it is not needed.

Although the trade state is just enum and probably would not make such difference in the weight, I think leaving these this piece of information of hub asset separated is fine. 

Is there anything else you can think of to consider ?

lumir-mrkva:
target should be changed to hydra-parachain

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1409 @@
...
+				Self::update_tvl(&state_changes.asset.delta_tvl)?;
+
+				// Total hub asset liquidity update
+				Self::update_hub_asset_liquidity(
+					&state_changes.asset.delta_hub_reserve,
+					HubAssetIssuanceUpdate::AdjustSupply,
+				)?;
+
+				// Storage update - asset state
+				<Assets<T>>::insert(asset, asset_state);
```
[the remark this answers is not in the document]
enthusiastmartin: yes. correct. not needed.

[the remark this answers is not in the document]
Roznovjak: The same "issue" is also in remove_liquidity, sell_hub_asset and buy_asset_for_hub_asset.

pallets/omnipool/README.md
```diff
@@ -0,0 +1,51 @@
+# Omnipool pallet
+
+Omnipool implementation
+
+## Overview
+
+Omnipool is type of AMM where all assets are pooled together into one single pool.
+
+Liquidity provider can provide any aset of their choice to the Omnipool and in return
```
dmoka, line 9: asset*

(let me know if we do not nitpick typos, I do not want to do it unnecessarily :))

pallets/omnipool/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+Omnipool is type of AMM where all assets are pooled together into one single pool.
+
+Liquidity provider can provide any aset of their choice to the Omnipool and in return 
+they will receive pool shares for this single asset.
+
+The position is represented with a NFT token which saves the amount of shares distributed 
+and the price of the asset at the time of provision.
+
+For traders this means that tehy can benefit from the fill asset position
```
dmoka, line 15: they*

pallets/omnipool/README.md
```diff
@@ -0,0 +1,51 @@
...
+* **Hub Asset:** dedicated 'hub' token for trade executions (LRNA)
+* **Native Asset:** governance token
+
+## Assumptions
+
+Below are assumptions that must be held when using this pallet.
+
+* First two asset in pool must be Stable Asset and Native Asset. This must be achieved by calling
+  `initialize_pool` dispatchable.
+* Stable asset balance and native asset balance must be transffered to omnipool account manually.
```
dmoka, line 36: transferred*

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1337 @@
...
+				.checked_add(hub_asset_amount)
+				.ok_or(ArithmeticError::Overflow)?;
+			<Assets<T>>::insert(T::NativeAssetId::get(), native_subpool);
+		}
+		Ok(())
+	}
+
+	/// Update total hub asset liquidity and write new value to storage.
+	/// Update total issueance if AdjustSupply is specified.
+	#[require_transactional]
```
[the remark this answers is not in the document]
apopiak: We need to use `require_transactional` if they both:
+ mutate storage
+ and are also fallible

And could thus lead to a partial update, i.e. the first write succeeds, but the second write is skipped because the function errors out.

pallets/omnipool/src/benchmarks.rs
```diff
@@ -0,0 +1,276 @@
...
+		T: crate::pallet::Config
+	}
+
+	initialize_pool{
+		let stable_amount: Balance = 1_000_000_000_000_000u128;
+		let native_amount: Balance = 1_000_000_000_000_000u128;
+		let stable_price: FixedU128= FixedU128::from((1,2));
+		let native_price: FixedU128= FixedU128::from(1);
+
+	}: _(RawOrigin::Root, stable_price, native_price)
```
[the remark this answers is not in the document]
apopiak: This is the actual call to be benchmarked. This is a short-hand macro syntax for a block that calls the extrinsic with the same name as the benchmark with the parameters provided.

pallets/omnipool/src/lib.rs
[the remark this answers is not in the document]
dmoka: @apopiak got it now, thank you, Alex!

pallets/omnipool/README.md
enthusiastmartin: btw you can add a suggestion in the comment which can be directly committed. 

i would suggest to do that, so i can just click "commit change".

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1337 @@
...
+					*value = state;
+					Self::deposit_event(Event::TradableStateUpdated { asset_id, state });
+					Ok(())
+				})
+			} else {
+				Assets::<T>::try_mutate(asset_id, |maybe_asset| -> DispatchResult {
+					let asset_state = maybe_asset.as_mut().ok_or(Error::<T>::AssetNotFound)?;
+
+					asset_state.tradable = state;
+					Self::deposit_event(Event::TradableStateUpdated { asset_id, state });
```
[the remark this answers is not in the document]
enthusiastmartin: yes. i am aware. The tests of event still pending.

pallets/omnipool/src/benchmarks.rs
```diff
@@ -0,0 +1,276 @@
...
+use orml_traits::MultiCurrencyExtended;
+
+benchmarks! {
+	 where_clause {  where T::AssetId: From<u32>,
+		T::Currency: MultiCurrencyExtended<T::AccountId, Amount=i128>,
+		T: crate::pallet::Config
+	}
+
+	initialize_pool{
+		let stable_amount: Balance = 1_000_000_000_000_000u128;
```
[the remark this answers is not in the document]
enthusiastmartin: no longer needed. This is leftover from previous implementation. I will remove. 

i wonder why clippy does not pick this up - probably because it is in macro ?

pallets/omnipool/src/benchmarks.rs
```diff
@@ -0,0 +1,276 @@
...
+
+		crate::Pallet::<T>::initialize_pool(RawOrigin::Root.into(), stable_price,native_price)?;
+
+		// Register new asset in asset registry
+		let token_id = T::AssetRegistry::create_asset(&b"FCK".to_vec(), 1u128)?;
+
+		// Create account for token provider and set balance
+		let caller: T::AccountId = account("caller", 0, 1);
+
+		let token_price = FixedU128::from((1,5));
```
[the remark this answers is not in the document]
enthusiastmartin: Not possible - FixedU128::from is non-const so it is not possible to use in constants. We could try hacky workaround with lazy static loading. 

we could also try to have a init function.I'll think about it.

pallets/omnipool/src/tests/scenario_05.rs
```diff
@@ -0,0 +1,105 @@
+use super::*;
+
+/// Auto-generated test
+#[test]
+fn scenario_05() {
+	ExtBuilder::default()
+		.with_endowed_accounts(vec![
```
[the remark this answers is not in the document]
enthusiastmartin: Endowed basically means that a new account is created with some balance. 

eg. pallet-balances emits Endowed event when a new account is created.

pallets/omnipool/README.md
dmoka:
```suggestion
For traders this means that they can benefit from the fill asset position
```

pallets/omnipool/README.md
dmoka:
```suggestion
* Stable asset balance and native asset balance must be transferred to omnipool account manually.
```

pallets/omnipool/src/tests/scenario_05.rs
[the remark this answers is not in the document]
dmoka: thanks for the explanation, got it now. ;)

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1337 @@
...
+			let lp_position = Position::<Balance, T::AssetId> {
+				asset_id: asset,
+				amount,
+				shares: amount,
+				price: initial_price.into_inner(),
+			};
+
+			let instance_id = Self::create_and_mint_position_instance(&who)?;
+
+			<Positions<T>>::insert(instance_id, lp_position);
```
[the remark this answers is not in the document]
enthusiastmartin: there are checks of positon insereted vs expected position.  However, it is part of another test, which might be clear.

Tests are due to refactor to follow our new testing guidelines, so this will become more clear.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1337 @@
...
+				stable_asset,
+				asset == T::StableCoinAssetId::get(),
+			)
+			.ok_or(ArithmeticError::Overflow)?;
+
+			let new_asset_state = asset_state
+				.delta_update(&state_changes.asset)
+				.ok_or(ArithmeticError::Overflow)?;
+
+			let hub_reserve_ratio = FixedU128::checked_from_rational(
```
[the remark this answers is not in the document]
enthusiastmartin: not really a fan of helper method with just one line of code and used only once. 

i might refactor this, so it does not look so heavy though.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1337 @@
...
+			let delta_imbalance = Self::recalculate_imbalance(&new_asset_state, state_changes.delta_imbalance)
+				.ok_or(ArithmeticError::Overflow)?;
+			Self::update_imbalance(<HubAssetImbalance<T>>::get(), delta_imbalance)?;
+
+			Self::update_tvl(&state_changes.asset.delta_tvl)?;
+
+			Self::update_hub_asset_liquidity(&state_changes.asset.delta_hub_reserve)?;
+
+			// LP receives some hub asset
+			if state_changes.lp_hub_amount > Balance::zero() {
```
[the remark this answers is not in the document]
enthusiastmartin: there are definitely tests where this is 0. but again it might not be obvious as there is not dedicated test for this. 

but it wll be when tests are refactored.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1337 @@
...
+			Self::update_hub_asset_liquidity(&delta_hub_asset)?;
+
+			Self::update_hdx_subpool_hub_asset(state_changes.hdx_hub_amount)?;
+
+			Self::update_imbalance(current_imbalance, state_changes.delta_imbalance)?;
+
+			Self::set_asset_state(asset_in, new_asset_in_state);
+			Self::set_asset_state(asset_out, new_asset_out_state);
+
+			Self::deposit_event(Event::BuyExecuted {
```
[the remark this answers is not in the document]
enthusiastmartin: tests are being added to verify events.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+			Self::deposit_event(Event::PositionCreated {
+				position_id: instance_id,
+				owner: who.clone(),
+				asset,
+				amount,
+				shares: *state_changes.asset.delta_shares,
+				price: updated_asset_price,
+			});
+
+			T::Currency::transfer(
```
lumir-mrkva, line 653: everywhere we are assuming this transfers the exact amount of tokens to the protocol account, which doesn't have to be always a case and you are not in control of the actual implementation, we should always check that actual asset balances are consistent with whatever we are tracking in asset state (we have enough available balance)

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+
+				<Positions<T>>::remove(position_id);
+				T::NFTHandler::burn_from(&T::NFTClassId::get(), &position_id)?;
+
+				Self::deposit_event(Event::PositionDestroyed {
+					position_id,
+					owner: who.clone(),
+				});
+			} else {
+				<Positions<T>>::insert(position_id, updated_position);
```
lumir-mrkva, line 780: we should emit event that position was updated

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+		/// Sacrifice LP position in favor of pool.
+		///
+		/// A position is destroyed and liquidity owned by LP becomes pool owned liquidity.
+		///
+		/// Only owner of position can perform this action.
+		///
+		/// Emits `PositionDestroyed`.
+		#[pallet::weight(<T as Config>::WeightInfo::sacrifice_position())]
+		#[transactional]
+		pub fn sacrifice_position(origin: OriginFor<T>, position_id: T::PositionInstanceId) -> DispatchResult {
```
lumir-mrkva, line 804: what is the usecase for this? i could imaging if the initialization of omnipool would be creating position this could be then used, but still i would prefer to have option to contol the position

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+			);
+
+			let new_asset_in_state = asset_in_state
+				.delta_update(&state_changes.asset_in)
+				.ok_or(ArithmeticError::Overflow)?;
+			let new_asset_out_state = asset_out_state
+				.delta_update(&state_changes.asset_out)
+				.ok_or(ArithmeticError::Overflow)?;
+
+			T::Currency::transfer(
```
lumir-mrkva, line 1042: deltas of the actual transfers should be probably checked, this can be resolved easily by wrapping the transfer function

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+			+ TypeInfo;
+
+		/// Multi currency mechanism
+		type Currency: MultiCurrency<Self::AccountId, CurrencyId = Self::AssetId, Balance = Balance>;
+
+		/// Add token origin
+		type AddTokenOrigin: EnsureOrigin<Self::Origin>;
+
+		/// Origin to be able to change asset tradabilit state and initialize Omnipool.
+		type ManageOrigin: EnsureOrigin<Self::Origin>;
```
lumir-mrkva, line 138: `TechnicalOrigin` would be maybe better name for this, manage is way too broad, and implies control, we are not managing anyone assets

lumir-mrkva (changes requested):

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+			amount_in: *state_changes.asset.delta_hub_reserve,
+			amount_out: *state_changes.asset.delta_reserve,
+		});
+
+		Ok(())
+	}
+
+	/// Buy hub asset from the pool
+	/// Special handling of buy trade where asset out is Hub Asset.
+	fn buy_hub_asset(_who: &T::AccountId, _asset_in: T::AssetId, _amount: Balance, _limit: Balance) -> DispatchResult {
```
lumir-mrkva, line 1427: this should be documented in buy/sell function

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1462 @@
...
+				.ok_or(ArithmeticError::Overflow)?;
+			Self::update_imbalance(<HubAssetImbalance<T>>::get(), delta_imbalance)?;
+
+			Self::update_tvl(&state_changes.asset.delta_tvl)?;
+
+			Self::update_hub_asset_liquidity(&state_changes.asset.delta_hub_reserve)?;
+
+			Self::set_asset_state(asset, new_asset_state);
+
+			Self::deposit_event(Event::LiquidityAdded {
```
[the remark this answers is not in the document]
enthusiastmartin: which event you refer to ? there is only one LiquidityAdded event .

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+
+			Self::update_tvl(&BalanceUpdate::Increase(
+				native_asset_tvl
+					.checked_add(stable_asset_reserve)
+					.ok_or(ArithmeticError::Overflow)?,
+			))?;
+
+			<Assets<T>>::insert(T::StableCoinAssetId::get(), stable_asset_state);
+			<Assets<T>>::insert(T::NativeAssetId::get(), native_asset_state);
+
```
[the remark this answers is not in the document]
enthusiastmartin: initial liquidity is owner by protocol. asset state has a property called "protocol shares" where the amount of shares owned by protocol is stored.

pallets/omnipool/src/lib.rs
enthusiastmartin: @poliwop  can tell the exact motivation to have ability to give up a position in favor of pool. 

the initial liquidity is already owned by protocol, the sacrifice position adds shares of position to protocol shares.

pallets/omnipool/src/lib.rs
enthusiastmartin: it is mentioned in the buy / sell that buy/sell is special case. but it does not require to add the extrisnic docs part. it is the same math, there is nothing really different. 

the reason why you need to handle this separate because you deal with one asset state only and not two ( if you trade 2 other assets). 
that means, you load one asset state only. but math is the same only simplified, some parts are removed as no need to deal with second asset state.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1459 @@
...
+				&(&asset_out_state).into(),
+				amount,
+				T::AssetFee::get(),
+				T::ProtocolFee::get(),
+				current_imbalance.value,
+			)
+			.ok_or(ArithmeticError::Overflow)?;
+
+			ensure!(
+				T::Currency::free_balance(asset_in, &who) >= *state_changes.asset_in.delta_reserve,
```
lumir-mrkva, line 1075: this check can still pass if your free_balance is not transferable (you have some amount locked by vesting for example)

pallets/omnipool/src/lib.rs
lumir-mrkva: where are protocol shares tracked?

pallets/omnipool/src/lib.rs
lumir-mrkva: it should be documented that you cannot buy hub asset - every outcome (including errors) should be covered by documentation

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1462 @@
...
+		#[pallet::weight(<T as Config>::WeightInfo::add_token())]
+		#[transactional]
+		pub fn add_token(
+			origin: OriginFor<T>,
+			asset: T::AssetId,
+			initial_price: Price,
+			position_owner: T::AccountId,
+		) -> DispatchResult {
+			T::AddTokenOrigin::ensure_origin(origin)?;
+
```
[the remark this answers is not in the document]
enthusiastmartin: added some comments to divide some functions into 3 parts - preconditions, main, post - state updates

pallets/omnipool/src/lib.rs
enthusiastmartin: > where are protocol shares tracked?

in state of each asset.

pallets/omnipool/src/lib.rs
enthusiastmartin: it is already there that asset trade state of asset in must include SELL flag and asset out must include BUY flag.

pallets/omnipool/src/lib.rs
enthusiastmartin: replaced with ensure_can_withdraw where possible.

pallets/omnipool/src/lib.rs
enthusiastmartin: discussed "offline". Not something omnipool should be considering. It relies on correct MultiCurrency trait implemenation provided.

pallets/omnipool/src/lib.rs
lumir-mrkva: so how can user tell beforehand that this wont be allowed?

pallets/omnipool/src/lib.rs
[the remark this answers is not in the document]
lumir-mrkva: you are actually emitting `PositionCreated` just before ... and NFT ofc emits its own event as well

pallets/omnipool/src/lib.rs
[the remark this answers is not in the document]
lumir-mrkva: so you are finally ready to get rid of empty lines as well ? :)

pallets/omnipool/src/lib.rs
enthusiastmartin: HubAssetLiquidity has been removed.

.codecov.yml
```diff
@@ -0,0 +1,15 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down 
+  range: "70...100"
```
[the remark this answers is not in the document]
enthusiastmartin: this is global codecov file.

pallets/omnipool/src/lib.rs
enthusiastmartin: when FE loads asset state - there is the information what possible to do with asset (sell,buy,add/remove liquid), so it should display the information and allow/not allow an operation.

pallets/omnipool/src/lib.rs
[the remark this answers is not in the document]
enthusiastmartin: > so you are finally ready to get rid of the empty lines? :) the change just increased the size of the method even more

;-)  if we add it to our code guidelines, then i would consider it.

I might remove some where the statements do the same things ( 2 transfers, 2 storage reads/writes - usually i do this), otherwise, i prefer with empty lines.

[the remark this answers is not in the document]
enthusiastmartin: I agree that the implementation might look bit too long but the "meat of it" like you named it is just one or two lines - the rest is precheck and state updates. And it is the way it must be done, bit unfortunate that there are many things to check and update.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1337 @@
...
+			*current_value = current_value
+				.checked_add(&T::PositionInstanceId::one())
+				.ok_or(ArithmeticError::Overflow)?;
+
+			Ok(next_position_id)
+		})
+	}
+
+	/// Update Hub asset side of HDX subpool annd add given amount to hub_asset_reserve
```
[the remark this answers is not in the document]
enthusiastmartin: yes, kind of agree. but sometimes i prefer to add a comment.

pallets/omnipool/src/benchmarks.rs
```diff
@@ -0,0 +1,276 @@
+// This file is part of HydraDX-node
```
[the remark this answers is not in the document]
enthusiastmartin: It should be always as part of pallet if possible.

pallets/omnipool/src/lib.rs
```diff
@@ -0,0 +1,1509 @@
...
+
+	/// Retrieve state of asset from the pool and its pool balance
+	fn load_asset_state(asset_id: T::AssetId) -> Result<AssetReserveState<Balance>, DispatchError> {
+		let state = <Assets<T>>::get(asset_id).ok_or(Error::<T>::AssetNotFound)?;
+		let reserve = T::Currency::free_balance(asset_id, &Self::protocol_account());
+		Ok((state, reserve).into())
+	}
+
+	/// Set new state of asset.
+	/// This converts the new state into correct state type ) by removing the reserve)
```
apopiak, line 1247:
```suggestion
	/// This converts the new state into correct state type (by removing the reserve)
```

Issue #389, opened 2022-03-29
Fixed block reward for collators


lumir-mrkva:
deposit configurable amount (constant in runtime) on finalize to the block producer (collator)

maybe the collator selection pallet pot can be used for this, but just minting to the collator directly could be more straightforward

PR #399, opened 2022-04-12
feat: collator rewards pallet

martinfridrich:
Added collator-rewards pallet to runtime.

#389


Roznovjak (approved):

runtime/testing-hydradx/Cargo.toml
```diff
@@ -55,6 +55,7 @@ hydradx-traits = { git = "https://github.com/galacticcouncil/warehouse", rev = "
 pallet-relaychain-info = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false }
 pallet-transaction-multi-payment = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false}
 pallet-asset-registry = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false}
+pallet-collator-rewards = { git = "https://github.com/galacticcouncil/warehouse", rev = "08a9b3f8e7c690ce8282e0222d1e594988e58831", default-features = false}
```
apopiak, line 58: this could potentially pull in different substrate versions which we'd want to avoid, no?

enthusiastmartin: yes. we should try to have same revisions.

however, the approach we should stick with is that we use `branch = "main"` instead specific revision everywhere for warehouse pallets. 

the reason is that that we patch warehouse dependencies in top level Cargo file in both projects: HydraDX and Basilisk. 
This gives better control over dependencies imo and easier to upgrade to new version where you just change the top level cargo file.

lumir-mrkva (approved):

PR #413, opened 2022-06-05
feat: added multisig pallet

lumir-mrkva:


jak-pan:
LGTM if tested.

runtime/common/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "common-runtime"
-version = "100.0.3"
+version = "106.0.0"
```
dmoka, line 3: what is the reason for the 6 major version bumb? Keeping it in sync with hydradx-runtime and testing-hydradx-runtime?

dmoka (changes requested):
Left only one comment about the version change of common-runtime. If that is intentional, then I am fine the PR. ;)

I submitted my review as "Request changes" for safety reasons, because I am not sure that if I  submit with "Approve" then it will automatically merge it to master, or it is merged only after the PR owner resolves the new comments. What is our policy here? Please let me know, thank you!

runtime/common/Cargo.toml
enthusiastmartin: i believe it is just to match hydradx-runtime and testing-runtime versions. 

but it is necessary ? we dont do this in basilisk and common does not change everytime a runtime changes.

enthusiastmartin (commented):
looks ok to me.

if you want to have this benchmarked and not use default weights , you might want to add this to list of pallets to benchmark ( only in hydradx-runtime )

dmoka (approved):

PR #416, opened 2022-06-14
ci: version check fails if version was decreased

Roznovjak:


lumir-mrkva:
Task linked: [CU-2fkkub0 Version check fails if version was decreased](https://app.clickup.com/t/2fkkub0)

PR #418, opened 2022-06-21
feat!: Use new collator-rewards pallet with SessionManager

apopiak:
## Description
This PR uses the new collator rewards pallet implementation from https://github.com/galacticcouncil/warehouse/pull/57 to implement rewards in a different way:
`CollatorRewards` now provides a `SessionManager` implementation that wraps an inner `SessionManager` to
+ grab the collator set for each session
+ be notified of each session end and pay out the rewards to all non-excluded collators

without changing anything about the session keys.

## Related Issue

Addresses [#2hypfvk](https://app.clickup.com/t/2hypfvk)

## Motivation and Context

The original collator rewards implementation required us to change the session key types which would mean a runtime migration as well as changes for how to run the collators.

## How Has This Been Tested?

I ran a manual test with the testing runtime where I observed the events and token changes I would expect if the collators are paid. I needed to remove an account from the `ExcludedCollators` list to test it, though.
Do we want to not pay out collators when running the testing runtime?

## Checklist:

### DIFF
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx-runtime"
-version = "106.0.0"
+version = "107.0.0"
 authors = ["GalacticCouncil"]
 edition = "2021"
 license = "Apache 2.0"
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -51,7 +51,7 @@ hydradx-traits = { git = "https://github.com/galacticcouncil/warehouse", rev = "
 pallet-relaychain-info = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false }
 pallet-transaction-multi-payment = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false}
 pallet-asset-registry = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false}
-pallet-collator-rewards = { git = "https://github.com/galacticcouncil/warehouse", rev = "08a9b3f8e7c690ce8282e0222d1e594988e58831", default-features = false}
+pallet-collator-rewards = { git = "https://github.com/galacticcouncil/warehouse", branch = "apopiak/collator-rewards", default-features = false}
 
 # ORML dependencies
 orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "63b32194e7b9aff6a6350d2d4434525de4eec7c1", default-features = false }
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -381,6 +381,7 @@ impl pallet_session::Config for Runtime {
 	type ValidatorIdOf = pallet_collator_selection::IdentityCollator;
 	type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
 	type NextSessionRotation = pallet_session::PeriodicSessions<Period, Offset>;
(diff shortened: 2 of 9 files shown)


runtime/hydradx/src/lib.rs
```diff
@@ -382,7 +381,7 @@ impl pallet_session::Config for Runtime {
 	type ValidatorIdOf = pallet_collator_selection::IdentityCollator;
 	type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
 	type NextSessionRotation = pallet_session::PeriodicSessions<Period, Offset>;
-	type SessionManager = CollatorSelection;
+	type SessionManager = CollatorRewards;
```
apopiak, line 384: Implemented `SessionManager` on the pallet directly. Could move it to a dedicated type instead. (As this is the central piece of functionality of the pallet I don't think it's necessary, though.)

runtime/hydradx/Cargo.toml
```diff
@@ -51,7 +51,7 @@ hydradx-traits = { git = "https://github.com/galacticcouncil/warehouse", rev = "
 pallet-relaychain-info = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false }
 pallet-transaction-multi-payment = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false}
 pallet-asset-registry = { git = "https://github.com/galacticcouncil/warehouse", rev = "8b6bc9f287addba4ac1e4786118edc5154464d78", default-features = false}
-pallet-collator-rewards = { git = "https://github.com/galacticcouncil/warehouse", rev = "08a9b3f8e7c690ce8282e0222d1e594988e58831", default-features = false}
+pallet-collator-rewards = { git = "https://github.com/galacticcouncil/warehouse", branch = "apopiak/collator-rewards", default-features = false}
```
apopiak, line 54: This can lead to different warehouse pallets being included which then in turn could pull in different versions of substrate, no?

runtime/testing-hydradx/src/lib.rs
```diff
@@ -722,7 +722,7 @@ parameter_types! {
 		// Alice
 		hex!["d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"].into(),
 		// Bob
-		hex!["8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48"].into(),
+		// hex!["8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48"].into(),
```
apopiak, line 725: commented this out for the test to verify that Bob gets paid.
Why do we not pay testing collators?

runtime/hydradx/Cargo.toml
lumir-mrkva: not as long that branch stays on the same substrate, we can pin it on specific rev when the warehouse pr is merged

runtime/hydradx/src/lib.rs
lumir-mrkva: yeah, this is okay

runtime/testing-hydradx/src/lib.rs
lumir-mrkva: i guess this is just setup this way to test excluded collators set working correctly, we are not paying GC collators as well

apopiak:
Oh, this probably implies that we want to rerun session benchmarks to account for the different weight of `rotate_session`.

lumir-mrkva (approved):

lumir-mrkva:
i have tested this manually runtime upgrading on top of shell, it works like expected

runtime/testing-hydradx/src/lib.rs
apopiak: So remove this entry so we have one excluded collator (Alice) and one that gets rewards (Bob)?

lumir-mrkva: sure, that would be nice

PR #422, opened 2022-07-14
feat: replace bob with gc

green-jay:


lumir-mrkva (changes requested):
change vesting filter to use VestingPallet account when dispatched by root instead, remove GC account altogether as it is deprecated anyway

PR #423, opened 2022-07-18
feat: Change root origin to VestingPallet account

vgantchev:
This PR changes default root origin to Vesting pallet account.

There is an issue with the tests for `testing-hydradx` runtime, they are excluded by `cargo test`, any idea why?


runtime/testing-hydradx/src/mock.rs
```diff
@@ -0,0 +1,120 @@
+use frame_support::{parameter_types, traits::Everything, PalletId};
```
lumir-mrkva, line 1: not sure if its necessary to duplicate test code

runtime/hydradx/src/tests/vesting.rs
```diff
@@ -34,23 +34,18 @@ fn vested_transfer_from_vesting_account_should_work() {
...
 		let to: AccountId = BOB;
 
 		let vesting_schedule = schedule_object();
 
-		assert_ok!(Vesting::vested_transfer(
-			RawOrigin::Signed(from).into(),
-			to,
-			vesting_schedule
-		));
+		assert_ok!(Vesting::vested_transfer(RawOrigin::Root.into(), to, vesting_schedule));
```
lumir-mrkva, line 43: assert that balance is transfered as from the correct account (Vesting pallet)

lumir-mrkva (changes requested):

PR #430, opened 2022-08-02
fix: remove imbalance update from init_pool

### LINKED ISSUE #429: Bug: [Audit_RV] Incorrect imbalance update in pool initialization
The imbalance caused by the stable asset is added twice, while the imbalance caused by the native asset is ignored.

### Description

In initialize_pool (pallets/omnipool/src/lib.rs), recalculate_imbalance is called twice, and, each time after that, update_imbalance is being called. However, the second time this happens, the result from recalculate_imbalance seems to be mostly ignored (except for error checking), and update_imbalance is called with the same value as the first time.

According to the Discord discussion about this, currently the imbalance delta is 0 in both cases above, so this should not matter much.

Discord discussion: https://discord.com/channels/882700370307067966/1001114717827366913/1002461686034350130
Code link: https://github.com/galacticcouncil/HydraDX-node/blob/8de6222f967cc1bfe120207a775e8f0dc8eb92a5/pallets/omnipool/src/lib.rs#L441-L446

### Possible Fix

Either one of these should be fine

* Remove the imbalance code from initialize_pool (see the Discord discussion)
* Remove only the update_imbalance call and check that recalculate_imbalance returns 0
* Fix the second update_imbalance call.

### Context

Part of the RV audit of the omnipool.

enthusiastmartin:
Remove imbalance update from `init_pool`. Imbalance is always zero at the point of omnipool initialization. 

therefore, it is not needed to recalculate and update it. 

Fixed #429

### DIFF
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -424,27 +424,7 @@ pub mod pallet {
 				tradable: Tradability::default(),
 			};
 
-			// Imbalance update and total hub asset liquidity for stable asset first
-			// Note: cannot be merged with native, because the calculations depend on updated values
-			let delta_imbalance = Self::recalculate_imbalance(
-				&((&stable_asset_state, stable_asset_reserve).into()),
-				BalanceUpdate::Decrease(stable_asset_reserve),
-			)
-			.ok_or(ArithmeticError::Overflow)?;
-
-			// No imbalance yet, use default value
-			Self::update_imbalance(SimpleImbalance::default(), delta_imbalance)?;
-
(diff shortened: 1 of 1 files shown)


Roznovjak (approved):

lumir-mrkva:
benchmarks should be probably updated and version bumped

enthusiastmartin:
> benchmarks should be probably updated and version bumped

Omnipool was not yet benchmarked at all since it is not yet integrated in any runtime. 

Main omnipool PR is not yet merged , so I believe no need to increase version yet.

PR #443, opened 2022-09-01
fix: omnipool imbalance subtraction 

### LINKED ISSUE #432: [Audit_RV] Inconsistent sign for 0 in SimpleImbalance
The `SimpleBalance` implementation does not use a consistent sign for `0`. By default `0` is negative, and `add` also makes `0` negative, but `sub` makes it positive.

In the current implementation, `SimpleBalance` is used only with negative values, so `sub` should not produce a `0` value. Still, it may make sense to make the sign of `0` consistent, e.g. by making it always be `negative`.

https://github.com/galacticcouncil/HydraDX-node/blob/8de6222f967cc1bfe120207a775e8f0dc8eb92a5/pallets/omnipool/src/types.rs#L206

enthusiastmartin:
Fixed #432 

Sign should be consistent for 0 value. In case of sub - it could result in positive.

### DIFF
--- a/pallets/omnipool/src/tests/types.rs
+++ b/pallets/omnipool/src/tests/types.rs
@@ -130,7 +130,7 @@ fn simple_imbalance_subtraction_works() {
 		} - 300,
 		Some(SimpleImbalance {
 			value: 0,
-			negative: false
+			negative: true,
 		})
 	);
 	assert_eq!(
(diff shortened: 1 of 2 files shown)


Roznovjak (approved):
I would rename `sign` variable to `negative` and replace `self.negative` with bool values to make it a little bit more readable. But it's just a suggestion.

PR #444, opened 2022-09-01
feat: omnipool - ensure imbalance is negative

### LINKED ISSUE #434: [Audit_RV] Assumption not checked: LRNA imbalance is negative.
`sell` passes the absolute value of the hub asset imbalance to `calculate_sell_state_changes`, ignoring the sign: https://github.com/galacticcouncil/HydraDX-node/blob/8de6222f967cc1bfe120207a775e8f0dc8eb92a5/pallets/omnipool/src/lib.rs#L957 .

This should be safe, since the LRNA imbalance is always negative for now. However, in order to make the code more resilient to future changes, it may be worth adding a check.

enthusiastmartin:
ensure that imbalance after update is always negative. 

Fixes #434

### DIFF
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -329,6 +329,8 @@ pub mod pallet {
 		InsufficientTradingAmount,
 		/// Sell or buy with same asset ids is not allowed.
 		SameAssetTradeNotAllowed,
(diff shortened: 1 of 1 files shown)


Roznovjak (approved):
Looks good, but I would rename the method to something like `update_hub_imbalance`

PR #445, opened 2022-09-02
fix: correctly update hdx subpool hub reserve

### LINKED ISSUE #438: Bug: Incorrect update of HDX asset state when HDX token are involved in asset trading
When HDX token is the asset being traded in or traded out, its states are incorrectly updated.

### Description
According to the implementation, the function `update_hdx_subpool_hub_asset` is called before `set_asset_state()`. Thus, any changes resulted in the former is overwritten by the later when HDX is the asset traded in or the asset traded out. 
https://github.com/galacticcouncil/HydraDX-node/blob/20c8ac6079b8b3d6f16a9f15e9e43353ea6b9abc/pallets/omnipool/src/lib.rs#L983
https://github.com/galacticcouncil/HydraDX-node/blob/20c8ac6079b8b3d6f16a9f15e9e43353ea6b9abc/pallets/omnipool/src/lib.rs#L1106

In the case when `hub_asset_amount` is greater than `0`, a.k.a., `protocol fee` is greater than the `imbalance`, such imlementation could cause the `hub_reserve` for HDX is less than it should be. This indicates that in such a case, the price stored for HDX is less than its actual price which could lead to loss of Omnipool.

### Expected Behavior
Adding the `hub_asset_amount` to the new states of HDX.

### Actual Behavior
Asset state over written resulting less `hub_reserve` than it should be, when HDX is one of the assets being traded and when protocol fee is greater than the imbalance. 

### Possible Fix
There are three options: 

1. Execute function `update_hdx_subpool_hub_asset` after `set_asset_state()`
2. Merge the changes to the `new_asset_in_state` or `new_asset_out_state` if HDX is one of them. 
3. Merge the changes when computing state changes for traded assets:
https://github.com/galacticcouncil/HydraDX-node/blob/20c8ac6079b8b3d6f16a9f15e9e43353ea6b9abc/pallets/omnipool/src/lib.rs#L931
https://github.com/galacticcouncil/HydraDX-node/blob/20c8ac6079b8b3d6f16a9f15e9e43353ea6b9abc/pallets/omnipool/src/lib.rs#L1052

### Context
Audit by Runtime Verification.

### LINKED ISSUE #448: [Auidt_RV]Bug: Too much hub asset burned when removing liquidity
Remove liquidity would burn some hub asset. This would result in a violation of the invariant that 
The total value of all assets (except the hub asset) should equal to the balance of the hub asset, a.k.a.,
$$MultiCurrency::Accounts(Omnipool)(HubAssetId) = \Sigma_{t \in Assets}{Assets(t).hub\\_reserve}$$.

### Description
  After executing remove liquidity call, we have the following updates
```rust
Accounts(Omnipool)(HubAssetId) -= *state_changes.asset.delta_hub_reserve + state_changes.lp_hub_amount`
```
While 
$$\Sigma_{t \in Assets}{Assets(t).hub\\_reserve}$$ -=  `*state_changes.asset.delta_hub_reserve`
resulting a violation of the above invariant.

### Context
Audit by Runtime Verification

enthusiastmartin:
Fixes #438 

Fixed #448 

The PR also fixes issue where too much hub asset is burned in remove_liquidity. It should burn only difference between delta_hub and amount which is transferred to LP.

Additional change ensures that a trade can only burn LRNA and returns error when it tries to mint ( to prevent potential bugs if such would occur to avoid erroneous minting).

### DIFF
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -990,6 +990,9 @@ pub mod pallet {
 				.ok_or(ArithmeticError::Overflow)?;
 
 			match delta_hub_asset {
+				BalanceUpdate::Increase(val) if val == Balance::zero() => {
+					// nothing to do if zero.
+				}
 				BalanceUpdate::Increase(_) => {
 					// trade can only burn some.
 					return Err(Error::<T>::HubAssetUpdateError.into());
@@ -1126,6 +1129,9 @@ pub mod pallet {
 				.ok_or(ArithmeticError::Overflow)?;
 
 			match delta_hub_asset {
+				BalanceUpdate::Increase(val) if val == Balance::zero() => {
+					// nothing to do if zero.
+				}
 				BalanceUpdate::Increase(_) => {
 					// trade can only burn some.
 					return Err(Error::<T>::HubAssetUpdateError.into());
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -329,7 +329,7 @@ pub mod pallet {
 		InsufficientTradingAmount,
 		/// Sell or buy with same asset ids is not allowed.
 		SameAssetTradeNotAllowed,
-		/// LRNA update after trade results in non-zero value.
+		/// LRNA update after trade results in positive value.
 		HubAssetUpdateError,
 	}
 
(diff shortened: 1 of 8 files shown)


yanliu18 (commented):
Would also have to fix this [issue](https://github.com/galacticcouncil/HydraDX-math/issues/52) to make sure Line 993 works as intended.

enthusiastmartin:
> Would also have to fix this [issue](https://github.com/galacticcouncil/HydraDX-math/issues/52) to make sure Line 993 works as intended.

yes. it is already prepared in #444 and will be merged prior to this.

PR #459, opened 2022-09-08
fix: omnipool - ensure non-zero values

### LINKED ISSUE #452: [Audit_RV]Bug: Some condition checking in the specification is not implemented.
The specification requires some condition checking such as the shares to be withdrawn and amount to be swapped greater than 0, it is not checked in corresponding operations.

### Description
- [ ] Specification requires only registered asset can be added to the pool, however in operation `initialize_pool()`, it is not checked if `StableCoinAssetId` exists in AssetRegistry.
https://github.com/galacticcouncil/HydraDX-node/blob/16cdbcf25ef2f1dac5c3569d8cd44cc028ec4076/pallets/omnipool/src/lib.rs#L364
Note that, `HDXAssetId`  is registered at genesis. Furthermore, since `Omnipool::Assets` and `AssetRegistry` are managed separately, they do not imply each other.

- [ ] The amount to be added should be greater than 0 as specification document requires.
https://github.com/galacticcouncil/HydraDX-node/blob/16cdbcf25ef2f1dac5c3569d8cd44cc028ec4076/pallets/omnipool/src/lib.rs#L601
The condition checked above does not guarantee `amount >0`

- [ ] The amount of shares to be removed should be greater than 0.
https://github.com/galacticcouncil/HydraDX-node/blob/16cdbcf25ef2f1dac5c3569d8cd44cc028ec4076/pallets/omnipool/src/lib.rs#L734

enthusiastmartin:
Fixed #452

### DIFF
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -284,7 +284,7 @@ pub mod pallet {
 			recipient: T::AccountId,
 		},
 
-		/// Aseet's weight cap has been updated.
+		/// Asset's weight cap has been updated.
 		AssetWeightCapUpdated { asset_id: T::AssetId, cap: Permill },
 	}
 
@@ -331,6 +331,8 @@ pub mod pallet {
 		SameAssetTradeNotAllowed,
 		/// Imbalance results in positive value.
 		PositiveImbalance,
+		/// Amount of shares provided cannot be 0.
+		InvalidSharesAmount,
 	}
 
 	#[pallet::call]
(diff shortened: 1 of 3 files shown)


pallets/omnipool/src/lib.rs
```diff
@@ -724,6 +726,8 @@ pub mod pallet {
 			//
 			let who = ensure_signed(origin)?;
 
+			ensure!(amount > Balance::zero(), Error::<T>::InvalidSharesAmount);
```
Roznovjak, line 729: or `!amount.is_zero()`. It's up to you. Whatever you prefer more.

pallets/omnipool/src/lib.rs
```diff
@@ -1227,7 +1231,21 @@ pub mod pallet {
 	}
 
 	#[pallet::hooks]
-	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn integrity_test() {
+			if T::MinimumPoolLiquidity::get() == Balance::zero() {
```
Roznovjak, line 1236: We can replace `if`s with `assert_ne`s.

Roznovjak (approved):

PR #465, opened 2022-10-21
feat: enable HDX transfers

vgantchev:
This PR removes the call filters which were implemented to disable HDX transfers.

As this change only concerns HydraDX runtime, testing runtime version has not been bumped (which will result in version disparity - not sure if this is an issue).

Edit: Version check made me bump also testing runtime


lumir-mrkva (changes requested):
we may want to deploy another upgrade before transfers will be enabled, i would hold onto this until we know 100% we can immediately deploy this

lumir-mrkva (approved):

PR #466, opened 2022-10-23
chore: omnipool runtime integration

enthusiastmartin:
This PR adds following pallet to hydradx runtime:

- pallet-uniques
- pallet-nft
- pallet-omnipool


lumir-mrkva (changes requested):
pls add Uniques and NFT to call filter, so these pallets cannot be used by users directly

enthusiastmartin:
> pls add Uniques and NFT to call filter, so these pallets cannot be used by users directly

It is already in the filter.

PR #472, opened 2022-11-10
fix: omnipool asset trade state - derive typeinfo

enthusiastmartin:
Derive TypeInfo instead of implementing it manually as that did not work correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3298,7 +3298,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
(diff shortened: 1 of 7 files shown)


lumir-mrkva (approved):
@enthusiastmartin do you intent to merge this or replace with other solution?

enthusiastmartin:
> @enthusiastmartin do you intent to merge this or replace with other solution?

merge for now. so it would be possible to get the state. if another solution found, we can improve. 

merging.

PR #474, opened 2022-11-16
Fix hydradx testnet build

fakirAyoub:


green-jay:
was pointed out earlier by @lumir-mrkva that chain spec should be loaded from json file so it is not generated every time you build the node. on the other hand you lose visibility so not sure whether it is really worth it

node/src/testing_chain_spec.rs
```diff
@@ -180,10 +180,6 @@ pub fn local_parachain_config() -> Result<ChainSpec, String> {
 }
 
 pub fn devnet_parachain_config() -> Result<ChainSpec, String> {
-	ChainSpec::from_json_bytes(&include_bytes!("../res/devnet.json")[..])
```
lumir-mrkva, line 183: delete `devnet.json` from repo as well if it is not used anymore

lumir-mrkva (changes requested):

lumir-mrkva:
> was pointed out earlier by @lumir-mrkva that chain spec should be loaded from json file so it is not generated every time you build the node. on the other hand you lose readability so not sure whether it is really worth it

for devnet it is better to generate fresh chainspec all the time

enthusiastmartin:
> > was pointed out earlier by @lumir-mrkva that chain spec should be loaded from json file so it is not generated every time you build the node. on the other hand you lose readability so not sure whether it is really worth it
> 
> for devnet it is better to generate fresh chainspec all the time as you want to change initial state in chainspec and restart the network. for networks that cannot be restarted opposite applies

removed

lumir-mrkva (approved):

PR #477, opened 2022-11-21
feat: add uniques migration

Roznovjak:
I needed to add `dev` identifier for spec file, otherwise `try-runtime` tool was complaining that it can't find it.


runtime/hydradx/src/migrations.rs
```diff
@@ -0,0 +1,25 @@
...
+		frame_support::log::info!("Migrate Uniques Pallet start");
+		weight = weight.saturating_add(pallet_uniques::migration::migrate_to_v1::<Runtime, _, Uniques>());
+		frame_support::log::info!("Migrate Uniques Pallet end");
+
+		weight
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade() -> Result<(), &'static str> {
+		Ok(())
```
enthusiastmartin, line 23: Isnt post upgrade to verify that upgrade does it what is supposed to ?  Can we add a check for uniques ?

Roznovjak: Yes, but unfortunately it's not implemented for `uniques`, `pre_upgrade` and `post_upgrade` implementations are missing.

lumir-mrkva (approved):

PR #483, opened 2022-12-01
feat: add trade volume limit per block

dmoka:
## Checklist:


pallets/circuit-breaker/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "pallet-circuit-breaker"
+version = "1.0.0"
+description = "Circuit breakers for AMMs"
+authors = ["GalacticCouncil <hydradx@galacticcouncil.io>"]
+edition = "2021"
+license = "Apache-2.0"
+homepage = "https://substrate.dev"
```
enthusiastmartin, line 8: nope. i dont think this is the hompage.

pallets/circuit-breaker/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+repository = "https://github.com/paritytech/substrate/"
```
enthusiastmartin, line 9: Not this repo.

pallets/circuit-breaker/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+
+[package.metadata.docs.rs]
+targets = ['x86_64-unknown-linux-gnu']
+
+[dependencies]
+serde = { features = ["derive"], optional = true, version = "1.0.136" }
+codec = { default-features = false, features = ["derive"], package = "parity-scale-codec", version = "3.1.5" }
+scale-info = { version = "2.1.1", default-features = false, features = ["derive"] }
+derive_more = "0.99.17"
```
enthusiastmartin, line 18: what is derive more used for ?

pallets/circuit-breaker/README.md
```diff
@@ -0,0 +1,6 @@
+### Circuit Breaker pallet
+
+This pallet is used to keep track of the initial liquidity of assets (of an AMM), 
+then to validate the next liquidity volume per block per asset,
+
+Before using validation, the initial liquidity for an asset must be registered with the pallet, otherwise it results in error.
```
enthusiastmartin, line 6: hm.this readme does not make much sense to me.

after reading, i could not tell what it does. I'd prefer to rewrite it.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub use primitives::Balance;
```
enthusiastmartin, line 20: i dont get this.  

pallet's config has generic Balance but still there is this concrete type. 

either this or that. 

it does not seem to be used.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+			<AllowedLiquidityRangePerAsset<T>>::insert(asset_id, (min_limit, max_limit));
+		}
+		Ok(())
+	}
+
+	fn test_liquidity_limits(asset_id: T::AssetId, updated_liquidity: T::Balance) -> DispatchResult {
+		let (min_limit, max_limit) = Pallet::<T>::allowed_liqudity_range_per_asset(asset_id)
+			.ok_or(Error::<T>::LiquidityLimitNotStoredForAsset)?;
+
+		//TODO: tell don't ask, add this in some LimitRange object or so
```
enthusiastmartin, line 163: TODO

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+				.ok_or(ArithmeticError::Overflow)?;
+			let max_limit = initial_liquidity
+				.checked_add(&liquidity_diff)
+				.ok_or(ArithmeticError::Overflow)?;
+			<AllowedLiquidityRangePerAsset<T>>::insert(asset_id, (min_limit, max_limit));
+		}
+		Ok(())
+	}
+
+	fn test_liquidity_limits(asset_id: T::AssetId, updated_liquidity: T::Balance) -> DispatchResult {
```
enthusiastmartin, line 159: i would prefer something else than "test" . 

because i was looking at this like " why a test is here" first. 

eg.

```suggestion
	fn ensure_liquidity_limits(asset_id: T::AssetId, updated_liquidity: T::Balance) -> DispatchResult {
```

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	fn calculate_and_store_liquidity_limits(asset_id: T::AssetId, initial_liquidity: T::Balance) -> DispatchResult {
+		if !<AllowedLiquidityRangePerAsset<T>>::contains_key(asset_id) {
+			let liquidity_diff = Pallet::<T>::trade_volume_limit_per_asset(asset_id).mul_floor(initial_liquidity);
+			let min_limit = initial_liquidity
+				.checked_sub(&liquidity_diff)
+				.ok_or(ArithmeticError::Overflow)?;
```
enthusiastmartin, line 150:
```suggestion
				.ok_or(ArithmeticError::Underflow)?;
```

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+				.checked_add(&liquidity_diff)
+				.ok_or(ArithmeticError::Overflow)?;
+			<AllowedLiquidityRangePerAsset<T>>::insert(asset_id, (min_limit, max_limit));
+		}
+		Ok(())
+	}
+
+	fn test_liquidity_limits(asset_id: T::AssetId, updated_liquidity: T::Balance) -> DispatchResult {
+		let (min_limit, max_limit) = Pallet::<T>::allowed_liqudity_range_per_asset(asset_id)
+			.ok_or(Error::<T>::LiquidityLimitNotStoredForAsset)?;
```
enthusiastmartin, line 161: Sure that we want to error if this is not stored ?

pallets/circuit-breaker/src/lib.rs
Roznovjak: No. In this case it's `Overflow`. `Underflow` happens with floating point numbers when you try to store a value that is smaller than the smallest representable number.

pallets/circuit-breaker/src/lib.rs
Roznovjak: I would say yes. I think you want to always provide as much info as you can. You can always ignore it or transform it to `Option` or `bool`. But in our case we expect that all trades are within the trade volume limit. After all, it's a security feature, not some unreliable data provider.

pallets/circuit-breaker/README.md
dmoka: rewritten, please check again if it makes more sense.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+		/// Set trade volume limit for an asset.
+		///
+		/// Parameters:
+		/// - `asset_id`: The identifier of an asset
+		/// - `trade_volume_limit`: Trade volume limit represented as a percentage
+		///
+		/// Doesn't emit any event.
+		///
+		#[pallet::weight(<T as Config>::WeightInfo::set_trade_volume_limit())]
+		#[transactional]
```
apopiak, line 281: this should not be necessary anymore, right?

pallets/omnipool/src/tests/sell.rs
```diff
@@ -675,3 +676,209 @@ fn sell_lrna_should_fail_when_exceeds_max_out_ratio() {
...
+				Origin::signed(TRADER),
+				LRNA,
+				ACA,
+				sell_amount,
+				min_limit
+			),);
+		});
+}
+
+//TODO Dani: add integration tests for complex cases
```
dmoka, line 884: Can be removed, some integration tests have been already added, and possible one more will be about liquidity addition

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+			.ok_or(ArithmeticError::Overflow)?;
+		self.amount_out = self
+			.amount_out
+			.checked_add(&amount_out)
+			.ok_or(ArithmeticError::Overflow)?;
+		Ok(())
+	}
+
+	pub fn check_min_limit(&self) -> DispatchResult {
+		if self.amount_out > self.amount_in {
```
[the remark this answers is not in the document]
Roznovjak: This one (`check_min_limit`) checks the lower limit (if we removed more liquidity than we can). And it can only happen if `amount_out > amount_in`. We actually test that `abs(amount_in - amount_out) <= limit`. And by dividing it into two checks we can tell whether we reached the lower or upper limit. There is another method (`check_max_limit`) to check the upper limit. We always call both methods, but the logic is divided into two cases.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+	use frame_system::pallet_prelude::*;
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_finalize(_n: T::BlockNumber) {
+			let _ = <AllowedTradeVolumeLimitPerAsset<T>>::clear(u32::MAX, None);
+			let _ = <AllowedLiquidityAmountPerAsset<T>>::clear(u32::MAX, None);
+		}
+
+		fn integrity_test() {
```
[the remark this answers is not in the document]
Roznovjak: Could you be more specific what tests you would like to see there? The test is meant to be used for verifying that some required configuration conditions are met. I believe everything from the config is covered.

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
dmoka: got it, thanks for the explanation!

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
dmoka: I would want to see some tests showing that this integrity check fails when the configs are not configured properly.

Right now it passes as it is configured properly in the mock.

enthusiastmartin:
resolve conflicts

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+			+ CheckedAdd
+			+ CheckedSub
+			+ AtLeast32BitUnsigned;
+
+		/// Origin able to change the trade volume limit of an asset.
+		type TechnicalOrigin: EnsureOrigin<Self::Origin>;
+
+		/// The maximum percentage of a pool's liquidity that can be traded in a block.
+		/// Represented as a non-zero fraction (nominator, denominator) with the max value being 10_000.
+		type DefaultMaxNetTradeVolumeLimitPerBlock: Get<(u32, u32)>;
```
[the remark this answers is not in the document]
enthusiastmartin: why PerThing cannot be used ?

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+
+		/// Omnipool's hub asset id. The limits are not tracked for this asses.
+		type OmnipoolHubAsset: Get<Self::AssetId>;
+
+		/// Weight information for extrinsics in this pallet.
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::pallet]
+	#[pallet::without_storage_info]
```
enthusiastmartin, line 199: remove without_storage_info

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
enthusiastmartin: fair enough

pallets/circuit-breaker/src/lib.rs
Roznovjak: any idea how to bound the storage?

apopiak: it should be bounded automatically? no vecs or other dynamic structures

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+#[cfg(test)]
+mod tests;
+
+/// Max trade volume limit multiplier of liquidity that can be traded in a block
+pub const MAX_LIMIT_VALUE: u32 = 10_000;
+
+#[derive(Clone, Encode, Decode, RuntimeDebug, MaxEncodedLen, TypeInfo, Eq, PartialEq)]
+#[scale_info(skip_type_params(T))]
+pub struct TradeVolumeLimit<T: Config> {
+	pub amount_in: T::Balance,
```
mrq1911, line 42: is it necessary to track both amount? wouldn't be easier to have just one signed value?

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+	use codec::HasCompact;
+	use frame_support::pallet_prelude::*;
+	use frame_system::pallet_prelude::*;
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_finalize(_n: T::BlockNumber) {
```
mrq1911, line 133: i would consider to clear values on initialize instead, given that this happens before DCA and Scheduler

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+}
+
+impl<T: Config> OnLiquidityChangeHandler<T::AssetId, T::Balance> for Pallet<T> {
+	fn after_add_liquidity(
+		asset_id: T::AssetId,
+		initial_liquidity: T::Balance,
+		added_liquidity: T::Balance,
+	) -> DispatchResult {
+		Pallet::<T>::calculate_and_store_liquidity_limit(asset_id, initial_liquidity)?;
+		Pallet::<T>::ensure_and_update_liquidity_limit(asset_id, added_liquidity)?;
```
mrq1911, line 467: we will probably want to ignore limits when the liquidity is added by the protocol (root origin)

mrq1911 (changes requested):

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+
+			<AllowedLiquidityAmountPerAsset<T>>::insert(asset_id, allowed_liquidity_limit);
+		}
+
+		Ok(())
+	}
+
+	pub fn validate_limit(limit: (u32, u32)) -> DispatchResult {
+		ensure!(
+			limit.0 <= MAX_LIMIT_VALUE && limit.1 <= MAX_LIMIT_VALUE,
```
mrq1911, line 423: use destructuring for better readibility

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+		///
+		#[pallet::weight(<T as Config>::WeightInfo::set_trade_volume_limit())]
+		pub fn set_trade_volume_limit(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+			trade_volume_limit: (u32, u32),
+		) -> DispatchResult {
+			T::TechnicalOrigin::ensure_origin(origin)?;
+
+			ensure!(asset_id != T::OmnipoolHubAsset::get(), Error::<T>::NotAllowed);
```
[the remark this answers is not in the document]
Roznovjak: This check is for HUB asset. We don't track limits for it.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+	pub struct Pallet<T>(_);
+
+	/// Default maximum net trade volume limit per block
+	#[pallet::type_value]
+	pub fn DefaultTradeVolumeLimit<T: Config>() -> (u32, u32) {
+		T::DefaultMaxNetTradeVolumeLimitPerBlock::get()
+	}
+
+	#[pallet::storage]
+	/// Trade volume limits of assets that don't use the default value
```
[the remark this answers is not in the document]
Roznovjak: This is intentional. It allows us to set some global default value for all assets, and overwrite it with `set_trade_volume_limit`.

pallets/circuit-breaker/src/lib.rs
Roznovjak: Yes, it's possible to use i128, but trade volume limit will be limited to u64 (in both directions). And the conversion will not be infallible.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+		},
+	}
+
+	#[pallet::error]
+	#[cfg_attr(test, derive(PartialEq, Eq))]
+	pub enum Error<T> {
+		/// Invalid value for a limit. Limit must be non-zero.
+		InvalidLimitValue,
+		/// Allowed liquidity limit is not stored for asset
+		LiquidityLimitNotStoredForAsset,
```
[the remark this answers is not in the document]
Roznovjak: In the previous implementation, there were two handlers. The first one for registering assets, the second one for updating the values. So it was possible call the second handler before registering the assets. In the current implementation we use just one handler, so it's not a problem anymore. But I decided to keep this error because the getter method returns `Option`.

pallets/circuit-breaker/src/lib.rs
mrq1911: ok, not worth probably, i would rename fields to be not confused with swap function params

pallets/circuit-breaker/src/lib.rs
Roznovjak: yes, but we need to figure out how much weight to reserve for it in `on_initialize`.

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
mrq1911: so that mean you cannot trade hub asset now?

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
mrq1911: i am just referring to the inaccurate comment as the store getter will in fact return default values that you configured

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
Roznovjak: not sure about this. Limit is calculated from an asset reserve. We can use `0` or `1` as the default value, something what prevents users from doing operations on it.

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
Roznovjak: What I wanted to say by this comment is that we store there limits that are not default values. I will update it to be more accurate.

pallets/circuit-breaker/src/lib.rs
mrq1911: you can do it ;)

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
mrq1911: sry i confused this with a limit config (which has default value), still this error shouldn't happen because the pair should be always initialized already. therefore i would use something like genric `InconsistentState` for it rather than named pallet event

pallets/circuit-breaker/src/lib.rs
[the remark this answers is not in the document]
mrq1911: but you can sell lerna to the every pair, it would make sense to limit how much can you sell it in one block

mrq1911 (approved):

PR #490, opened 2022-12-21
fix: add currency balance check to signed extras

enthusiastmartin:


runtime/testing-hydradx/src/lib.rs
```diff
@@ -221,20 +222,6 @@ impl Contains<Call> for BaseFilter {
 			return false;
 		}
 
-		// filter transfers of LRNA to the omnipool account
-		if let Call::Tokens(orml_tokens::Call::transfer { dest, currency_id, .. })
```
jak-pan, line 225: Is this intentional as part of this PR?

runtime/testing-hydradx/src/lib.rs
```diff
@@ -221,20 +222,6 @@ impl Contains<Call> for BaseFilter {
 			return false;
 		}
 
-		// filter transfers of LRNA to the omnipool account
```
lumir-mrkva, line 224: why did you remove this? some local changes slipped in?

runtime/testing-hydradx/src/lib.rs
```diff
@@ -696,6 +683,19 @@ impl pallet_claims::Config for Runtime {
 
 impl pallet_genesis_history::Config for Runtime {}
 
+pub struct NoSpotPriceProvider;
```
lumir-mrkva, line 686: looks like you reverted some changes as well

lumir-mrkva (changes requested):

PR #494, opened 2023-01-03
test: add integration tests for call filter

Roznovjak:
Fixes https://github.com/galacticcouncil/warehouse/issues/91


martinfridrich:
This can be done only for Basilisk. Hydra uses `uniques` instead of `pallet-nft`.

martinfridrich (approved):

PR #496, opened 2023-01-09
feat: make omnipool's tvl cap stored value

enthusiastmartin:
This PR removes TVL cap constant from pallet's config and introduces TVL cap as stored value.

This value can be changed using set_tvl_cap extrinsic.

### DIFF
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -297,6 +297,9 @@ pub mod pallet {
 
 		/// Asset's weight cap has been updated.
 		AssetWeightCapUpdated { asset_id: T::AssetId, cap: Permill },
+
+		/// TVL cap has been updated.
+		TVLCapUpdated { cap: Balance },
 	}
 
 	#[pallet::error]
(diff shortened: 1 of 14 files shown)


pallets/omnipool/src/lib.rs
```diff
@@ -223,6 +219,10 @@ pub mod pallet {
 	/// Position ids sequencer
 	pub(super) type NextPositionId<T: Config> = StorageValue<_, T::PositionItemId, ValueQuery>;
 
+	#[pallet::storage]
+	/// TVL cap
+	pub(super) type TvlCap<T: Config> = StorageValue<_, Balance, ValueQuery>;
```
apopiak, line 224: I would either make this an option query indicating that it can be unset (and define the default at call time) or set a reasonable default value.

pallets/omnipool/src/lib.rs
```diff
@@ -1312,6 +1315,21 @@ pub mod pallet {
...
+		///
+		/// Parameters:
+		/// - `cap`: new tvl cap
+		///
+		/// Emits `TVLCapUpdated` event when successful.
+		///
+		#[pallet::weight(<T as Config>::WeightInfo::set_asset_weight_cap())]
+		#[transactional]
+		pub fn set_tvl_cap(origin: OriginFor<T>, cap: Balance) -> DispatchResult {
+			T::AddTokenOrigin::ensure_origin(origin)?;
```
mrq1911, line 1328: rename this origin, add new one or at least document this behaviour

pallets/omnipool/src/lib.rs
mrq1911: 0 is a reasonable default value as it prevents anyone to add liquidity

enthusiastmartin: I was thinking to make it OptionQuery. 

and since we dont want to have default value as constant somewhere. this would introduce new error too ( in case of no tvl cap set ). 

I am not sure yet about this one. 

so far, i am for default value to be 0. no harm, since you can't add liquidity at all. and it does not have impact on other functionality.

mrq1911: MAX would be also acceptable, but less safe in my opinion

pallets/omnipool/src/lib.rs
enthusiastmartin: renamed to AuthorityOrigin.

mrq1911 (approved):

PR #500, opened 2023-01-16
feat: transfer dust to Treasury

Roznovjak:


integration-tests/src/dust.rs
```diff
@@ -0,0 +1,124 @@
...
+
+	Hydra::execute_with(|| {
+		let hdx_ed = <hydradx_runtime::Runtime as pallet_balances::Config>::ExistentialDeposit::get();
+
+		assert!(!hdx_ed.is_zero());
+
+		let transfer_amount = hdx_ed.checked_sub(1).unwrap();
+
+		// set Treasury balance to ED so it's not dusted
+		assert_ok!(Balances::set_balance(
```
enthusiastmartin, line 21: dont set ED for treasury account.

the treasury account cannot be dusted. You can assert that as well, that it is possible to have < ED in the treasury account and it wont get dusted.

Roznovjak: Well, no. The treasury account is dusted if the account doesn't exist, and the total issuance is decreased by the amount.

Roznovjak: The balance pallet (and the treasury pallet) are trying to `deposit_creating` to the treasury account (which doesn't exist yet) with the amount that is below ED, it results in no-op, and the amount id `Drop`ped.

enthusiastmartin: i see , it is strange. it feels like it should be allowed.

Roznovjak: Yes. The problem is that there is not a whitelist. Just a handler. So they are trying to solve dusting by sending dust somewhere.

enthusiastmartin (approved):

mrq1911 (approved):

enthusiastmartin:
bump version and merge

PR #501, opened 2023-01-19
test: extend omnipool integration test - add ETH and BTC

enthusiastmartin:
This PR adds ETH and BTC to omnipool integration tests. it mainly helps to verify correct prices and amounts in omnipool state.


vgantchev (approved):
LGTM

PR #506, opened 2023-01-25
feat: add omnipool pallet interface

enthusiastmartin:
This PRs adds some interface that can be used by other pallets that build on top of -  subpools or liquidity mining.

### DIFF
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -100,7 +100,7 @@ pub use pallet::*;
 pub use weights::WeightInfo;
 
 /// NFT class id type of provided nft implementation
-type NFTCollectionIdOf<T> =
+pub type NFTCollectionIdOf<T> =
 	<<T as Config>::NFTHandler as Inspect<<T as frame_system::Config>::AccountId>>::CollectionId;
 
 #[frame_support::pallet]
--- a/pallets/omnipool/src/lib.rs
+++ b/pallets/omnipool/src/lib.rs
@@ -78,7 +78,7 @@ use sp_std::ops::{Add, Sub};
 use sp_std::prelude::*;
 
 use frame_support::traits::tokens::nonfungibles::{Create, Inspect, Mutate};
-use hydra_dx_math::omnipool::types::{BalanceUpdate, I129};
+use hydra_dx_math::omnipool::types::{AssetStateChange, BalanceUpdate, HubTradeStateChange, TradeStateChange, I129};
 use hydradx_traits::Registry;
 use orml_traits::MultiCurrency;
 use scale_info::TypeInfo;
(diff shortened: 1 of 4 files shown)


pallets/omnipool/src/lib.rs
```diff
@@ -1659,4 +1673,111 @@ impl<T: Config> Pallet<T> {
...
+		let state = Self::load_asset_state(asset_id)?;
+		let updated_state = state.delta_update(&delta).ok_or(ArithmeticError::Overflow)?;
+		Self::set_asset_state(asset_id, updated_state);
+
+		Ok(())
+	}
+
+	/// Load position and check its owner
+	/// Returns Forbidden if not position owner
+	pub fn load_position(
```
Roznovjak, line 1763: I would rewrite it to use optional `maybe_check_owner` parameter. But I also don't have any problem to leave it as it is.

pallets/omnipool/src/lib.rs
```diff
@@ -1659,4 +1673,111 @@ impl<T: Config> Pallet<T> {
...
+
+		Self::update_asset_state(asset_in, trade.asset_in)?;
+		Self::update_asset_state(asset_out, trade.asset_out)?;
+
+		Self::update_hdx_subpool_hub_asset(trade.hdx_hub_amount)?;
+
+		Ok(())
+	}
+
+	/// Updates states of anasset given calculated trade result where HUb asset was traded.
```
Roznovjak, line 1741:
```suggestion
	/// Updates states of an asset given calculated trade result where hub asset was traded.
```

pallets/omnipool/src/lib.rs
```diff
@@ -1659,4 +1673,111 @@ impl<T: Config> Pallet<T> {
 
 		Err(Error::<T>::NotAllowed.into())
 	}
+
+	/// Remove asset from list of Omnipool assets.
+	pub fn remove_asset(asset_id: T::AssetId) -> DispatchResult {
```
Roznovjak, line 1678: We can emit new `TokenRemoved` event from this one. Other methods can emit events as well.

pallets/omnipool/src/lib.rs
```diff
@@ -1659,4 +1673,111 @@ impl<T: Config> Pallet<T> {
...
+	}
+
+	/// Insert or update position with given position data.
+	pub fn set_position(position_id: T::PositionItemId, position: &Position<Balance, T::AssetId>) -> DispatchResult {
+		<Positions<T>>::insert(position_id, position);
+		Ok(())
+	}
+
+	/// Add new asset to list of Omnipool assets.
+	pub fn add_asset(asset_id: T::AssetId, state: AssetState<Balance>) -> DispatchResult {
```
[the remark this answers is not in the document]
enthusiastmartin: this is ok.

pallets/omnipool/src/lib.rs
enthusiastmartin: can be refactor later when it would be need it, i dont see the need atm

pallets/omnipool/src/lib.rs
enthusiastmartin: this could be confusing, because current use case does not really remove the asset, it is just moved to subpool. 

subpool implemenation actually emit an event "AssetMigrated"

Roznovjak: It's a public interface. Add there a comment that these methods modify the storage and don't emit any event, so any consumer is aware of it.

martinfridrich (approved):
LGTM

Roznovjak (commented):
IMHO the minor version should be increased

Roznovjak (approved):

PR #507, opened 2023-01-25
chore: update to latest warehouse

enthusiastmartin:
This PR updates the version of warehouse pallets to the latest.

The most significant change is asset registry which now supports registering assets with given ID.

### DIFF
--- a/integration-tests/src/polkadot_test_net.rs
+++ b/integration-tests/src/polkadot_test_net.rs
@@ -179,6 +179,11 @@ pub fn hydra_ext() -> sp_io::TestExternalities {
 			(b"DAI".to_vec(), 1_000u128),
 			(b"USDC".to_vec(), 1_000u128),
 		],
+		asset_ids: vec![
+			(b"LRNA".to_vec(), 1_000u128, LRNA),
+			(b"DAI".to_vec(), 1_000u128, DAI),
+			(b"USDC".to_vec(), 1_000u128, DOT),
+		],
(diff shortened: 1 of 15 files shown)


integration-tests/src/polkadot_test_net.rs
```diff
@@ -179,6 +179,11 @@ pub fn hydra_ext() -> sp_io::TestExternalities {
 			(b"DAI".to_vec(), 1_000u128),
 			(b"USDC".to_vec(), 1_000u128),
 		],
+		asset_ids: vec![
+			(b"LRNA".to_vec(), 1_000u128, LRNA),
+			(b"DAI".to_vec(), 1_000u128, DAI),
+			(b"USDC".to_vec(), 1_000u128, DOT),
+		],
```
apopiak, line 186: This will add the same assets twice which is probably not what you want.
Maybe I should have set up the genesis config differently :thinking: (For one asset you basically want to add it to *either* `asset_names` *or* `asset_ids`.)

apopiak (approved):
LGTM if you decide on using either `asset_names` or `asset_ids` for the assets in the integration test. (Probably `asset_ids` as LRNA should be asset 1)

integration-tests/src/polkadot_test_net.rs
enthusiastmartin: i was wondering about that too. I was thinking just to have one and you have to provide asset id.

PR #509, opened 2023-01-26
test: add test to ensure omnipool's imbalance is reduced to 0 correctly

enthusiastmartin:
Adding a test to ensure that after certain amount of trades, imbalance is correctly reduced to 0 and subsequent operations work correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6149,7 +6149,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-omnipool"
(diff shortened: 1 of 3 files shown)


Roznovjak (approved):
LGTM. But I would add a counter to the loop and break it with some error if the number of iterations reaches some limit. As a prevention to avoid endless running of the test if we change something in the future. Afaik tests don't have timeouts.

PR #512, opened 2023-01-30
feat!: Integrate EMA oracle

apopiak:
## Checklist:


apopiak:
Note: We only want TKNA - LRNA and TKNB - LRNA, not TKNA - TKNB to keep number of oracles down

PR #515, opened 2023-02-08
feat: OTC desk

vgantchev:
Related PR:
https://github.com/galacticcouncil/warehouse/pull/165

This PR introduces a pallet-otc which enables over-the-counter trading.


integration-tests/src/otc.rs
```diff
@@ -0,0 +1,141 @@
...
+			HDX,
+			20 * UNITS,
+			100 * UNITS,
+			false,
+		));
+
+		// Assert
+		let order = hydradx_runtime::OTC::orders(0);
+		assert!(order.is_some());
+		assert_eq!(
```
dmoka, line 45: small thing but for this I like using methods like:

assert_alice_native_named_reserve_balance

Or making it more generic with the name as param:

assert_native_named_reserve_balance

So the the domain becomes more explicit.

integration-tests/src/otc.rs
```diff
@@ -0,0 +1,141 @@
...
+			true,
+		));
+
+		// Act
+		assert_ok!(hydradx_runtime::OTC::fill_order(
+			hydradx_runtime::Origin::signed(BOB.into()),
+			0,
+		));
+
+		//Assert
```
dmoka, line 103:
```suggestion
		// Assert
```

dmoka (approved):

mrq1911 (approved):

PR #531, opened 2023-03-07
chore: reorder scheduler pallet in construct_runtime

martinfridrich:
## Description
This PR changes order of the pallets in `construct_runtime` specifically `Scheduler` pallet which must be after `ParachainSystem` so we can get relaychain block number is the scheduler and democracy. 
This is necessary for liquidity mining to calculate `period`.

This PR is also adding `static_assertion` to make sure native existential deposit is less than liquidity mining `pot` default balance. `Pot`'s default balance is used to prevent dusting.

### DIFF
--- a/primitives/src/constants.rs
+++ b/primitives/src/constants.rs
@@ -28,6 +28,12 @@ pub mod currency {
 	pub fn deposit(items: u32, bytes: u32) -> Balance {
 		items as Balance * 2 * DOLLARS + (bytes as Balance) * 30 * MILLICENTS
 	}
+
+	// Value on the right side of this condition represents balance deposited to liquidity mining's pot
+	// account in native currency to prevent dusting of the pot. Pot exists for every instance of warehouse
+	// liq. mining and this value has to be deposited to all pot's instances.
+	// WARN: More tokens must be sent to pots when this value is changed.
+	static_assertions::const_assert!(NATIVE_EXISTENTIAL_DEPOSIT < 100 * UNITS);
 }
 
 pub mod time {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3387,7 +3387,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "131.0.0"
+version = "132.0.0"
 dependencies = [
  "common-runtime",
  "cumulus-pallet-aura-ext",
@@ -8178,13 +8178,14 @@ dependencies = [
 
 [[package]]
 name = "primitives"
-version = "5.0.6"
+version = "5.7.0"
 dependencies = [
  "frame-support",
  "hex-literal",
  "parity-scale-codec",
  "scale-info",
  "serde",
+ "static_assertions",
 ]
 
 [[package]]
@@ -11504,7 +11505,7 @@ dependencies = [
 
 [[package]]
 name = "testing-hydradx-runtime"
-version = "131.0.0"
+version = "132.0.0"
 dependencies = [
  "common-runtime",
  "cumulus-pallet-aura-ext",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1027,7 +1027,7 @@ dependencies = [
 
 [[package]]
 name = "common-runtime"
-version = "100.3.0"
+version = "100.4.0"
 dependencies = [
  "cumulus-pallet-xcmp-queue",
  "frame-support",
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "primitives"
-version = "5.0.6"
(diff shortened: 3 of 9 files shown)


primitives/src/constants.rs
```diff
@@ -28,6 +28,12 @@ pub mod currency {
 	pub fn deposit(items: u32, bytes: u32) -> Balance {
 		items as Balance * 2 * DOLLARS + (bytes as Balance) * 30 * MILLICENTS
 	}
+
+	// Value on the right side of this condition represents balance deposited to liquidity mining's pot
+	// account in native currency to prevent dusting of the pot. Pot exists for every instance of warehouse
+	// liq. mining and this value has to be deposited to all pot's instances.
+	// WARN: More tokens must be sent to pots when this value is changed.
+	static_assertions::const_assert!(NATIVE_EXISTENTIAL_DEPOSIT < 100 * UNITS);
```
Roznovjak, line 36: I would write an integration test for this

Roznovjak (commented):
bump runtime versions

primitives/src/constants.rs
martinfridrich: Why, what would be benefits of integration test? 
This only needs to be checked when you are changing existential deposit constant and this way it will fail immediately, without `cargo test`

Roznovjak: 1. I wouldn't expect a test in `constants.rs`
2. in an integration test, you can directly compare the existential deposit to the balance of the pot account. The existential deposit doesn't have to be less than `100*UNITS`. The balance of the pot account needs to be greater than the ED.
3. now when I'm thinking about it, it actually doesn't test anything, we are just comparing two values. It doesn't test that the amount is actually deposited to the pot account. So this test is probably not necessary.
4. I don't see a reason why compilation should fail because of this. And if it fails, we can easily fix it by increasing the number on the right side. But it actually doesn't fix anything.

Roznovjak: Let's ask @enthusiastmartin what is his opinion on this.

martinfridrich: > 1. I wouldn't expect a test in `constants.rs`
> 2. in an integration test, you can directly compare the existential deposit to the balance of the pot account. The existential deposit doesn't have to be less than `100*UNITS`. The balance of the pot account needs to be greater than the ED.
> 3. now when I'm thinking about it, it actually doesn't test anything, we are just comparing two values. It doesn't test that the amount is actually deposited to the pot account. So this test is probably not necessary.
> 4. I don't see a reason why compilation should fail because of this. And if it fails, we can easily fix it by increasing the number on the right side. But it actually doesn't fix anything.

1.- it's not "test", it's just configuration "assert".

2.- you can't directly compare it. You would have to still do transfer of 100 UNITS to `pot` account first. "default" balance(100 HDX) in the `pot` is done on chain by doing transfer to the pot from treasury when we first time create LM. ED have to be less than `100*units`, you can't use on chain balance in the `pot` because it may contain unpaid rewards e.g.: pot balance: 100(default) + 10(unpaid) and you will set ED to 101 - pot will be dusted when 10(unpaid) rewards will be paid.  

3. - yes, it doesn't test if 100 was deposited to the `pot`. We can't test it without on chain data and this test wouldn't be useful because what I wrote in 2-nd.
This test basically saying: if you set ED > 100 * UNIT, you have to change `100 * UNIT` and transfer additional amount to `pot`.

4. - compilation should fail to notify you that you have to do additional on chain action.  You always have to make sure ED < "default pot balance" and if you want to set higher ED than current limit you need to add tokens to pot.

Roznovjak: 1. but you don't know the balance of the pot account, so there is nothing to assert. That's why I believe it's not a configuration test.
4. we have tests for this. I still don't see a reason why compilation should fail.

enthusiastmartin: Personally, i don't like this to be where it is.  It is little bit confusing. 

Wouldn't be integrity test in a corresponding pallet a better place ? 

that's usually a place where you ensure the configuration of a pallet makes sense.

martinfridrich: hm, the problem is that the pallet is in the warehouse so  passing `native_existential_deposit` into it is not straightforward and changing ED to > 100 would result in the necessity to upgrade integrity test in warehouse pallet + it would lock both chains to the same max ED.

FYI this is how it's done in the basilisk.

Roznovjak: You can get the ED from `MultiCurrency::minimum_balance(native_asset_id)`.

martinfridrich: > You can get the ED from `MultiCurrency::minimum_balance(native_asset_id)`.

Thank you but my bigger issue is second part

mrq1911 (approved):

PR #534, opened 2023-03-10
feat!: Integrate EMA oracle with Omnipool Hooks

apopiak:
Add the EMA oracle pallet to the runtime and ingest data by using the `OmnipoolHooksAdapter`.

## Motivation and Context

We want to have on-chain data that aggregates over timespans. E.g. aggregate trading volume over some time to change fees based on it.

## How Has This Been Tested?

Includes an integration test.
Most of the tests are in the pallet and its corresponding math package.

## Checklist:

- [x] Run benchmarks
- [x] Decide on `SupportedPeriods`
- [x] Add hook weights to omnipool weights

+ supersedes #512 
+ based on #532 (to be merged after)
+ uses warehouse PR https://github.com/galacticcouncil/warehouse/pull/187


runtime/hydradx/src/lib.rs
```diff
@@ -851,6 +859,24 @@ impl pallet_transaction_pause::Config for Runtime {
 	type WeightInfo = weights::transaction_pause::HydraWeight<Runtime>;
 }
 
+// constants need to be in scope to use as types
+use pallet_ema_oracle::MAX_PERIODS;
+
+parameter_types! {
+	pub SupportedPeriods: BoundedVec<OraclePeriod, ConstU32<MAX_PERIODS>> = BoundedVec::truncate_from(vec![
+		OraclePeriod::LastBlock, OraclePeriod::TenMinutes, OraclePeriod::Day, OraclePeriod::Week]);
```
apopiak, line 867: Same question as for Basilisk: which periods do we want?

pallets/omnipool/src/traits.rs
```diff
@@ -56,16 +56,20 @@ where
...
+		_: Origin,
+		_: AssetInfo<AssetId, Balance>,
+		_: AssetInfo<AssetId, Balance>,
+	) -> Result<Weight, Self::Error> {
+		Ok(Weight::zero())
 	}
 
-	fn on_hub_asset_trade(_: Origin, _: AssetInfo<AssetId, Balance>) -> Result<(), Self::Error> {
-		Ok(())
+	fn on_hub_asset_trade(_: Origin, _: AssetInfo<AssetId, Balance>) -> Result<Weight, Self::Error> {
```
jak-pan, line 71: Is this triggered on every swap? Even on Asset <> Asset?

jak-pan (approved):
Nitpicks but we can fix later just push the periods we need

PR #544, opened 2023-03-23
feat(omnipool liq mining): default storage version

Roznovjak:
Set default storage version to 1.

### DIFF
--- a/pallets/omnipool-liquidity-mining/src/lib.rs
+++ b/pallets/omnipool-liquidity-mining/src/lib.rs
@@ -77,7 +77,11 @@ pub mod pallet {
 	use frame_support::pallet_prelude::*;
 	use frame_system::pallet_prelude::*;
 
+	/// The current storage version.
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
 	#[pallet::pallet]
+	#[pallet::storage_version(STORAGE_VERSION)]
 	#[pallet::generate_store(pub(crate) trait Store)]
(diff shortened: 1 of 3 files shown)


pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -77,7 +77,11 @@ pub mod pallet {
 	use frame_support::pallet_prelude::*;
 	use frame_system::pallet_prelude::*;
 
+	/// The current storage version.
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
 	#[pallet::pallet]
+	#[pallet::storage_version(STORAGE_VERSION)]
```
dmoka, line 84: Do we need to do this for every pallet where we use storage?

Roznovjak: When this pallet will be integrated somewhere for the first time, the storage version will be set to the correct version (in this case 1, instead of the default storage version 0).

martinfridrich: I don't think this pallet will be integrated anywhere else this pallet is specific to omnipool and hydra... 
Migration to v1 can't be skipped, it creates nft class

Roznovjak: Ah. Why do we update the storage version then. We don't need to increase it if we need to perform just some operations, without modifying the storage types. 
 Anyway, closing this PR.

PR #548, opened 2023-04-03
feat: add omnipool price adapter

dmoka:
## Description

Add common oracle price adapter for getting oracke price for arbitrary asset pair and oracle perior

## Checklist:


runtime/common/src/adapters.rs
```diff
@@ -138,3 +141,47 @@ where
...
+			OMNIPOOL_SOURCE,
+		);
+
+		let price_asset_a_lrna = match price_asset_a_lrna {
+			Ok(price) => price.0,
+			Err(OracleError::SameAsset) => EmaPrice::from(1),
+			Err(_) => return None,
+		};
+
+		let price_lrna_asset_b = pallet_ema_oracle::Pallet::<Runtime>::get_price(
```
apopiak, line 167: why use the pallet directly and not depend on the trait?

apopiak (approved):
looking good

Issue #529, opened 2023-03-07
`lrna_price_adjustment` can be updated to any rate without restriction


yanliu18:
There are two places where `lrna_price_adjustment` is udpated, i.e., when global farm is created and when global farm is updated.

The former will have to go though governance to make sure the global farm is reasonably created. However, the latter is under full control of the farm owner (can be anyone approved by TechnicalComittee).
The code does not implement any validation towards this exchange rate. Similarly there is no parameter validation in the underlying liquidity mining pallet.
https://github.com/galacticcouncil/HydraDX-node/blob/e7e515b9dea0e31fa6577f0df244826ff67ae845/pallets/omnipool-liquidity-mining/src/lib.rs#L391

This would allow the farm owner setting a very small rate leading to minimal rewards claimable for the LP.
It would be safer to provide a price reference such as a trusted oracle or let this update go through technical comittee (however significant delay would be expected).

PR #549, opened 2023-04-03
feat: LM use oracle for price adjustment

martinfridrich:
## Description
This PR removes possibility to manually enter and update `price_adjustment` from `omnipool-liquidity-mining`. `price_adjustment` is replaced with 10min price provided by `ema-oracle` pallet.

## Related Issue
#529 

## Checklist:
Warehouse PR should be reviewed and merged first: https://github.com/galacticcouncil/warehouse/pull/193


pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -369,7 +364,8 @@ pub mod pallet {
 				owner.clone(),
 				yield_per_period,
 				min_deposit,
-				lrna_price_adjustment,
+				//NOTE: oracle's price is used for `price_adjustment`
+				FixedU128::one(),
```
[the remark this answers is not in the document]
martinfridrich: Warehouse LM require `price_adjustment` as a param for case when oracle is not used for price adjustment and it's updated manually as on the Basilisk. 
One - value doesn't matter that much in this case as it's not used but it can't be zero because there is validation for it in the warehouse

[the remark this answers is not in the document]
enthusiastmartin: hm.. this looks like a bad workaround.  would not be better to have it as option or alternatively, having another interface where this does not need to specify  ?

i feel like this could be very misleading.

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -355,11 +350,11 @@ pub mod pallet {
 			owner: T::AccountId,
 			yield_per_period: Perquintill,
 			min_deposit: Balance,
-			lrna_price_adjustment: FixedU128,
 		) -> DispatchResult {
 			<T as pallet::Config>::CreateOrigin::ensure_origin(origin)?;
 
-			let (id, max_reward_per_period) = T::LiquidityMiningHandler::create_global_farm(
+			//NOTE: Oracle is used as `price_adjustment` provider.
+			let (id, max_reward_per_period) = T::LiquidityMiningHandler::create_global_farm_without_price_adjustment(
```
enthusiastmartin, line 357: this is bit misleading to me. Comment says that price adjustment is provided by oracle, but then method "create_without_price_adjustment" is called. 

I have quite hard time undestanding at what point the oracle is queried

martinfridrich: This pallet is built on top of warehouse LM, oracle for `price_adjustment` is plugged into warehouse LM, warehouse LM query oracle for price adjustment. 
Comment here is just so when someone look at it after few months, he knows why we are creating global farm without price adjustment param.

runtime/common/src/adapters.rs
```diff
@@ -161,3 +163,29 @@ where
...
+
+pub struct PriceAdjustmentAdapter<Runtime, LMInstance>(PhantomData<(Runtime, LMInstance)>);
+
+impl<Runtime, LMInstance> PriceAdjustment<GlobalFarmData<Runtime, LMInstance>>
+	for PriceAdjustmentAdapter<Runtime, LMInstance>
+where
+	Runtime: warehouse_liquidity_mining::Config<LMInstance>
+		+ pallet_ema_oracle::Config
+		+ pallet_omnipool_liquidity_mining::Config,
+	u32: From<<Runtime as warehouse_liquidity_mining::Config<LMInstance>>::AssetId>,
```
enthusiastmartin, line 175: why this is needed ? looks bit out of place;-)

martinfridrich: Because oracle is using `u32` as AssetId and LM has it as associated type without any restriction to u32. 
https://github.com/galacticcouncil/warehouse/blob/edbff0b40145fe0afd124fde9be3a1962c27fade/price-oracle/src/types.rs#L29
https://github.com/galacticcouncil/warehouse/blob/edbff0b40145fe0afd124fde9be3a1962c27fade/liquidity-mining/src/lib.rs#L176
You requested some changes in the warehouse PR so I can change it in the warehouse and this doesn't have to be here

enthusiastmartin (approved):

pallets/omnipool-liquidity-mining/src/lib.rs
yanliu18: @martinfridrich In the liquidity mining pallet, I could not find the implementation of using price oracle for price adjustment. The trait `` sets `price_adjustment` to 1:  
https://github.com/galacticcouncil/warehouse/blob/0047e9ceff47b2a058ae9ecc25da96d1e827a26a/liquidity-mining/src/lib.rs#L1771

Would you advise where can I find the implementation of using oracle price?

Furthermore, is it expected to use an oracle price derived from Omnipool in Omnipool Liquidity Mining? (Assuming in future case, there are data source from other pools). If so, should the price_adjustment be configured by Omnipool Liquidity Mining?

PR #552, opened 2023-04-12
fix: asset registry checks for duplicate location

mrq1911:
updates warehouse version to relevant version https://github.com/galacticcouncil/warehouse/pull/196


mrq1911:
monorepo will resolve this

PR #553, opened 2023-04-12
feat: omnipool - dynamic fees integration

enthusiastmartin:


integration-tests/src/dynamic_fees.rs
```diff
@@ -0,0 +1,393 @@
...
+				asset_fee: Permill::from_float(0.0025_f64),
+				protocol_fee: Permill::from_float(0.0005_f64),
+				timestamp: 10_u32
+			}
+		);
+	});
+}
+
+#[test]
+fn fees_should_initialize_lazyly_when_buy_happen() {
```
jak-pan, line 48:
```suggestion
fn fees_should_initialize_lazily_when_buy_happen() {
```

jak-pan:
We need to update to latest master (I guess warehouse change is there already)

integration-tests/src/dynamic_fees.rs
```diff
@@ -0,0 +1,424 @@
...
+			DOT,
+			HDX,
+			2 * DOT_UNITS,
+			0,
+		));
+
+		// REVIEW: IS THIS CORRECT? 
+		// If yes, are you testing this correctly in other tests? 
+		// looks like you are not testing the change or maybe this is a special case when not initialized?
+		assert!(hydradx_runtime::DynamicFees::current_fees(HDX).is_none());
```
jak-pan, line 41: Just this, other than that LGTM

enthusiastmartin: this is special case. it could happen on first trade when there are no data in oracle yet. 

in this case, min fee should be applied. this case is tested within dynamic fees pallet, where it ensures that min fee is returned if no data available.

in production- this wont happen there are already oracle data.

enthusiastmartin: oh yes. the min fee will be applied in that case.

I will add additional test, more obvious one.

jak-pan (approved):

pallets/omnipool/src/lib.rs
```diff
@@ -272,6 +268,7 @@ pub mod pallet {
 			asset_out: T::AssetId,
 			amount_in: Balance,
 			amount_out: Balance,
+			asset_fee: Permill,
```
mrq1911, line 271: i think it would be better to just add calculated fee and protocol fee here, since we are doing changes to this event, also which amount should be fee deducted from, naming could clear this

mrq1911 (changes requested):

pallets/omnipool/src/lib.rs
enthusiastmartin: yes makes sense. this will be needed for the change that staking needs too. 

however this require math to change and to return the value.

runtime/common/src/lib.rs
```diff
@@ -357,6 +358,23 @@ parameter_types! {
 	pub MaxNumberOfRetriesOnError: u8 = 3;
 }
 
+// Dynamic fees
+parameter_types! {
+	pub AssetFeeParams: FeeParams<Permill> = FeeParams{
+		min_fee: Permill::from_rational(25u32,10000u32),
+		max_fee: Permill::from_percent(40),
+		decay: FixedU128::zero(),
```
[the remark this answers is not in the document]
enthusiastmartin: this needs to be configured.  i believe Colin would know what is the correct acceptable value here.

pallets/omnipool/src/lib.rs
mrq1911: mby you can just change when the event is emitted - somewhere where you have all numbers in scope

mrq1911: to be clear, i dont want to be permill, but balance

pallets/omnipool/src/lib.rs
```diff
@@ -280,6 +280,10 @@ pub mod pallet {
...
+			asset_fee_amount: Balance,
+			protocol_fee: Permill,
```
mrq1911, line 285: why you are emitting both `Balance` and `Permill` in event? isn't the Balance sufficient? that is only value i am interested in

enthusiastmartin: i believe it is useful to have the exact fee as well. so when you look at your trade, you see the event and you'd have exact idea what percentage did you pay. 

the amount itself might not be that clear. i think fee percentage might give users what is the fee quicker than amount alone.

mrq1911: but it is redundant, if you have an amount you can calculate the percentage right?

mrq1911: that's why i prefer to return Balance which is exact, derived Permill doesn't have to be at there is always room for rouding issues

enthusiastmartin: removed.

mrq1911 (approved):

PR #561, opened 2023-04-24
feat!: xcm transfer rate limit pallet

jak-pan:
Integration of https://github.com/galacticcouncil/cumulus/pull/3 into Hydradx.
Provides a pallet that implements a rate limit for tokens coming in via XCM that can be configured in the asset registry.
Should not change runtime logic as the deferrer for the xcmp pallet is set to `()`.

### Limitations
- Counts accumulated amounts via `MultiLocation`s of incoming messages without reanchoring or anything. Might be worth checking whether that's an issue.
- Only processes XCMv3, though we have a test that converted XCMv2 works fine as well. Non-convertible XCMv1 or XCMv2 should not be executed by the executor anyway.
- Only tracks and limits incoming tokens, not outgoing.
- Only tracks and limits `ReserveAssetDeposited` and `ReceiveTeleportedAsset`, meaning that HDX tokens "returning" from other chains are not tracked or limited.

### Todos
- [x] write more doc comments
- [x] write readme
- ~~[ ] add integration test?~~
- [x] run benchmarks

~~requires https://github.com/galacticcouncil/warehouse/pull/200~~


rdovgan (approved):

pallets/otc/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-otc'
-version = '1.0.1'
+version = "1.0.2"
```
dmoka, line 3:
```suggestion
version = '1.0.2'
```

pallets/xcm-rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+			AccumulatedAmounts::<T>::insert(
+				location,
+				AccumulatedAmount {
+					amount: new_accumulated_amount,
+					last_updated: current_time,
+				},
+			);
+
+			if deferred_by > 0 {
```
dmoka, line 196: small thing, but here we could use a guard statement for returning None, then returning Some in the next line.

By doing so we can get rid of the `else` stament.

pallets/xcm-rate-limiter/src/benchmarking.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use super::*;
+
+use frame_benchmarking::benchmarks;
+
+use sp_std::prelude::*;
+
+benchmarks! {
+
+	asd {
```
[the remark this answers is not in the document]
apopiak: we agreed to return a manually specified weight

pallets/otc/Cargo.toml
apopiak: Btw. IIRC the convention is to use double quotes in rust Cargo.toml files

mrq1911 (changes requested):
how do you potentially stop deferred messages to be processed?

integration-tests/src/xcm_rate_limiter.rs
```diff
@@ -0,0 +1,158 @@
...
+		));
+
+		// Assert
+		assert_eq!(
+			hydradx_runtime::Balances::free_balance(&AccountId::from(ALICE)),
+			ALICE_INITIAL_NATIVE_BALANCE - amount
+		);
+	});
+
+	let fee = 400641025641;
```
mrq1911, line 147: very fragile, try to assert without need to know fee ahead of time

runtime/hydradx/src/xcm.rs
```diff
@@ -133,6 +133,10 @@ impl cumulus_pallet_xcmp_queue::Config for Runtime {
 	type ControllerOriginConverter = XcmOriginToCallOrigin;
 	type PriceForSiblingDelivery = ();
 	type WeightInfo = weights::xcmp_queue::HydraWeight<Runtime>;
+	type ExecuteDeferredOrigin = EnsureRoot<AccountId>;
```
mrq1911, line 136: so we'll have t create a referendum every time a single message is deferred?

pallets/dca/src/lib.rs
```diff
@@ -1071,10 +1071,6 @@ where
 	}
 }
 
-pub trait RelayChainBlockHashProvider {
```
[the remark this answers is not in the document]
apopiak: I can extract this part out into a separate PR if you like. It just made sense to pull the trait somewhere else to make it accessible without depending on DCA

runtime/hydradx/src/lib.rs
```diff
@@ -1096,7 +1036,9 @@ construct_runtime!(
 		OmnipoolLiquidityMining: pallet_omnipool_liquidity_mining = 63,
 		OTC: pallet_otc = 64,
 		CircuitBreaker: pallet_circuit_breaker = 65,
+		// DCA = 66
```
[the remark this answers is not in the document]
apopiak: It is not removed. It was placed further down because of ordering. I thought it made sense to add a comment here where DCA "should be" in the order if we were ordering by index.

pallets/otc/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-otc'
-version = '1.0.1'
+version = '1.0.2'
```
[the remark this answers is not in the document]
apopiak: oops, good catch

apopiak, line 3:
```suggestion
version = '1.0.1'
```

runtime/common/src/weights/xcmp_queue.rs
```diff
@@ -70,4 +70,28 @@ impl<T: frame_system::Config> WeightInfo for HydraWeight<T> {
...
+	// Proof Skipped: XcmpQueue QueueConfig (max_values: Some(1), max_size: None, mode: Measured)
+	// Storage: XcmpQueue DeferredXcmMessages (r:1 w:1)
+	// Proof Skipped: XcmpQueue DeferredXcmMessages (max_values: None, max_size: None, mode: Measured)
+	// Storage: XcmpQueue CounterForOverweight (r:1 w:1)
+	// Proof: XcmpQueue CounterForOverweight (max_values: Some(1), max_size: Some(4), added: 499, mode: MaxEncodedLen)
+	// Storage: XcmpQueue OverweightCount (r:1 w:1)
+	// Proof Skipped: XcmpQueue OverweightCount (max_values: Some(1), max_size: None, mode: Measured)
+	// Storage: XcmpQueue Overweight (r:100 w:100)
+	// Proof Skipped: XcmpQueue Overweight (max_values: None, max_size: None, mode: Measured)
+	fn service_deferred() -> Weight {
```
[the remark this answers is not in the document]
apopiak: It is quite heavy as it is hard to bound, so we assume a pretty bad worst case.
We decided it's not worth it to optimize as it will be executed by a privileged origin (the `ExecuteDeferredOrigin`).
It is this extrinsic:
https://github.com/galacticcouncil/cumulus/pull/3/files#diff-3d3febd8cfcd42ba5bb949d63fdafb9a71da3db1feb35a45601404b64fcd70c4R331

apopiak:
> i saw lot of test about deferring process, none about processing/rejecting deferred messages

do checkout the PR in cumulus that has tests for this, but I can add an integration test about execution as well.
https://github.com/galacticcouncil/cumulus/pull/3/files#diff-f516e3fdc13a7eddee189799381162efddc665fd931691277586a18bbdc93aeb

runtime/hydradx/src/xcm.rs
apopiak: Thinking about it further: If we expect to need to discard messages with short deferral periods, this might be troublesome and we might have to devise another mechanism, e.g. having a less privileged origin that can extend the deferral of deferred messages to a time that governance can properly react to it.

apopiak:
as discussed offline, split into base PR and integration. deploy Basilisk integration first

pallets/xcm-rate-limiter/Cargo.toml
```diff
@@ -0,0 +1,71 @@
+[package]
+name = "pallet-xcm-rate-limiter"
+version = "0.1.0"
+authors = ["GalacticCouncil <hydradx@galacticcouncil.io>"]
+edition = "2021"
+license = "Apache-2.0"
+homepage = 'https://github.com/galacticcouncil/hydradx-node'
+repository = 'https://github.com/galacticcouncil/hydradx-node'
+description = "Circuit breakers for AMMs"
```
enthusiastmartin, line 9: i believe this description is incorrect.

pallets/xcm-rate-limiter/Cargo.toml
```diff
@@ -0,0 +1,71 @@
...
+
+[dependencies]
+serde = { features = ["derive"], optional = true, version = "1.0.136" }
+codec = { default-features = false, features = ["derive"], package = "parity-scale-codec", version = "3.4.0" }
+scale-info = { version = "2.3.1", default-features = false, features = ["derive"] }
+
+# Hydradx
+hydradx-traits = { workspace = true }
+hydra-dx-math = { workspace = true }
+pallet-omnipool = { workspace = true }
```
enthusiastmartin, line 22: omnipool is not required afaik.

pallets/xcm-rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+			let accumulated_liquidity = AccumulatedAmounts::<T>::get(location);
+
+			// We assume that it's fine to not track assets whose id cannot be determined...
+			let Some(asset_id) = T::CurrencyIdConvert::convert(location) else {
+				total_weight.saturating_accrue(T::DbWeight::get().reads(1));
+				continue
+			};
+			// ... or that don't have a rate limit configured.
+			let Some(limit_per_duration) = T::RateLimitFor::get(&asset_id) else {
+				total_weight.saturating_accrue(T::DbWeight::get().reads(2));
```
enthusiastmartin, line 214: I assume the 1 one the read here refers to RateLimirFor::get().

Have you considered that eg T::RateLimitFor::get() can do more than 1 read?   

It is generic , so provided implementation might do even more reads.

runtime/common/src/weights/xcmp_queue.rs
```diff
@@ -70,4 +70,28 @@ impl<T: frame_system::Config> WeightInfo for HydraWeight<T> {
...
+		// Minimum execution time: 90_934_683 nanoseconds.
+		Weight::from_ref_time(91_491_151_000 as u64)
+			.saturating_add(T::DbWeight::get().reads(104 as u64))
+			.saturating_add(T::DbWeight::get().writes(103 as u64))
```
enthusiastmartin, line 87: 100 reads/writes? just looks suspicious.

runtime/hydradx/Cargo.toml
```diff
@@ -154,7 +155,6 @@ runtime-benchmarks = [
     "cumulus-pallet-xcmp-queue/runtime-benchmarks",
     "pallet-uniques/runtime-benchmarks",
     "pallet-omnipool/runtime-benchmarks",
-    "pallet-transaction-pause/runtime-benchmarks",
```
enthusiastmartin, line 157: why removed?

pallets/xcm-rate-limiter/src/lib.rs
apopiak: Yes, we considered it and decided against doing the benchmarking to ship faster.
Happy to document it more obviously if you want.

runtime/common/src/weights/xcmp_queue.rs
apopiak: Yeah, the worst case is quite terrible: it happens when all messages are overweight and involves a write for every message.

runtime/hydradx/Cargo.toml
apopiak: cause it was duplicated. If you check the file further up you find "pallet-transaction-pause/runtime-benchmarks"

pallets/xcm-rate-limiter/Cargo.toml
apopiak: good point, should be a dev dependency

runtime/common/src/weights/xcmp_queue.rs
mrq1911: what impact this will have on the cost of transfer ?

mrq1911: this has impact on production runtime even without integration enabled

mrq1911: i guess this function is called by the normal xcm inherent as well, do we account for this extra weight consumed?

mrq1911 (approved):

runtime/common/src/weights/xcmp_queue.rs
apopiak: The XCMP queue actually does not account for the weight of storing overweight messages.
So we're doing no worse than the previous implementation, but it's not great, yeah.

apopiak: > what impact this will have on the cost of transfer ?

I don't think it will have any? These XCM processing costs are eaten by the chain

PR #567, opened 2023-05-01
feat: refactor omnipool price barrier

enthusiastmartin:


pallets/omnipool/src/traits.rs
```diff
@@ -143,17 +143,28 @@ where
...
-		let max_allowed_difference = external_price.saturating_mul(max_allowed);
+
+		let oracle_price = ExternalOracle::get_price(asset_a, asset_b).map_err(|_| ())?;
+		let external_price = FixedU128::checked_from_rational(oracle_price.n, oracle_price.d).ok_or(())?;
 		let current_spot_price = FixedU128::checked_from_rational(current_price.n, current_price.d).ok_or(())?;
 
+		let max_allowed_difference = max_allowed
+			.checked_mul(
+				&current_spot_price
+					.checked_add(&external_price)
```
apopiak, line 155: I don't get it, why average the two?

enthusiastmartin: we need to calculate percentage difference between 2 values. using formula 

`| oracle price - spot price } / ( (oracle price + spot price )  / 2 ) `

pallets/omnipool/src/traits.rs
```diff
@@ -143,17 +143,28 @@ where
...
+			)
+			.ok_or(())?;
+
 		let diff = if current_spot_price >= external_price {
 			current_spot_price.saturating_sub(external_price)
 		} else {
 			external_price.saturating_sub(current_spot_price)
 		};
+
 		ensure!(diff <= max_allowed_difference, ());
```
apopiak, line 168: btw. you could pull out the `checked_div(2)` to here (multiplying it with `diff` instead)

pallets/omnipool/src/traits.rs
enthusiastmartin: > Ok, I feel like I'm getting the gist. Can you just add a test that compares this function to a simple annotated calculation?

added simple testcases

pallets/omnipool/src/traits.rs
enthusiastmartin: yes, good idea. so we avoid divisions where possible.

PR #596, opened 2023-06-01
Fix: Oracle Inversion

apopiak:
Adds logic to calculate the inverted oracle (i.e. A/B and B/A) separately and a storage migration that seeds it with the inversion of the current price.

## Motivation and Context

A simple example illustrating the non-commutability of division and addition/why we don't want to just invert the aggregated oracle price the same way we do with a spot price:

suppose the smoothing factor is 0.5
block 1: spot price is 2, inverse price is 0.5
EMA after block1: EMA = 2,     EMA_inverse = 0.5

block 2: spot price is 4, inverse spot price is 0.25
EMA = 2 * (1- 0.5) + 4 * 0.5 = 3
EMA_inverse = 0.5 (1 - 0.5) + 0.25 0.5 = 0.375

EMA =/= 1/EMA_inverse

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6745,7 +6745,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-genesis-history"
-version = "2.0.0"
+version = "2.0.1"
 dependencies = [
  "derive_more",
  "frame-support",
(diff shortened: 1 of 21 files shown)


apopiak:
The troubles induced by having 2 oracles per pair are greater than the issues with the inaccuracy, closing.

PR #599, opened 2023-06-05
fix: activate std for runtime deps in integration tests

apopiak:
Allow runtime integration tests to compile on their own.

## Description
The monorepo has lead to it being difficult to run tests for some individual crates.

Here's the way to debug/address this compile error:
```
error: cannot find macro `vec` in this scope
  --> .cargo/git/checkouts/substrate-7e08433d4c370a21/bcff60a/frame/identity/src/types.rs:68:35
   |
68 |                 let mut r: BoundedVec<_, _> = vec![0u8; n as usize - 1]
   |                                               ^^^
   |
   = note: consider importing one of these items:
           codec::alloc::vec
           scale_info::prelude::vec
           sp_std::vec

error: could not compile `pallet-identity` due to 4 previous errors
```
The issue is here that the `std` feature is not activated for `pallet-identity`. (This is why I'm such a stickler about it when reviewing PRs).
`cargo tree` allows you to identify which crates are depending on `pallet-identity`.
```cargo tree -p pallet-identity -i```
results in the following for me:
```
pallet-identity v4.0.0-dev (https://github.com/paritytech/substrate?branch=polkadot-v0.9.38#bcff60a2)
├── common-runtime v102.3.11 
│   ├── [...elided...]
├── hydradx-runtime v156.0.0 (*)
├── kusama-runtime v0.9.38 (https://github.com/galacticcouncil/polkadot?branch=fix-xcm-executor-atomicity-9-38#a225599a)
│   ├── [...elided...]
├── polkadot-runtime v0.9.38 (https://github.com/galacticcouncil/polkadot?branch=fix-xcm-executor-atomicity-9-38#a225599a)
│   ├── [...elided...]
└── testing-hydradx-runtime v156.0.0 (*)
```
So now comes the tedious work: You'll need to check all dependents of pallet-identity.
In our example here you'll want to check `common-runtime` and `hydradx-runtime` whether they activate the `std` feature correctly. And then the crates that depend on `common-runtime` etc. (in this case the integration tests didn't have std for the runtime deps.)

## Checklist:


Roznovjak (approved):
Bump the crate version pls

Issue #589, opened 2023-05-24
configure XCM safe call filter


mrq1911:
The comment explains how it should be determined:
https://github.com/paritytech/polkadot/blob/e05b78689d3fb0deb555ede9f0f0e191cc5e18b5/runtime/polkadot/src/xcm_config.rs#L177-L183

> Calls that are allowed through this filter must:
> 1. Have a fixed weight;
> 2. Cannot lead to another call being made;
> 3. Have a defined proof size weight, e.g. no unbounded vecs in call parameters.

_Originally posted by @apopiak in https://github.com/galacticcouncil/HydraDX-node/pull/546#discussion_r1202403932_

PR #602, opened 2023-06-05
feat: xcm transact call filter

Roznovjak:
Fixes: #589


runtime/hydradx/src/xcm.rs
```diff
@@ -309,3 +309,92 @@ pub type LocalAssetTransactor = ReroutingMultiCurrencyAdapter<
...
+				return true;
+			}
+		}
+
+		// check the runtime call filter
+		if !CallFilter::contains(call) {
+			return false;
+		}
+
+		matches!(
```
jak-pan, line 335: I think this is very extreme. This will disallow some use-cases which we already "support". (Like providing liq)

jak-pan: Oh... It's actually allowed list right? Very confusing with the thing above !

Roznovjak: Yes. The extrinsics in the macro are allowed.

runtime/hydradx/src/xcm.rs
```diff
@@ -309,3 +309,92 @@ pub type LocalAssetTransactor = ReroutingMultiCurrencyAdapter<
...
+			call,
+			RuntimeCall::System(frame_system::Call::kill_prefix { .. } | frame_system::Call::set_heap_pages { .. })
```
jak-pan, line 337: I think we might not want to allow this (It is probably done because of the switch to governance parachain should be able to do this but we don't have external entity that could call this AFAIK) cc @mrq1911

runtime/hydradx/src/xcm.rs
```diff
@@ -309,3 +309,92 @@ pub type LocalAssetTransactor = ReroutingMultiCurrencyAdapter<
...
+	fn contains(call: &RuntimeCall) -> bool {
+		#[cfg(feature = "runtime-benchmarks")]
+		{
+			if matches!(call, RuntimeCall::System(frame_system::Call::remark_with_event { .. })) {
+				return true;
+			}
+		}
+
+		// check the runtime call filter
+		if !CallFilter::contains(call) {
```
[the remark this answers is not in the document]
Roznovjak: It’s correct. We want to return false from the if statement if the filter returns false. So we need to negate the result from the filter to pass the logical if test

runtime/hydradx/src/xcm.rs
Roznovjak: My idea behind this filter is to list all possible extrinsics that pass requirements (fixed weight, no unbound vecs in params) and combine it with the runtime call filter. So the result are extrinsics that we want to allow, and can be executed

runtime/hydradx/src/xcm.rs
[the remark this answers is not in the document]
apopiak: Ah and you need the double negation because you want to only do an early return if the CallFitler returns false.

runtime/hydradx/src/xcm.rs
```diff
@@ -309,3 +309,92 @@ pub type LocalAssetTransactor = ReroutingMultiCurrencyAdapter<
...
+						| pallet_uniques::Call::force_item_status { .. }
+						| pallet_uniques::Call::set_attribute { .. }
+						| pallet_uniques::Call::clear_attribute { .. }
+						| pallet_uniques::Call::set_metadata { .. }
+						| pallet_uniques::Call::clear_metadata { .. }
+						| pallet_uniques::Call::set_collection_metadata { .. }
+						| pallet_uniques::Call::clear_collection_metadata { .. }
+						| pallet_uniques::Call::set_accept_ownership { .. }
+						| pallet_uniques::Call::set_price { .. }
+						| pallet_uniques::Call::buy_item { .. },
```
apopiak, line 373: So uniques has calls we need to filter?

apopiak (approved):
LGTM

runtime/hydradx/src/xcm.rs
```diff
@@ -309,3 +309,92 @@ pub type LocalAssetTransactor = ReroutingMultiCurrencyAdapter<
...
+					| pallet_identity::Call::quit_sub { .. },
+			) | RuntimeCall::Omnipool(..)
+				| RuntimeCall::OmnipoolLiquidityMining(..)
+				| RuntimeCall::OTC(..)
+				| RuntimeCall::CircuitBreaker(..)
+				| RuntimeCall::DCA(..)
+				| RuntimeCall::MultiTransactionPayment(..)
+				| RuntimeCall::Currencies(..)
+				| RuntimeCall::Tokens(..)
+				| RuntimeCall::OrmlXcm(..)
```
apopiak, line 397: Actually looked at this again and not sure we want to allow it? (It's forbidden by the CallFilter anyway, though.)

apopiak: Oh, no I mean I'm not sure this filter should allow it. Are extrinsics accepting `Box<VersionedXcm>` well defined size-wise?

Roznovjak: What do you think is the problem? Missing `MaxEncodedLen`?

apopiak: Checked and it does, so it's fine I guess.

PR #605, opened 2023-06-06
feat: dca slippages and retries configurable

dmoka:
Changes:

- [x] make max number of retries configurable
- [x] make price stability threshold configurable
- [x] pull out slippage from order to Schedule as it is a generic parameter
- [x] rename order `min_limit` to `min_amount_out`
- [x] rename order `max_amount_in` to `max_amount_in`
- [x] fix edge case for sell: we replan a schedule only when the budget leftover is bigger than 5 * FEE

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3837,7 +3837,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "157.0.0"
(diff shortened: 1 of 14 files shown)


mrq1911 (changes requested):
bump runtime!

mrq1911 (approved):

PR #609, opened 2023-06-11
refactor: remove common runtime

Roznovjak:


apopiak (commented):
looks fine, not a huge fan of splitting the runtime/lib.rs (because I like doing ctrl+F in the file 😅 ), but might be sensible

mrq1911 (approved):

PR #611, opened 2023-06-15
feat: add randomness for searching free block

dmoka:
- [x] add validation for min trade amount > 20*fee
- [x] double free block search range
- [x] add randomness for scheduling


pallets/dca/src/lib.rs
```diff
@@ -911,13 +915,30 @@ where
 	fn find_next_free_block(blocknumber: T::BlockNumber) -> Result<T::BlockNumber, DispatchError> {
 		let mut next_execution_block = blocknumber;
 
-		// In a bound fashion, we search for next free block with the delays of 1 - 2 - 4 - 8 - 16.
-		for retry_index in 0u32..=RETRY_TO_SEARCH_FOR_FREE_BLOCK {
+		let mut randomness_generator =
+			Self::get_randomness_generator(frame_system::Pallet::<T>::current_block_number());
+
+		let radius_ranges = [
```
mrq1911, line 921: you dont have to store this in code, it is simple loop where ranges are `(2^i, 2^i+1 - 1)`

pallets/dca/src/lib.rs
```diff
@@ -911,13 +915,30 @@ where
 	fn find_next_free_block(blocknumber: T::BlockNumber) -> Result<T::BlockNumber, DispatchError> {
 		let mut next_execution_block = blocknumber;
 
-		// In a bound fashion, we search for next free block with the delays of 1 - 2 - 4 - 8 - 16.
-		for retry_index in 0u32..=RETRY_TO_SEARCH_FOR_FREE_BLOCK {
+		let mut randomness_generator =
```
mrq1911, line 918: don't create new instance of randomenes generator, reuse the one you already have!

mrq1911: besides like this you will generate exactly same numbers for every schedule in the block...

mrq1911 (changes requested):

pallets/dca/src/lib.rs
```diff
@@ -430,7 +431,14 @@ pub mod pallet {
 
 			let blocknumber_for_first_schedule_execution = Self::get_next_execution_block(start_execution_block)?;
 
-			Self::plan_schedule_for_block(&who, blocknumber_for_first_schedule_execution, next_schedule_id)?;
+			let mut randomness_generator =
```
mrq1911, line 434: this leads to the same generator created with every schedule call, therefore every schedule in single block searches for the same blocks. you have to add some salt to the seed unique to the schedule to make sure numbers generated are different, probably `next_schedule_id` could be used

weights-pallet.rs
```diff
@@ -0,0 +1,325 @@
+// This file is part of HydraDX.
```
mrq1911, line 1: i guess this should be moved to runtime

pallets/dca/src/lib.rs
mrq1911: alternative solution could be to re-use generator created by dca `on_initialize` and just continue generating number with that one

mrq1911 (changes requested):
bump runtime version

mrq1911 (approved):

PR #617, opened 2023-06-19
fix: oracle RV audit feedback

apopiak:
Various small improvements.
- add `Short` period to `all_periods`
- remove `MAX_UNIQUE_ENTRIES` and `trades_estimation.py`
- rename `timestamp` field to `updated_at` and add constructor
- add `inverted` function to `Ratio`
- add some comments


pallets/ema-oracle/src/types.rs
```diff
@@ -35,33 +35,50 @@ pub type Balance = u128;
...
 pub struct OracleEntry<BlockNumber> {
 	pub price: Price,
 	pub volume: Volume<Balance>,
 	pub liquidity: Liquidity<Balance>,
-	pub timestamp: BlockNumber,
+	pub updated_at: BlockNumber,
+}
+
+impl<BlockNumber> OracleEntry<BlockNumber> {
+	/// Construct a new `OracleEntry`. Will invert the price to determine `inverted_price`.
```
Roznovjak, line 50: I couldn't find any mention of `inverted_price` in this file. Could you clarify it to me please?
We can also place this method next to other methods defined for the struct.

apopiak: oops, copy pasta error, good catch 👍

mrq1911 (approved):

apopiak:
> I wonder whether we need to change something there due to this current PR. What do you think?

As discussed offline, not applicable for this PR.

PR #619, opened 2023-06-21
fix: router ED leftover balance check

dmoka:
- [x] fix bug for the case when after a trade there is less than the existential deposit as user balance
- [x] use more descriptive errors instead of UnexpectedError

### DIFF
--- a/pallets/route-executor/src/lib.rs
+++ b/pallets/route-executor/src/lib.rs
@@ -336,12 +336,8 @@ impl<T: Config> Pallet<T> {
 	) -> Result<(), DispatchError> {
 		if spent_amount < user_balance_of_asset_in_before_trade {
 			let user_balance_of_asset_in_after_trade = T::Currency::reducible_balance(asset_in, &who, true);
-			let user_expected_balance_of_asset_in_after_trade = user_balance_of_asset_in_before_trade
-				.checked_sub(&spent_amount)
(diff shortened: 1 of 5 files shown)


pallets/route-executor/src/lib.rs
```diff
@@ -136,8 +136,10 @@ pub mod pallet {
 		RouteHasNoTrades,
 		///The user has not enough balance to execute the trade
 		InsufficientBalance,
-		///Unexpected error which should never really happen, but the error case must be handled to prevent panics.
-		UnexpectedError,
+		///The balance after the trade is not as expected according to the trade amount
+		PostBalanceCheckError,
```
mrq1911, line 140: Error at the end is bit redundant

mrq1911: i would call this InvalidRouteExecution instead

pallets/route-executor/src/lib.rs
```diff
@@ -136,8 +136,10 @@ pub mod pallet {
...
+		///The calculation of route trade amounts failed in the underlying AMM
+		AmmTradeAmountCalculationError,
```
mrq1911, line 142: RouteCalculationFailed

mrq1911 (changes requested):

mrq1911 (approved):

PR #623, opened 2023-07-03
feat: update xcm config

Roznovjak:
Based on a discussion with @apopiak, where we reviewed the XCM changes proposed by the Substrate builders programme in this PR #591, we decided to make the following changes:
1. Use `WithComputedOrigin` derivative barrier with `AllowTopLevelPaidExecutionFrom` and `AllowSubscriptionsFrom` barriers. This outer barrier computes the effective origin (checks for any origin altering operations) and passes this updated origin to inner barriers.
2. Change `XcmExecuteFilter` from `Everything` to `Nothing`. This change disallows `pallet_xcm::execute` extrinsic which allows sending arbitrary XCM messages. We are not aware of any issues with this param set to `Everything`, but better be sure. Moreover, `pallet_xcm` is blocked by the runtime call filter.


enthusiastmartin:
Can you add description why the computedorigin is required ?

runtime/hydradx/src/xcm.rs
```diff
@@ -192,7 +199,7 @@ impl pallet_xcm::Config for Runtime {
 	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
 	type XcmRouter = XcmRouter;
 	type ExecuteXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
-	type XcmExecuteFilter = Everything;
+	type XcmExecuteFilter = Nothing;
```
mrq1911, line 202: @apopiak @dmoka this effectively breaks the exchange and send back assets scenario correct?

mrq1911: if this is the case i would reconsider this change

apopiak: I would say it makes sense to define the more permissive filter in the context of the ExchangeAsset stuff, not here.

runtime/hydradx/src/xcm.rs
```diff
@@ -35,11 +35,18 @@ pub type LocalOriginToLocation = SignedToAccountId32<RuntimeOrigin, AccountId, R
...
-	AllowSubscriptionsFrom<Everything>,
+	// Evaluate the barriers with the effective origin
+	WithComputedOrigin<
+		(
+			AllowTopLevelPaidExecutionFrom<Everything>,
+			// Subscriptions for version tracking are OK.
+			AllowSubscriptionsFrom<Everything>,
+		),
+		UniversalLocation,
+		ConstU32<1>,
```
apopiak, line 48: Seems like 1 should be fine, but I don't see a reason to deviate from the ecosystem which is going for 8.

apopiak (approved):
looks fine after adjusting computed origin count

runtime/hydradx/src/xcm.rs
apopiak, line 48:
```suggestion
		ConstU32<8>,
```

PR #629, opened 2023-07-04
feat: stableswap gradual amplification change

enthusiastmartin:
This PR introduces gradual change of amplification for stableswap pool.


pallets/stableswap/src/lib.rs
```diff
@@ -326,59 +342,94 @@ pub mod pallet {
...
+		///
+		/// Emits `AmplificationUpdated` event if successful.
+		#[pallet::call_index(2)]
+		#[pallet::weight(<T as Config>::WeightInfo::update_pool())]
+		#[transactional]
+		pub fn update_amplification(
+			origin: OriginFor<T>,
+			pool_id: T::AssetId,
+			future_amplification: u16,
+			future_timestamp: T::BlockNumber,
```
mrq1911, line 404: Please stop referring to the block number as the timestamp. i recommend the same naming as I proposed for `AmplificationChanging` event. i would name it timestamp only in the case we really used timestamp

pallets/stableswap/src/lib.rs
```diff
@@ -326,59 +342,94 @@ pub mod pallet {
...
+		/// - `origin`: Must be T::AuthorityOrigin
+		/// - `pool_id`: pool to update
+		/// - `future_amplification`: new desired pool amplification
+		/// - `future_timestamp`: future block number when the amplification is updated
+		///
+		/// Emits `AmplificationUpdated` event if successful.
+		#[pallet::call_index(2)]
+		#[pallet::weight(<T as Config>::WeightInfo::update_pool())]
+		#[transactional]
+		pub fn update_amplification(
```
mrq1911, line 400: `start_block` should also be definable, as this gives you the ability to exactly time the change

pallets/stableswap/src/lib.rs
```diff
@@ -208,6 +212,13 @@ pub mod pallet {
 			asset_id: T::AssetId,
 			state: Tradability,
 		},
+
+		///
+		AmplificationUpdated {
```
mrq1911, line 217: ```AmplificationChanging { pool_id, start_block, end_block, current_amplification, final_amplification }```

math/src/stableswap/math.rs
```diff
@@ -346,6 +347,40 @@ pub(crate) fn calculate_y<const N: u8>(xp: &[Balance], d: Balance, amplification
...
+	future_timestamp: u128,
+	current_timestamp: u128,
+) -> u128 {
+	// short circuit if timestamp are invalid
+	if current_timestamp < initial_timestamp || future_timestamp < initial_timestamp {
+		return initial_amplification;
+	}
+
+	// Short circuit if timestamp or amplification are equal
+	if initial_timestamp == future_timestamp || initial_amplification == future_amplification {
```
mrq1911, line 363: this block is redundant, the following condition `current_timestamp >= future_timestamp` is satisfied in the same case

pallets/stableswap/src/lib.rs
```diff
@@ -313,6 +327,8 @@ pub mod pallet {
 		) -> DispatchResult {
 			T::AuthorityOrigin::ensure_origin(origin)?;
 
+			let amplification = NonZeroU16::new(amplification).ok_or(Error::<T>::InvalidAmplification)?;
+
 			let pool_id = Self::do_create_pool(share_asset, &assets, amplification, trade_fee, withdraw_fee)?;
```
mrq1911, line 332: should emit the event `AmplificationChanging` for consistency

mrq1911 (changes requested):

pallets/stableswap/src/lib.rs
enthusiastmartin: i dont think this is necessary. It creates pool with given amplification. It is not going to change.

mrq1911: yeah, but if you hook on events to get fees, you will miss the initial setting

mrq1911: not like we are creating that many pools so we cannot afford extra event

pallets/stableswap/src/lib.rs
```diff
@@ -326,59 +344,99 @@ pub mod pallet {
...
+		///
+		/// Parameters:
+		/// - `origin`: Must be T::AuthorityOrigin
+		/// - `pool_id`: pool to update
+		/// - `future_amplification`: new desired pool amplification
+		/// - `future_block`: future block number when the amplification is updated
+		///
+		/// Emits `AmplificationUpdated` event if successful.
+		#[pallet::call_index(2)]
+		#[pallet::weight(<T as Config>::WeightInfo::update_pool())]
```
mrq1911, line 400: is the weight intentionally reused? i am ok with that as weight will be pretty much same and it is governance only action anyway

pallets/stableswap/src/lib.rs
```diff
@@ -281,6 +294,9 @@ pub mod pallet {
 
 		/// Not allowed to perform an operation on given asset.
 		NotAllowed,
+
+		/// Future block number is in the past.
+		InvalidBlock,
```
mrq1911, line 299: according to the error description provided, error could be named for example `PastBlock` instead of general `Invalid`

mrq1911 (approved):

pallets/stableswap/src/lib.rs
```diff
@@ -416,9 +423,18 @@ pub mod pallet {
 			Pools::<T>::try_mutate(pool_id, |maybe_pool| -> DispatchResult {
 				let mut pool = maybe_pool.as_mut().ok_or(Error::<T>::PoolNotFound)?;
 
+				ensure!(
+					pool.final_block <= current_block,
+					Error::<T>::AmplificationChangeNotCompleted
```
mrq1911, line 428: we actually may want do do this, to cancel or start reverting some application as change could take months and situation may change to adjust it

pallets/stableswap/src/lib.rs
```diff
@@ -416,9 +423,18 @@ pub mod pallet {
...
+				);
+				ensure!(
+					pool.final_amplification.get() != final_amplification,
+					Error::<T>::SameAmplification
```
mrq1911, line 432: what if you want to speed up or slow down ramp up? this check seems not very effective as it could be bypassed by just changing final amp by 1

pallets/stableswap/src/lib.rs
enthusiastmartin: added

pallets/stableswap/src/lib.rs
enthusiastmartin: benchmakrs refactor still under way, it will change accoardingly

pallets/stableswap/src/lib.rs
enthusiastmartin: i was considering this. but as it is - the initial_amp would just to final. 

but now that i think of it - i can actually compute the correct amplification and set as initial.

pallets/stableswap/src/lib.rs
enthusiastmartin: this just ensures that you dont set the same value as there is now as it does not make any change

mrq1911: it does, as you can set different end block therefore slow down or hasten the change

pallets/stableswap/src/lib.rs
mrq1911: why would it jump ? you can configure it so it doesn't

pallets/stableswap/src/lib.rs
enthusiastmartin: yes. now it is different - this was a case when you had to wait for previous change to finish. 

now it makes sense to ensure that the new final amplification != initial ampfliciation, isnt it ?

PR #631, opened 2023-07-05
feat: stableswap whitelist pool account to exclude from dusting [A4]

enthusiastmartin:


Roznovjak (approved):
LGTM, but don't forget to bump the crate version before merging it.

PR #632, opened 2023-07-06
fix: stableswap add liquidity does not accept same assets [A3]

enthusiastmartin:


pallets/stableswap/src/lib.rs
```diff
@@ -305,6 +305,9 @@ pub mod pallet {
 
 		/// New amplification is equal to the previous value.
 		SameAmplification,
+
+		/// Provided list of assets contains same assets.
+		SameAssets,
```
Roznovjak, line 310: We can reuse existing `IncorrectAssets` error.

Roznovjak (approved):

PR #633, opened 2023-07-06
fix: stableswap math - ensure non zero balance assets [A5]

enthusiastmartin:


math/src/stableswap/tests/two_assets.rs
```diff
@@ -24,6 +24,12 @@ fn test_d_with_zero_reserves() {
 	assert_eq!(calculate_d::<D_ITERATIONS>(&reserves, 1), Some(0u128));
 }
 
+#[test]
+fn test_d_with_one_zero_reserves() {
```
Roznovjak, line 28: Maybe add the same test for `calculate_y`.

Roznovjak (approved):

math/src/stableswap/tests/two_assets.rs
enthusiastmartin: calculate_y is tested within other tests where it is usually tested with more context.

PR #634, opened 2023-07-07
feat: stableswap add limit to remove liquidity [A6]

enthusiastmartin:


pallets/stableswap/src/lib.rs
```diff
@@ -179,15 +179,14 @@ pub mod pallet {
...
+			assets: Vec<AssetBalance<T::AssetId>>,
 		},
 		/// Liquidity removed.
 		LiquidityRemoved {
 			pool_id: T::AssetId,
 			who: T::AccountId,
 			shares: Balance,
-			asset: T::AssetId,
-			amount: Balance,
+			amounts: Vec<AssetBalance<T::AssetId>>,
```
Roznovjak, line 189: This change is not necessary

pallets/stableswap/src/lib.rs
```diff
@@ -565,15 +569,19 @@ pub mod pallet {
 			)
 			.ok_or(ArithmeticError::Overflow)?;
 
+			ensure!(amount >= min_amount_out, Error::<T>::MinimumAmountNotReached);
```
Roznovjak, line 572: Is 0 a valid value for the amount? Can it be 0?

pallets/stableswap/src/lib.rs
enthusiastmartin: it is preparation for remove_liquidity where you obtain all assets. This functionality will be added later. 

so just to make sure that if anything/anyone already works with events, there wont be breaking changes.

pallets/stableswap/src/lib.rs
enthusiastmartin: you can specify 0 as min amount. up to you.

Roznovjak: If a user can receive zero amount, then ok.

enthusiastmartin: well, then we have some issue somewhere of he specify so very low amount that he does not get anything.

same thing is in omnipool and xyk.

PR #638, opened 2023-07-12
feat!: XCM Rate Limiter Integration

apopiak:
Adds the XCM rate limit pallet from #561 into the runtime.
Updates the XCMP queue pallet to the version from https://github.com/galacticcouncil/cumulus/pull/4

## Checklist:


mrq1911 (changes requested):
lets test this on basilisk 1st

integration-tests/src/xcm_rate_limiter.rs
```diff
@@ -0,0 +1,253 @@
...
+	use cumulus_pallet_xcmp_queue::Event;
+	use hydradx_runtime::RuntimeEvent;
+	let RuntimeEvent::XcmpQueue(Event::XcmpMessageSent { message_hash }) = &last_hydra_events(n)[0] else {
+		panic!("expecting to find message sent event");
+	};
+	*message_hash
+}
+
+#[test]
+fn xcm_rate_limiter_should_limit_aca_when_limit_is_exceeded() {
```
jak-pan, line 23: I would probably also test clearing of the limits in similar fashion after blocks pass

jak-pan:
Hmm where is the latest version of rate limiter? @apopiak

PR #640, opened 2023-07-13
feat: XCM exchange asset

dmoka:
## Description

Implementing the handling of ExchanceAsset xcm instruction

Adding custom XcmExecuteFilter to allow Transfer-And-Swap functionality and filter out non-desired xcm messages.

NOTE: The filter and exchange asset won't be configured for now in hydra runtime. First we want to enable this in basilisk to test it out.

Leftover TODOs:
- [x] integration test for transfer and swap including 4 hops
- [x] clippy and version updates


enthusiastmartin:
Main issue i have here that i am not a fan of this feature being part of omnipool pallet. 

it feels like it should be somwhere else.

runtime/adapters/src/xcm_exchange.rs
```diff
@@ -0,0 +1,123 @@
...
+
+impl<Runtime, TempAccount, CurrencyIdConvert, Currency, DefaultPoolType> AssetExchange
+	for XcmAssetExchanger<Runtime, TempAccount, CurrencyIdConvert, Currency, DefaultPoolType>
+where
+	Runtime: pallet_route_executor::Config,
+	TempAccount: Get<Runtime::AccountId>,
+	CurrencyIdConvert: Convert<MultiAsset, Option<Runtime::AssetId>>,
+	Currency: MultiCurrency<Runtime::AccountId, CurrencyId = Runtime::AssetId, Balance = Runtime::Balance>,
+	Runtime::Balance: From<u128> + Zero + Into<u128>,
+	DefaultPoolType: Get<PoolType<Runtime::AssetId>>,
```
enthusiastmartin, line 23: the default prefix is not really necessary imo, just PoolType.

runtime/adapters/src/xcm_exchange.rs
```diff
@@ -0,0 +1,123 @@
...
+					asset_out,
+					amount.into(),
+					max_sell_amount.into(),
+					vec![Trade {
+						pool: DefaultPoolType::get(),
+						asset_in,
+						asset_out,
+					}],
+				)?;
+				let mut assets = sp_std::vec::Vec::new();
```
enthusiastmartin, line 105: with_capacity(2) instead  ?

99% of the time, there will be leftover and there will be 2 assets to return.

runtime/adapters/src/xcm_execute_filter.rs
```diff
@@ -0,0 +1,87 @@
...
+where
+	MaxXcmDepth: Get<u16>,
+	MaxInstructions: Get<u16>,
+{
+	if depth > MaxXcmDepth::get() {
+		return false;
+	}
+	let instructions_count = instructions;
+	let iter = xcm.inner().iter();
+	for inst in iter {
```
enthusiastmartin, line 47:
```suggestion
	for inst in xcm.inner().iter() {
```

runtime/adapters/src/xcm_execute_filter.rs
```diff
@@ -0,0 +1,87 @@
...
+			Either::Right(xcm) => {
+				if check_instructions_recursively::<MaxXcmDepth, MaxInstructions, ()>(
+					xcm,
+					depth + 1,
+					instructions_count,
+				) {
+					continue;
+				} else {
+					return false;
+				}
```
enthusiastmartin, line 65:
```suggestion
				if !check_instructions_recursively::<MaxXcmDepth, MaxInstructions, ()>(
					xcm,
					depth + 1,
					instructions_count,
				) {
					return false;
				}
```

enthusiastmartin (commented):
looks good to me. just some minor suggestions.

runtime/adapters/src/xcm_exchange.rs
```diff
@@ -0,0 +1,123 @@
...
+					vec![Trade {
+						pool: DefaultPoolType::get(),
+						asset_in,
+						asset_out,
+					}],
+				)?;
+				debug_assert!(
+					Currency::free_balance(asset_in, &account) == Runtime::Balance::zero(),
+					"Sell should not leave any of the incoming asset."
+				);
```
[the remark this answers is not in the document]
enthusiastmartin: so what happens if there is leftover ? it stays in the temp account ?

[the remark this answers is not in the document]
apopiak: I guess it does. Free money for the next guy 😛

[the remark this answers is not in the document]
enthusiastmartin: that was my next question : next guy could receive more in such case.

i agree that it is programming error in the amm and likely wont happen. 

i am in favor of  ensuring that the temp account is always empty after asset exchange.

Not sure though what is the best to do here.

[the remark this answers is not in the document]
apopiak: mmh, I guess we could burn?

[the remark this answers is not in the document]
apopiak: To the treasury? :-D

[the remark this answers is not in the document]
enthusiastmartin: i guess just leave as is. 

if the issue is in the amm, it can be simply exploit by direclty calling sell - although we would never notice this

runtime/adapters/src/xcm_execute_filter.rs
```diff
@@ -0,0 +1,84 @@
...
+
+		let instructions_count = Cell::new(0u16);
+		check_instructions_recursively::<MaxXcmDepth, MaxInstructions, RuntimeCall>(xcm, 0, &instructions_count)
+	}
+}
+
+fn check_instructions_recursively<MaxXcmDepth, MaxInstructions, RuntimeCall>(
+	xcm: &Xcm<RuntimeCall>,
+	depth: u16,
+	instructions: &Cell<u16>,
```
apopiak, line 36:
```suggestion
	instructions_count: &Cell<u16>,
```

runtime/adapters/src/xcm_execute_filter.rs
```diff
@@ -0,0 +1,84 @@
...
+) -> bool
+where
+	MaxXcmDepth: Get<u16>,
+	MaxInstructions: Get<u16>,
+{
+	if depth > MaxXcmDepth::get() {
+		return false;
+	}
+	let instructions_count = instructions;
```
apopiak, line 45:
```suggestion
```

apopiak (commented):
seems like we don't need to rename the parameter, no?

runtime/adapters/src/xcm_execute_filter.rs
```diff
@@ -0,0 +1,84 @@
+use sp_std::cell::Cell;
+use sp_std::marker::PhantomData;
+
+use frame_support::traits::Contains;
+use polkadot_xcm::v3::prelude::*;
+use sp_core::Get;
+use sp_runtime::Either;
+
+pub struct AllowTransferAndSwap<MaxXcmDepth, MaxInstructions, RuntimeCall>(
```
apopiak, line 9: let's a description

apopiak:
> lgtm.
> 
> although , one question : how this is integrated into runtime ? i see no changes in the runtime xcm config but there are already integration tests.

It's not yet. Going to basilisk first (see https://github.com/galacticcouncil/Basilisk-node/pull/639). Integration will look very similar.

apopiak:
This is the implementation, but we want to test on Basilisk first, so it doesn't actually add it to the runtime, yet. I Imagine it will be a two step process with another PR. Same pattern as XCM rate limiter.

apopiak:
> seems like we don't need to redefine the parameter, no?

this was about the particular review comments I left about implementation details, don't worry about it.

PR #642, opened 2023-07-15
fix: genesis history bounded storage

Roznovjak:
This PR removes `without_storage_info` from the genesis history pallet and adds a storage migration fixing the corrupted storage value.


pallets/genesis-history/src/migration.rs
```diff
@@ -0,0 +1,71 @@
...
+// limitations under the License..
+
+pub use crate::*;
+pub use frame_support::{
+	log,
+	traits::{Get, StorageVersion},
+	weights::Weight,
+};
+
+// This migration fixes the corrupted value in the storage and fixes it
```
apopiak, line 25: so the current values are not correct?

apopiak:
looks fine except that I don't feel I can judge whether the genesis hash migration stuff is correct

mrq1911 (approved):

pallets/genesis-history/src/migration.rs
Roznovjak: Yes. We manually set the storage value to something, but the storage has different data type.

PR #645, opened 2023-07-19
feat: bonds

Roznovjak:


pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+		///
+		#[pallet::call_index(0)]
+		#[pallet::weight(<T as Config>::WeightInfo::issue())]
+		pub fn issue(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+			amount: T::Balance,
+			maturity: Moment,
+		) -> DispatchResult {
+			let who = T::IssueOrigin::ensure_origin(origin)?;
```
mrq1911, line 217: anyone can issue bonds

pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+		/// Parameters:
+		/// - `origin`: needs to be `T::UnlockOrigin`
+		/// - `bond_id`: the asset id of the bonds
+		///
+		/// Emits `BondsUnlocked` event when successful.
+		///
+		#[pallet::call_index(2)]
+		#[pallet::weight(<T as Config>::WeightInfo::redeem())]
+		pub fn unlock(origin: OriginFor<T>, bond_id: T::AssetId) -> DispatchResult {
+			T::UnlockOrigin::ensure_origin(origin)?;
```
mrq1911, line 312: noone can unlock bonds, bonds are unlocked only by maturity, was this feature requested?

pallets/bonds/src/lib.rs
Roznovjak: ok. I will remove the config param. The idea was to have it configurable, but configure it so that everyone can do that.

pallets/bonds/src/lib.rs
mrq1911: also the implementation is wrong, as you effectivelly transform one bond to another by changing its maturity

pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+
+			let fee = T::ProtocolFee::get().mul_ceil(amount); // TODO: check
+			let amount_without_fee = amount.checked_sub(&fee).ok_or(ArithmeticError::Overflow)?;
+			let pallet_account = Self::pallet_account_id();
+
+			T::Currency::transfer(asset_id, &who, &pallet_account, amount_without_fee)?;
+			T::Currency::transfer(asset_id, &who, &T::FeeReceiver::get(), fee)?;
+			T::Currency::deposit(bond_asset_id, &who, amount_without_fee)?;
+
+			Bonds::<T>::insert(
```
mrq1911, line 237: not every issuance creates new token
you can reissue more of the same bonds, bond token is effectively tied to the pair of asset & maturity, therefore when you issue HDX with the same maturity there will be just more of the same tokens issued

mrq1911 (changes requested):

pallets/bonds/src/lib.rs
Roznovjak: No this feature wasn't requested. But I think it might be helpful when someone accidentally locks some significant amount of tokens with wrongly set maturity. But we can remove it if you want.

pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+
+			Bonds::<T>::try_mutate_exists(bond_id, |maybe_bond_data| -> DispatchResult {
+				let bond_data = maybe_bond_data.as_mut().ok_or(Error::<T>::BondNotRegistered)?;
+
+				let now = T::TimestampProvider::now();
+				ensure!(now >= bond_data.maturity, Error::<T>::BondNotMature);
+
+				T::Currency::withdraw(bond_id, &who, amount)?;
+
+				bond_data.amount = bond_data.amount.checked_sub(&amount).ok_or(ArithmeticError::Overflow)?;
```
mrq1911, line 282: no need to track this, issuance of the token provides this info

pallets/bonds/src/lib.rs
mrq1911: i guess you need to have storage `(asset, maturity) -> bondId` to be able to do this as well

pallets/bonds/src/lib.rs
mrq1911: but we could keep it if we plan to enable this for others only later

pallets/asset-registry/src/lib.rs
```diff
@@ -590,6 +590,13 @@ impl<T: Config> ShareTokenRegistry<T::AssetId, Vec<u8>, T::Balance, DispatchErro
 	}
 }
 
+impl<T: Config> BondRegistry<T::AssetId, Vec<u8>, T::Balance, DispatchError> for Pallet<T> {
+	fn create_bond_asset(existential_deposit: T::Balance) -> Result<T::AssetId, DispatchError> {
+		let bounded_name: BoundedVec<u8, T::StringLimit> = Self::to_bounded_name(Vec::<u8>::new())?;
+		Self::register_asset(bounded_name, AssetType::Bond, existential_deposit, None, None)
+	}
+}
+
```
enthusiastmartin, line 599: i would make this little more generic. 

there is a need to create stableswap share asset too which requires similar change in registry. 

I can do it in separate PR - just this generic create asset trait which takes type as parameter. 

sounds good ?

Roznovjak: Good idea. I would also consider moving the asset type to `primitives` crate so we don't need to make other pallets dependent on asset registry pallet.

pallets/bonds/src/lib.rs
Roznovjak: removed

pallets/bonds/src/lib.rs
Roznovjak: fixed

integration-tests/src/bonds.rs
```diff
@@ -0,0 +1,124 @@
...
+		// Act
+		let bond_id = AssetRegistry::next_asset_id().unwrap();
+		assert_ok!(Bonds::issue(RuntimeOrigin::signed(ALICE.into()), HDX, amount, maturity));
+
+		// Assert
+		assert_eq!(Bonds::bonds(bond_id).unwrap(), (HDX, maturity));
+
+		let bond_asset_details = AssetRegistry::assets(bond_id).unwrap();
+
+		assert!(bond_asset_details.asset_type == pallet_asset_registry::AssetType::Bond);
```
dmoka, line 33: I would use for `assert_eq` for this to have a better failing message once it fails

integration-tests/src/bonds.rs
```diff
@@ -0,0 +1,124 @@
...
+			amount,
+			maturity
+		));
+
+		// Assert
+		assert_eq!(Bonds::bonds(bond_id).unwrap(), (shared_asset_id, maturity));
+
+		let bond_asset_details = AssetRegistry::assets(bond_id).unwrap();
+
+		assert!(bond_asset_details.asset_type == pallet_asset_registry::AssetType::Bond);
```
dmoka, line 82: I would use for `assert_eq` for this to have a better failing message once it fails

pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+
+			let underlying_asset_type = T::AssetRegistry::retrieve_asset_type(asset_id)?;
+			ensure!(
+				!T::AssetTypeBlacklist::contains(&underlying_asset_type),
+				Error::<T>::InvalidAssetType
+			);
+
+			BondIds::<T>::try_mutate_exists((asset_id, maturity), |maybe_bond| -> DispatchResult {
+				let fee = T::ProtocolFee::get().mul_ceil(amount); // TODO: check
+				let amount_without_fee = amount.checked_sub(&fee).ok_or(ArithmeticError::Overflow)?;
```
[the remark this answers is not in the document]
Roznovjak: I believe it's still overflow, just from the opposite side. Underflow happens with floating point numbers, when the result is smaller than the smallest representable value.

pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+
+						*bond_id
+					}
+					None => {
+						// register new bonds
+						ensure!(maturity >= T::TimestampProvider::now(), Error::<T>::InvalidMaturity);
+
+						let ed = T::ExistentialDeposits::get(&asset_id);
+
+						// not covered in the tests.
```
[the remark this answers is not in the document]
Roznovjak: The reason is that `T::AssetRegistry` is "external" to our crate, so we don't know (and don't care) how it is implemented. The implementation can also change in the future. The pallet should work even without registering assets in the asset registry. If we write tests for it, we would be testing just our mock of asset registry. But we have it covered in the integration tests.

pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+		///
+		#[pallet::call_index(1)]
+		#[pallet::weight(<T as Config>::WeightInfo::redeem())]
+		pub fn redeem(origin: OriginFor<T>, bond_id: T::AssetId, amount: T::Balance) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let (underlying_asset_id, maturity) = Self::bonds(bond_id).ok_or(Error::<T>::BondNotRegistered)?;
+
+			let now = T::TimestampProvider::now();
+			ensure!(now >= maturity, Error::<T>::BondNotMature);
```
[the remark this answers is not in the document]
Roznovjak: Thanks. I will add this test case.

mrq1911 (changes requested):

pallets/bonds/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+
+		/// The pallet id, used for deriving its sovereign account ID.
+		#[pallet::constant]
+		type PalletId: Get<PalletId>;
+
+		/// The origin which can issue new bonds.
+		type IssueOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
+
+		/// Asset types that are not permitted to be used as underlying assets.
+		type AssetTypeBlacklist: Contains<AssetKind>;
```
mrq1911, line 133: flipping this to allowlist would be probably safer when we lan to indroduce new asset types

runtime/hydradx/src/assets.rs
```diff
@@ -480,3 +480,32 @@ impl pallet_dynamic_fees::Config for Runtime {
 	type AssetFeeParameters = AssetFeeParams;
 	type ProtocolFeeParameters = ProtocolFeeParams;
 }
+
+// Bonds
+parameter_types! {
+	pub ProtocolFee: Permill = Permill::from_percent(2);
```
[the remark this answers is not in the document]
Roznovjak: Oh. I just needed to use some value. We need to discuss and determine what value to use.

mrq1911 (approved):

mrq1911 (approved):

PR #647, opened 2023-07-29
feat: enable polkadotXcm.send

green-jay:
Allow polkadotXcm.send to be called locally as an extrinsic and remotely within a XCM Transact call

### DIFF
--- a/runtime/hydradx/src/xcm.rs
+++ b/runtime/hydradx/src/xcm.rs
@@ -413,7 +413,6 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 				| RuntimeCall::Currencies(..)
 				| RuntimeCall::Tokens(..)
 				| RuntimeCall::OrmlXcm(..)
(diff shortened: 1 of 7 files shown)


runtime/hydradx/src/xcm.rs
```diff
@@ -413,6 +413,7 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 				| RuntimeCall::Currencies(..)
 				| RuntimeCall::Tokens(..)
 				| RuntimeCall::OrmlXcm(..)
+				| RuntimeCall::PolkadotXcm(pallet_xcm::Call::send { .. },)
```
mrq1911, line 416: @green-jay i don't think we can unfilter it here, is shouldn't be necessary anyway

cc @Roznovjak

mrq1911: btw. seems weird that OrmlXcm is unfiltered as well

mrq1911, line 416:
```suggestion
```

mrq1911 (approved):

PR #654, opened 2023-08-14
chore: staking period length for rococo

martinfridrich:
Staking pallet with period length of 1 block for rococo testing.


mrq1911:
no longer relevant

PR #655, opened 2023-08-15
feat: permissionless asset-registry 

martinfridrich:
# Permisionless asset-registry

Changes in asset-registry
* asset-registry refactor, storage reorganization and make properties optional
* added extrinsic to permissionlessly register that has only asset location
* created new traits for asset-registry. Refactor of other pallets to use new traits will done in separate PR
* implemented ED for insufficient assets 

## TODO:
* [x] asset-registry migration
* [x] Review and merge #693 first


runtime/hydradx/src/assets.rs
```diff
@@ -79,12 +85,128 @@ pub struct CurrencyHooks;
...
+			return Ok(());
+		}
+
+		//NOTE: account existence means it already paid ED
+		if orml_tokens::Accounts::<Runtime>::try_get(paying_account, asset).is_err()
+			&& !AssetRegistry::is_sufficient(asset)
+		{
+			let fee_payment_asset = MultiTransactionPayment::account_currency(paying_account);
+
+			//NOTE: unwrap should never fail. MutiTransctionPayment should always provide price.
```
jak-pan, line 116: This is not true. We won't have price for newly created assets

martinfridrich: These assets can't be used as fee asset, you can use only accepted currencies https://github.com/galacticcouncil/HydraDX-node/blob/2cc7a5fec807744d382689d5847172bac50008c5/pallets/transaction-multi-payment/src/lib.rs#L239-L242 for which you have to provide a price when you are adding it into accepted currencies.

Based on this comment we always provides some kind of fallback price:
https://github.com/galacticcouncil/HydraDX-node/blob/2cc7a5fec807744d382689d5847172bac50008c5/pallets/transaction-multi-payment/src/lib.rs#L186

But you are right, I didn't think about case when currency was removed from accepted currencies and user still have it set as fee asset, I'll make it fail with:
https://github.com/galacticcouncil/HydraDX-node/blob/2cc7a5fec807744d382689d5847172bac50008c5/pallets/transaction-multi-payment/src/lib.rs#L155

traits/src/registry.rs
```diff
@@ -76,3 +77,81 @@ pub trait AccountIdFor<Assets> {
 	/// Create a name to uniquely identify a share token for given assets and an identifier.
 	fn name(assets: &Assets, identifier: Option<&[u8]>) -> Vec<u8>;
 }
+
+use frame_support::dispatch::Parameter;
+
+pub trait Inspect<AssetNativeLocation, Balance> {
```
enthusiastmartin, line 83: Why this traits needs generics? it does not seem to use them.

Same for associated type Error.

also AssetId should be probably generic here.

traits/src/registry.rs
```diff
@@ -76,3 +77,81 @@ pub trait AccountIdFor<Assets> {
...
+		kind: AssetKind,
+		existential_deposit: Option<Balance>,
+		symbol: Option<&[u8]>,
+		decimals: Option<u8>,
+		location: Option<AssetNativeLocation>,
+		xcm_rate_limit: Option<Balance>,
+		is_sufficient: bool,
+	) -> Result<Self::AssetId, Self::Error>;
+
+	fn register_insufficient_asset(
```
enthusiastmartin, line 104: Why need of special insufficient and sufficient functions ? 

isnt the register_asset enough ?

pallets/asset-registry/src/lib.rs
```diff
@@ -265,181 +323,170 @@ pub mod pallet {
...
 		) -> DispatchResult {
-			T::RegistryOrigin::ensure_origin(origin)?;
+			let is_registry_origin = match T::RegistryOrigin::ensure_origin(origin.clone()) {
+				Ok(_) => true,
+				Err(_) => {
+					T::UpdateOrigin::ensure_origin(origin)?;
+
+					false
+				}
+			};
```
enthusiastmartin, line 396: just nitpick really.

i would probably go for something like this. It looks more readable

```rust
let is_registry_origin = T::RegistryOrigin::ensure_origin(origin.clone()).is_ok();
if !is_registry_origin {
	T::UpdateOrigin::ensure_origin(origin)?;
}
```

pallets/asset-registry/src/lib.rs
```diff
@@ -162,27 +184,34 @@ pub mod pallet {
 	pub type LocationAssets<T: Config> =
 		StorageMap<_, Blake2_128Concat, T::AssetNativeLocation, T::AssetId, OptionQuery>;
 
-	#[pallet::storage]
-	#[pallet::getter(fn asset_metadata)]
-	/// Metadata of an asset.
-	pub type AssetMetadataMap<T: Config> =
-		StorageMap<_, Twox64Concat, T::AssetId, AssetMetadata<BoundedVec<u8, T::StringLimit>>, OptionQuery>;
```
enthusiastmartin, line 169: as I cannot comment on non-changed code, I will put the comment here

as it is permissionless now - for the Assets storage, we need to change the Hasher type to Blake2 instead of Twox64.

Same for assetlocations. or basically everywhere. 

this might need to be considered in migration though.

enthusiastmartin (commented):
Initially findings, comments. 

There are stuff that it not really tested - as i try to change few things here and there and no tests failed. 

Also the registry traits need to be consolidated - i am aware that you mentioend that you keep the olds one for now.

pallets/asset-registry/src/lib.rs
```diff
@@ -74,28 +81,30 @@ pub mod pallet {
...
 
 		/// Native Asset Id
 		#[pallet::constant]
 		type NativeAssetId: Get<Self::AssetId>;
 
+		/// Storage fees for external asset creation.
+		#[pallet::constant]
+		type StorageFees: Get<Balance>;
+
+		/// Storage fees for external asset creation.
```
jak-pan, line 104:
```suggestion
		/// Beneficiary account of storage fees for external asset creation.
```

pallets/asset-registry/src/lib.rs
```diff
@@ -134,6 +143,19 @@ pub mod pallet {
...
+		/// Origin is fobiddent to set/update value
+		Forbidden,
+
+		/// Balance too low
+		InsufficientBalance,
+	}
+
+	#[pallet::type_value]
+	/// Default value of NextAssetId if storage is empty. 1 is used to offset for native token
+	/// which id is 0.
```
jak-pan, line 156:
```suggestion
	/// Default value of NextAssetId if storage is empty.
	/// 1 is used to offset the native asset with id 0.
```

pallets/asset-registry/src/lib.rs
```diff
@@ -265,181 +323,170 @@ pub mod pallet {
...
-				let bounded_name = Self::to_bounded_name(name)?;
+				let new_bounded_name = if let Some(n) = name {
+					let new_name = Self::to_bounded_name(n)?;
+					ensure!(Self::asset_ids(&new_name).is_none(), Error::<T>::AssetAlreadyRegistered);
+
+					if let Some(old_name) = &details.name {
+						AssetIds::<T>::remove(old_name);
+					}
+
+					if Some(new_name.clone()) != details.name {
```
jak-pan, line 409: As I read this, it will remove the name if you send none in the function but the docs for it says it doesn't get updated.

pallets/asset-registry/src/lib.rs
```diff
@@ -265,181 +323,170 @@ pub mod pallet {
...
+					false
+				}
+			};
 
 			Assets::<T>::try_mutate(asset_id, |maybe_detail| -> DispatchResult {
-				let mut detail = maybe_detail.as_mut().ok_or(Error::<T>::AssetNotFound)?;
+				let mut details = maybe_detail.as_mut().ok_or(Error::<T>::AssetNotFound)?;
 
-				let bounded_name = Self::to_bounded_name(name)?;
+				let new_bounded_name = if let Some(n) = name {
```
jak-pan, line 401: If name param is not provided is this executed?

runtime/hydradx/src/assets.rs
```diff
@@ -79,12 +85,128 @@ pub struct CurrencyHooks;
...
+pub const SUFFICIENCY_LOCK: LockIdentifier = *b"insuffEd";
+
+parameter_types! {
+	pub InsufficientEDinHDX: Balance = FixedU128::from_rational(11, 10)
+		.saturating_mul_int(<Runtime as pallet_balances::Config>::ExistentialDeposit::get());
+}
+
+pub struct SufficiencyCheck;
+impl SufficiencyCheck {
+	fn on_funds(asset: AssetId, paying_account: &AccountId) -> DispatchResult {
```
jak-pan, line 105: Docs pls would like to see how and why this works

runtime/hydradx/src/assets.rs
```diff
@@ -79,12 +85,128 @@ pub struct CurrencyHooks;
...
+			//NOTE: Account doesn't have enough funds to pay ED if this fail.
+			<Currencies as MultiCurrency<AccountId>>::transfer(
+				fee_payment_asset,
+				paying_account,
+				&TreasuryAccount::get(),
+				ed,
+			)
+			.map_err(|_| orml_tokens::Error::<Runtime>::KeepAlive)?;
+
+			let to_lock = pallet_balances::Locks::<Runtime>::get(TreasuryAccount::get())
```
jak-pan, line 130: would like to have this explained in the docs also

runtime/hydradx/src/assets.rs
```diff
@@ -162,17 +284,21 @@ impl pallet_claims::Config for Runtime {
 parameter_types! {
 	pub const RegistryStrLimit: u32 = 32;
 	pub const SequentialIdOffset: u32 = 1_000_000;
+	pub const StoreFees: Balance = 100 * UNITS; //TODO:
```
jak-pan, line 287: todo

traits/src/registry.rs
martinfridrich: `register_asset` is enough but sufficiency is important (dest. account has to pay ED for insufficient assets on deposit/transfer).
I think it help with readability and can ease debugging/review when 3-rd person look at the code

pallets/asset-registry/src/lib.rs
martinfridrich: no, if name is `None` it jumps into the else block

pallets/asset-registry/src/lib.rs
martinfridrich: I don't think so, this whole block is executed only if name is not `None`. 
There is this test:
https://github.com/galacticcouncil/HydraDX-node/blob/1156f597f469731e9b9e64a6dc0dd2978499b4a0/pallets/asset-registry/src/tests/update.rs#L170C1-L224C2

I can write another one if you have specific scenario in mind.

traits/src/registry.rs
martinfridrich: I probably added these generics prematurely but I think they will be necessary after we will remove old traits. I will remove it for now.

I wanted to use this one as supretrait for other registry traits that's why I added these associated types.

Why do you think AssetId` should be generic?

pallets/asset-registry/src/lib.rs
```diff
@@ -615,27 +668,90 @@ impl<T: Config> GetByKey<T::AssetId, T::Balance> for Pallet<T> {
...
+		} else {
+			None
+		};
+
+		let bounded_symbol = if let Some(symbol) = symbol {
+			Some(Self::to_bounded_name(symbol.to_vec())?)
+		} else {
+			None
+		};
+
```
apopiak, line 733: might be worth extracting the common logic of these two into a function?
so you get `let bounded_symbol = try_into_bounded(symbol)?;`

pallets/asset-registry/src/benchmarking.rs
```diff
@@ -19,115 +19,90 @@
...
-		};
+	 where_clause { where
+		T::Currency: MultiCurrencyExtended<T::AccountId, Amount=i128>,
+		T: crate::pallet::Config,
+	}
 
-	}: _(RawOrigin::Root, name.clone(), AssetType::Token, ed, None, Some(metadata), Some(Default::default()), None)
+	register {
+		let asset_id= T::AssetId::from(3);
+		let name = b"Test name".to_vec();
```
apopiak, line 37: should be length of max bounded, no?

apopiak (commented):
some review comments, will continue later today

pallets/asset-registry/src/lib.rs
martinfridrich: > I think the name is removed if it is the same as the previous one

You can't name to previous one. It will fail on this check
https://github.com/galacticcouncil/HydraDX-node/blob/cdfc19aca97d1e0c011d28f50646ee1c07e78c3b/pallets/asset-registry/src/lib.rs#L402

pallets/asset-registry/src/lib.rs
```diff
@@ -134,16 +144,29 @@ pub mod pallet {
...
+		/// Balance too low
+		InsufficientBalance,
+	}
+
+	#[pallet::type_value]
+	/// Default value of NextAssetId if storage is empty. 1 is used to offset for native token
+	/// which id is 0.
+	pub fn DefaultNextAssetId<T: Config>() -> T::AssetId {
+		1.into()
```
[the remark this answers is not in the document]
martinfridrich: yes, but it has to be `0`.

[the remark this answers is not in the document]
apopiak: I don't get it, either you make the assumption for the pallet that the native asset id is zero, or you make it configurable

math/proptest-regressions/stableswap/tests/invariants.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc dfba9d405ba61bbd3bc1917de4383fc12c9b7ad4707420aa05918f7bd2952d8c # shrinks to pool = [AssetReserve { amount: 10000000000, decimals: 6 }, AssetReserve { amount: 160124693000000, decimals: 6 }], amount = 1257, amp = 2
```
enthusiastmartin, line 7: why ?

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+		}
+
+		let locked_ed = pallet_balances::Locks::<Runtime>::get(TreasuryAccount::get())
+			.iter()
+			.find(|x| x.id == SUFFICIENCY_LOCK)
+			.map(|p| p.amount)
+			.unwrap_or_default();
+
+		let paid_accounts = pallet_asset_registry::ExistentialDepositCounter::<Runtime>::get();
+		let ed_to_pay = if paid_accounts != 0 {
```
enthusiastmartin, line 224: refund would be more explanatory, i guess

```suggestion
		let ed_to_refund = if paid_accounts != 0 {
```

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+				.saturating_mul_int(InsufficientEDinHDX::get());
+
+			//NOTE: Account doesn't have enough funds to pay ED if this fail.
+			<Currencies as MultiCurrency<AccountId>>::transfer(
+				fee_payment_asset,
+				paying_account,
+				&TreasuryAccount::get(),
+				ed_in_fee_asset,
+			)
+			.map_err(|_| orml_tokens::Error::<Runtime>::KeepAlive)?;
```
enthusiastmartin, line 154: why you need to map error to this specific ? 

the transfer would fail with ExistentialDeposit or BalanceTooLow.  Keepalive might be misleading if there is not enough to pay.

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+		if AssetRegistry::is_sufficient(*asset) {
+			return;
+		}
+
+		let locked_ed = pallet_balances::Locks::<Runtime>::get(TreasuryAccount::get())
+			.iter()
+			.find(|x| x.id == SUFFICIENCY_LOCK)
+			.map(|p| p.amount)
+			.unwrap_or_default();
+
```
enthusiastmartin, line 222: you can short-circuit here. if amount is 0, you can jump out, no need to carry on.

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+	/// `set_balance` - bypass `MutationHooks` so no one pays ED for these account but ED is still released
+	/// when account is killed.
+	///
+	/// Emits `pallet_asset_registry::Event::ExistentialDepositPaid` when ED was paid.
+	fn on_funds(asset: AssetId, paying_account: &AccountId, to: &AccountId) -> DispatchResult {
+		if orml_tokens::Accounts::<Runtime>::try_get(to, asset).is_err() && !AssetRegistry::is_sufficient(asset) {
+			let fee_payment_asset = MultiTransactionPayment::account_currency(paying_account);
+
+			let ed_in_fee_asset = MultiTransactionPayment::price(fee_payment_asset)
+				.ok_or(pallet_transaction_multi_payment::Error::<Runtime>::UnsupportedCurrency)?
```
enthusiastmartin, line 144: this is incorrect imo.

if you can fee payment asset, it should be already supported. something like UnknownPrice would be better ( if there such error )

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+				.saturating_mul_int(InsufficientEDinHDX::get());
+
```
enthusiastmartin, line 146: i would probably ensure here that calculated amount is > 0. just in case

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+	/// ED is always released on account kill to killed account, whitelisting doesn't matter.
+	/// Released ED amount is calculated from locked HDX divided by number of accounts that paid
+	/// ED.
+	///
+	/// WARN:
+	/// `set_balance` - bypass `MutationHooks` so no one pays ED for these account but ED is still released
+	/// when account is killed.
+	///
+	/// Emits `pallet_asset_registry::Event::ExistentialDepositPaid` when ED was paid.
+	fn on_funds(asset: AssetId, paying_account: &AccountId, to: &AccountId) -> DispatchResult {
```
enthusiastmartin, line 139: is this executed on each transfer ? how do we weight this ?

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+		if orml_tokens::Accounts::<Runtime>::try_get(to, asset).is_err() && !AssetRegistry::is_sufficient(asset) {
```
enthusiastmartin, line 140: just curious. why is_err?   I cant really understand, so perhaps a comment why in case of error would be helpul

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,176 @@ pub struct CurrencyHooks;
...
+			<Currencies as MultiLockableCurrency<AccountId>>::set_lock(
+				SUFFICIENCY_LOCK,
+				NativeAssetId::get(),
+				&TreasuryAccount::get(),
+				to_lock,
+			)?;
+
+			frame_system::Pallet::<Runtime>::inc_sufficients(paying_account);
+
+			let _ = pallet_asset_registry::ExistentialDepositCounter::<Runtime>::try_mutate(
```
enthusiastmartin, line 172: it seems, that you dont care about an error here. so simple mutate would work here.  no need for try.

runtime/hydradx/src/assets.rs
```diff
@@ -165,19 +334,24 @@ impl pallet_claims::Config for Runtime {
 }
 
 parameter_types! {
+	#[derive(PartialEq, Debug)]
```
enthusiastmartin, line 337: why this is needed here ? u32 should already support these traits

pallets/asset-registry/src/lib.rs
```diff
@@ -265,181 +346,148 @@ pub mod pallet {
...
-		/// Emits `LocationSet` event when successful.
-		#[pallet::call_index(3)]
-		#[pallet::weight(<T as Config>::WeightInfo::set_location())]
-		pub fn set_location(
-			origin: OriginFor<T>,
-			asset_id: T::AssetId,
-			location: T::AssetNativeLocation,
-		) -> DispatchResult {
-			T::RegistryOrigin::ensure_origin(origin)?;
+		//NOTE: call indices 2 and 3 were used by removed extrinsics.
```
enthusiastmartin, line 466: i wonder why we cannot reuse those ?

runtime/hydradx/src/assets.rs
martinfridrich: it's hook from orml_tokens so I assume it's included in its weights

runtime/hydradx/src/assets.rs
martinfridrich: it's how `try_get()` works, it returns Ok if exists or Err if not, I can add comment

runtime/hydradx/src/assets.rs
martinfridrich: This can happen only if currency was supported at some point, and user set it, but it's no longer anymore. 
There is no `UnknowPrice` error in multi payment pallet.
We can use `FallbackPriceNotFound` but this is in my opinion confusing because you have no idea why fallback  price was not found.

runtime/hydradx/src/assets.rs
martinfridrich: I think it would fail with `BalanceToLow` and this may be confusing because you can have enough balance to transfer but not enough to pay ED.

I'll use `ExistentialDeposit` error, I somehow missed this one when I was looking at orml-tokens errors.

runtime/hydradx/src/assets.rs
martinfridrich: We can't. We still must decrement counters (sufficients and ExistentilaDepositCounter)

runtime/hydradx/src/assets.rs
martinfridrich: Rust started to complain and I had to add these when I removed `BoundedVec` from `AssetDetails`

math/proptest-regressions/stableswap/tests/invariants.txt
martinfridrich: I missed one, I'll remove it.

pallets/asset-registry/src/lib.rs
martinfridrich: I'm not sure, @apopiak pointed out it's better to retire old and introduce new indices

runtime/hydradx/src/assets.rs
enthusiastmartin: but this i know, i wonder why only in case of error.

martinfridrich: We want to collect ED only if it wasn't paid yet so I assume account already paid ED if storage record exits.
I'll added a comment.

pallets/asset-registry/src/lib.rs
apopiak: It's up to you, but you need to bump the transaction version if you change the parameters of an extrinsic.

integration-tests/src/asset_registry.rs
```diff
@@ -0,0 +1,169 @@
...
+					X2(Parachain(MOONBEAM_PARA_ID), GeneralIndex(0))
+				))),
+			),
+			pallet_asset_registry::Error::<hydradx_runtime::Runtime>::Forbidden
+		);
+	});
+}
+
+#[test]
+fn root_should_update_location_when_asset_exists() {
```
jak-pan, line 124: techcomm test?

runtime/hydradx/src/assets.rs
enthusiastmartin: isnt this caused by the fact that in the config for the StringLimit you explicity added these two traits ? which i dont think it should be needed.

pallets/asset-registry/src/lib.rs
```diff
@@ -74,27 +84,30 @@ pub mod pallet {
...
-			+ MaybeSerializeDeserialize
-			+ MaxEncodedLen;
-
 		/// Asset location type
 		type AssetNativeLocation: Parameter + Member + Default + MaxEncodedLen;
 
-		/// The maximum length of a name or symbol stored on-chain.
-		type StringLimit: Get<u32>;
+		/// Multi currency mechanism
+		type Currency: MultiCurrency<Self::AccountId, CurrencyId = Self::AssetId, Balance = Balance>;
```
enthusiastmartin, line 91: i wonder if we should just try to use fungibles traits from now one where multi currecny support is required,.

this way, we just rely on substrate instead of orml.

pallets/asset-registry/src/lib.rs
```diff
@@ -163,83 +194,134 @@ pub mod pallet {
...
-			Assets::<T>::insert(T::NativeAssetId::get(), details);
-
-			self.registered_assets.iter().for_each(|(name, ed, id)| {
-				let bounded_name = Pallet::<T>::to_bounded_name(name.to_vec())
-					.map_err(|_| panic!("Invalid asset name!"))
-					.unwrap();
-				let _ = Pallet::<T>::register_asset(bounded_name, AssetType::Token, *ed, *id, None)
-					.map_err(|_| panic!("Failed to register asset"));
-			})
+			let _ = with_transaction(|| {
```
enthusiastmartin, line 235: i wonder why you ignore the error.

pallets/asset-registry/src/lib.rs
```diff
@@ -452,45 +504,45 @@ impl<T: Config> Pallet<T> {
...
-		name.try_into().map_err(|_| Error::<T>::TooLong)
+	fn try_into_bounded(name: Option<Vec<u8>>) -> Result<Option<BoundedVec<u8, T::StringLimit>>, Error<T>> {
+		if let Some(name) = name {
+			TryInto::<BoundedVec<u8, T::StringLimit>>::try_into(name)
+				.map_err(|_| Error::<T>::TooLong)
+				.map(Some)
+		} else {
+			Ok(None)
+		}
 	}
```
enthusiastmartin, line 515: just a nitpick really, but i find this more readable.

```suggestion
	fn try_into_bounded(name: Option<Vec<u8>>) -> Result<Option<BoundedVec<u8, T::StringLimit>>, Error<T>> {
		let Some(s) = name else {
		    return Ok(None);
		}
		s.try_into().map_err(|_| Error::<T>::TooLong).map(Some)
		
	}
```

integration-tests/src/asset_registry.rs
martinfridrich: I don't understand your comment

pallets/asset-registry/src/lib.rs
martinfridrich: I don't think this can fail. If something inside of this  block fail, it will panic on `expect` or `unwrap`.

pallets/xyk/src/benchmarking.rs
```diff
@@ -32,6 +32,9 @@ const SEED: u32 = 1;
 
 fn funded_account<T: Config>(name: &'static str, index: u32) -> T::AccountId {
 	let caller: T::AccountId = account(name, index, SEED);
+	//Necessary for ED for insufficien assets.
```
jak-pan, line 35:
```suggestion
	// Necessary for ED for insufficient assets.
```

pallets/asset-registry/src/lib.rs
```diff
@@ -528,15 +581,29 @@ impl<T: Config> Pallet<T> {
...
-		existential_deposit: T::Balance,
+		existential_deposit: Balance,
 		asset_id: Option<T::AssetId>,
+		is_sufficient: bool,
 	) -> Result<T::AssetId, DispatchError> {
-		let bounded_name: BoundedVec<u8, T::StringLimit> = Self::to_bounded_name(name)?;
+		let bounded_name = Self::try_into_bounded(Some(name))?;
 
-		if let Some(asset_id) = AssetIds::<T>::get(&bounded_name) {
+		//NOTE: this unwrap is safe.
```
jak-pan, line 590: Pls explain why in comment if you do smth like this

pallets/asset-registry/src/lib.rs
```diff
@@ -551,13 +618,14 @@ impl<T: Config> Pallet<T> {
...
 
-impl<T: Config> Registry<T::AssetId, Vec<u8>, T::Balance, DispatchError> for Pallet<T> {
+impl<T: Config> Registry<T::AssetId, Vec<u8>, Balance, DispatchError> for Pallet<T> {
 	fn exists(asset_id: T::AssetId) -> bool {
 		Assets::<T>::contains_key(asset_id)
 	}
 
 	fn retrieve_asset(name: &Vec<u8>) -> Result<T::AssetId, DispatchError> {
-		let bounded_name = Self::to_bounded_name(name.clone())?;
+		//NOTE: This unwrap is safe.
```
jak-pan, line 627: Same as b4 pls explain

integration-tests/src/insufficient_assets_ed.rs
```diff
@@ -0,0 +1,1390 @@
...
+		X2(Parachain(MOONBEAM_PARA_ID), GeneralIndex(general_index)),
+	));
+
+	let next_asset_id = Registry::next_asset_id().unwrap();
+	Registry::register_external(hydra_origin::signed(BOB.into()), location).unwrap();
+
+	next_asset_id
+}
+
+fn treasury_suffyciency_lock() -> Balance {
```
apopiak, line 1350:
```suggestion
fn treasury_sufficiency_lock() -> Balance {
```

pallets/asset-registry/src/lib.rs
apopiak: should be an `expect`, no?

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +88,179 @@ pub struct CurrencyHooks;
...
+
+parameter_types! {
+	pub InsufficientEDinHDX: Balance = FixedU128::from_rational(11, 10)
+		.saturating_mul_int(<Runtime as pallet_balances::Config>::ExistentialDeposit::get());
+}
+
+pub struct SufficiencyCheck;
+impl SufficiencyCheck {
+	/// This function is used by `orml-toknes` `MutationHooks` before a transaction is executed.
+	/// It is called from `PreDeposit` and `PreTransfer`.
```
apopiak, line 109:
```suggestion
	/// This function is used by `orml-tokens::MutationHooks` before a transaction is executed.
	/// It is called from `PreDeposit` and `PreTransfer`.
```

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +88,179 @@ pub struct CurrencyHooks;
...
+
+			let ed_in_fee_asset = MultiTransactionPayment::price(fee_payment_asset)
+				.ok_or(pallet_transaction_multi_payment::Error::<Runtime>::UnsupportedCurrency)?
+				.saturating_mul_int(InsufficientEDinHDX::get());
+
+			//NOTE: Not tested, this should never happen.
+			ensure!(
+				!ed_in_fee_asset.is_zero(),
+				pallet_asset_registry::Error::<Runtime>::ZeroExistentialDeposit
+			);
```
apopiak, line 152: why not add `.max(1)` to the ED instead?

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +88,179 @@ pub struct CurrencyHooks;
...
+				NativeAssetId::get(),
+				&TreasuryAccount::get(),
+			);
+		} else {
+			let _ = <Currencies as MultiLockableCurrency<AccountId>>::set_lock(
+				SUFFICIENCY_LOCK,
+				NativeAssetId::get(),
+				&TreasuryAccount::get(),
+				to_lock,
+			);
```
apopiak, line 245: I assume this returns a result? Would be good to use a defensive unwrap

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +88,179 @@ pub struct CurrencyHooks;
...
+			who,
+			ed_to_refund,
+		);
+
+		//NOTE: This is necessary because grandfathered accounts doesn't have incremented
+		//sufficients by `SufficiencyCheck` so without check it can overflow.
+		//`set_balance` also bypass `MutationHooks`
+		if frame_system::Pallet::<Runtime>::account(who).sufficients > 0 {
+			frame_system::Pallet::<Runtime>::dec_sufficients(who);
+		}
```
apopiak, line 260: This means we cannot introduce a different kind of sufficient in the runtime. So e.g. using pallet-assets

apopiak: Just now realizing: if we rely on the `sufficient` count not being touched elsewhere anyway we could also distinguish between legacy and new insufficients by way of the sufficient count and early return further up if the sufficient count is 0.

pallets/asset-registry/src/migration.rs
```diff
@@ -7,81 +7,170 @@
...
-					asset_type,
-					existential_deposit,
-					xcm_rate_limit: None,
-				})
-			},
+		let mut v2_assets_details = Vec::<(
+			<T as crate::Config>::AssetId,
+			AssetDetails<<T as crate::Config>::AssetId, <T as crate::Config>::StringLimit>,
+		)>::new();
+		for (k, v) in v1::Assets::<T>::iter() {
```
apopiak, line 106: you don't remove `v1::Assets`?

pallets/asset-registry/src/migration.rs
```diff
@@ -7,81 +7,170 @@
...
-		assert_eq!(StorageVersion::get::<Pallet<T>>(), 1, "Unexpected storage version.");
+		for a in Assets::<T>::iter_keys() {
+			let _ = Assets::<T>::get(a).expect("Assets data must be valid");
+		}
+
+		for l in AssetLocations::<T>::iter_keys() {
+			let _ = AssetLocations::<T>::get(l).expect("AssetLocations data must be valid");
+		}
+
+		assert_eq!(v1::AssetMetadataMap::<T>::iter().count(), 0);
```
apopiak, line 171:
```suggestion
		assert_eq!(v1::AssetMetadataMap::<T>::iter().count(), 0);
		assert_eq!(v1::Assets::<T>::iter().count(), 0);
```

pallets/asset-registry/src/migration.rs
```diff
@@ -7,81 +7,170 @@
...
+			AssetLocations::<T>::migrate_key::<Twox64Concat, <T as crate::Config>::AssetId>(k);
 
+			log::info!(
+				target: "runtime::asset-registry",
+				"Migrated asset's location: {:?}", k
+			);
+		}
+
+		StorageVersion::new(2).put::<Pallet<T>>();
 		T::DbWeight::get().reads_writes(i, i)
```
apopiak, line 159: I'm unsure whether you under- or overcount honestly.
Pretty sure you overcount reads?

integration-tests/src/insufficient_assets_ed.rs
```diff
@@ -0,0 +1,1390 @@
...
+			alice_balance + InsufficientEDinHDX::get()
+		);
+		assert_eq!(Currencies::free_balance(sht1, &ALICE.into()), 0);
+
+		assert_eq!(
+			Currencies::free_balance(HDX, &TreasuryAccount::get()),
+			treasury_balance - InsufficientEDinHDX::get()
+		);
+
+		assert_eq!(treasury_suffyciency_lock(), 0);
```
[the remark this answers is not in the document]
martinfridrich: no, `set_balance()` bypass `SufficiencyCheck`(it doesn't trigger `MutationHooks`). 
`SuffciencyCheck` assumes that the user already paid ED if he holds a transferring  token.
 That's why Alice's ED was unlocked and Bob's wasn't locked.

integration-tests/src/insufficient_assets_ed.rs
```diff
@@ -0,0 +1,1390 @@
...
+			}),
+			1
+		);
+	});
+}
+
+#[test]
+fn grandfathered_account_should_receive_hdx_when_account_is_killed() {
+	//NOTE: thiscase simulates old account that received insufficient asset before sufficiency
+	//check and didn't pay ED.
```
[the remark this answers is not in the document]
martinfridrich: `GRANDFATHERED_UNPAID_ED` received insufficient asset during chain state initialization so he didn't paid ED. 
I added comment to the code.

runtime/hydradx/src/assets.rs
martinfridrich: Would be `1` enough? Isn't better to fail?

pallets/asset-registry/src/migration.rs
martinfridrich: Based on the doc iter+remove is [undefined result](https://docs.rs/frame-support/latest/frame_support/storage/trait.IterableStorageDoubleMap.html#tymethod.iter) so I didn't want not risk it. 
All the elements are overwritten here:
https://github.com/galacticcouncil/HydraDX-node/blob/f0e4412cc6358c60247ba40f08e83a8b96f13aa6/pallets/asset-registry/src/migration.rs#L129C3-L136C4

pallets/asset-registry/src/migration.rs
martinfridrich: `v1::Assets` and latest `Assets` is the same storage so `v1::Assets` won't be empty

pallets/asset-registry/src/migration.rs
martinfridrich: I also think I overcounted reads, I wanted to be safe

pallets/asset-registry/src/migration.rs
apopiak: yeah, that's where you should remove.
At the point where you insert into v2 the iteration of v1 is done

pallets/asset-registry/src/migration.rs
apopiak: the hasher changes

runtime/hydradx/src/assets.rs
apopiak: well if it is one you don't fail, so 1 should be enough, yes :-D

pallets/asset-registry/src/migration.rs
apopiak: aah, the iterator cannot distinguish between different hashers, ok

pallets/asset-registry/src/migration.rs
apopiak: I would try to count accurately, but if you're confident enough that you overcounted, then sure, your call

runtime/hydradx/src/assets.rs
```diff
@@ -84,12 +89,173 @@ pub struct CurrencyHooks;
...
+			0
+		};
+		let to_lock = locked_ed.saturating_sub(ed_to_refund);
+
+		if to_lock.is_zero() {
+			let _ = <Currencies as MultiLockableCurrency<AccountId>>::remove_lock(
+				SUFFICIENCY_LOCK,
+				NativeAssetId::get(),
+				&TreasuryAccount::get(),
+			);
```
apopiak, line 238: should probably also be defensive?

apopiak (approved):
LGTM

runtime/hydradx/src/assets.rs
martinfridrich: yes, I added these because without it I wasn't able to assert in tests and benchmarks. If I remove it, I still need add these derives because of benchmakrs

pallets/asset-registry/src/lib.rs
```diff
@@ -529,16 +663,33 @@ impl<T: Config> Pallet<T> {
...
 		name: Vec<u8>,
-		asset_type: AssetType<T::AssetId>,
-		existential_deposit: T::Balance,
+		asset_type: AssetType,
+		existential_deposit: Balance,
 		asset_id: Option<T::AssetId>,
+		is_sufficient: bool,
 	) -> Result<T::AssetId, DispatchError> {
-		let bounded_name: BoundedVec<u8, T::StringLimit> = Self::to_bounded_name(name)?;
+		let bounded_name = Self::try_into_bounded(Some(name))?;
```
enthusiastmartin, line 671: if you use the type boundedvec instead of Vec<u8>, you dont have problems and no need for the inconsistentstate error at all.

enthusiastmartin (commented):
update readme too.

pallets/asset-registry/src/lib.rs
```diff
@@ -617,37 +722,121 @@ impl<T: Config> GetByKey<T::AssetId, T::Balance> for Pallet<T> {
...
+			is_sufficient,
+		);
+
+		Self::do_register_asset(asset_id, &details, location)
 	}
 
-	fn asset_symbol(asset_id: T::AssetId) -> Option<Vec<u8>> {
-		let asset_metadata = AssetMetadataMap::<T>::get(asset_id)?;
-		Some(asset_metadata.symbol.into_inner())
+	fn get_or_register_asset(
```
enthusiastmartin, line 807: hm..there is another function get_or_create_asset. It does basically the same. 

i wonder if it cannot be used - one way or another,.

pallets/asset-registry/src/lib.rs
```diff
@@ -267,182 +391,193 @@ pub mod pallet {
...
 			location: Option<T::AssetNativeLocation>,
-			xcm_rate_limit: Option<T::Balance>,
+			xcm_rate_limit: Option<Balance>,
+			is_sufficient: bool,
 		) -> DispatchResult {
 			T::RegistryOrigin::ensure_origin(origin)?;
 
-			let bounded_name = Self::to_bounded_name(name)?;
+			let bounded_name = Self::try_into_bounded(name)?;
+			let bounded_symbol = Self::try_into_bounded(symbol)?;
```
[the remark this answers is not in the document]
martinfridrich: I added a check to disallow white space characters in symbol and min length for both.
Names are using white spaces, non-alphanumeric and upper/lowercase mix is used by both name and symbol.

pallets/asset-registry/src/lib.rs
martinfridrich: I missed this one, removed

pallets/asset-registry/src/lib.rs
```diff
@@ -617,37 +722,121 @@ impl<T: Config> GetByKey<T::AssetId, T::Balance> for Pallet<T> {
...
+	fn exists(id: Self::AssetId) -> bool {
+		Assets::<T>::try_get(id).is_ok()
+	}
+
+	fn decimals(id: Self::AssetId) -> Option<u8> {
+		Self::assets(id).and_then(|a| a.decimals)
+	}
+
+	fn asset_type(id: Self::AssetId) -> Option<AssetKind> {
+		Self::assets(id).map(|a| a.asset_type.into())
```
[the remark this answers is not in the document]
martinfridrich: clippy doesn't like `and_then` in this case, it recommends to use `map()`.

enthusiastmartin (approved):

jak-pan (approved):

jak-pan (commented):
What do we miss still? Looks good.

martinfridrich:
> What do we miss still? Looks good.

I'm working on removing storage fess for external asset registration

PR #656, opened 2023-08-15
fix: asset names for bonds

Roznovjak:


pallets/bonds/src/lib.rs
```diff
@@ -86,16 +84,6 @@ pub mod pallet {
...
-		/// Identifier for the class of asset.
-		type AssetId: Member
-			+ Parameter
-			+ Default
-			+ Copy
-			+ HasCompact
-			+ MaybeSerializeDeserialize
-			+ MaxEncodedLen
-			+ TypeInfo;
-
```
enthusiastmartin, line 98: i guess you removed it because you could not do .to_le_bytes() on generic type

Have you considered making the "name" construction generic ? same as we do in xyk,stableswap ? 

at this point - you can remove Balance too. There is also quite forcing constraints such as  From<u128>, CheckedAdd/Sub, so no many options you have anyway.

enthusiastmartin (commented):
btw. there is a type in error variant : DisallowedAseet,

pallets/bonds/src/lib.rs
Roznovjak: Yes, that's the reason.
We don't use generic approach in xyk, and I couldn't find anything related in stableswap. Could you point me to the relevant code line? Thanks

enthusiastmartin: in xyk, we leave it up to the integration to provide a way how to contruct share asset name, here:

https://github.com/galacticcouncil/Basilisk-node/blob/91ee6ea8670877c1d189faa409205f31fdee1989/pallets/xyk/src/lib.rs#L83C3-L85C1

enthusiastmartin: using that approacah, you could have assetid type generic.

Roznovjak: `AssetPairAccountId` is used to generate an account. Something we don't use in the bonds pallet. You can find [here](https://github.com/galacticcouncil/Basilisk-node/blob/91ee6ea8670877c1d189faa409205f31fdee1989/pallets/xyk/src/lib.rs#L342C4-L342C4) what I was referring to.

enthusiastmartin: yes i know. but you can have similar way to construct the name.

it is only if you wanted to have the asset type generic. if not, no reason.

mrq1911 (approved):

PR #657, opened 2023-08-17
feat: add LBP pallet

Roznovjak:
Move LBP pallet from Basilisk repository.
As part of this PR, we also add some constants and `AssetPair` struct to `primitives` crate in order to avoid any changes in the pallet.


primitives/src/constants.rs
```diff
@@ -85,6 +85,18 @@ pub mod chain {
 	/// Core asset id
 	pub const CORE_ASSET_ID: AssetId = 0;
 
+	/// Max fraction of pool to buy in single transaction
+	pub const MAX_OUT_RATIO: u128 = 3;
```
mrq1911, line 89: these should be preferably constants, but if the XYK use same constants, i would already move the XYK as well

mrq1911 (changes requested):
LBP should be integrated into the runtime (incl. router) as well

mrq1911 (approved):

PR #668, opened 2023-09-11
feat: router benchmarks

Roznovjak:


runtime/hydradx/src/benchmarking/omnipool.rs
```diff
@@ -433,6 +436,128 @@ runtime_benchmarks! {
 		assert!(asset_state.cap == 100_000_000_000_000_000u128);
 	}
 
+	trade_execution_sell {
```
dmoka, line 439: What is the reason that these two tests are not placed in the omnipool pallet?

runtime/hydradx/src/assets.rs
```diff
@@ -429,6 +431,54 @@ impl pallet_dca::Config for Runtime {
...
+	pub fn buy_overhead_weight() -> Weight {
+		weights::route_executor::HydraWeight::<Runtime>::buy_in_lbp()
+			.saturating_sub(weights::lbp::HydraWeight::<Runtime>::trade_execution_buy())
+	}
+}
+impl AmmTradeWeights<AssetId> for AmmWeights {
+	fn sell_weight(route: &[Trade<AssetId>]) -> Weight {
+		let mut weight = Weight::zero();
+
+		for trade in route {
```
dmoka, line 450: awesome, like it! :muscle:

runtime/hydradx/src/benchmarking/route_executor.rs
```diff
@@ -16,232 +16,159 @@
...
 
-//TODO: Rebenchmark both buy and sell with dynamic length of route once we have other AMMs in hydra
-
 runtime_benchmarks! {
-	{ Runtime, pallet_route_executor}
+	{Runtime, pallet_route_executor}
 
-	sell {
-		let n in 1..2;
+	sell_in_lbp {
```
dmoka, line 99: I guess the only reason for this lbp-specific stuff is to calculate the overhead right?

Would not be better to use some mocked AMMs? The thing is that this sell_in_lbp has just one route included, and maybe would be better to include more amms to properly benchmark the overhead of looping over AMMs and its logic. What do you think?

runtime/hydradx/src/assets.rs
```diff
@@ -429,6 +431,54 @@ impl pallet_dca::Config for Runtime {
...
+			let amm_weight = match trade.pool {
+				PoolType::Omnipool => weights::omnipool::HydraWeight::<Runtime>::trade_execution_sell(),
+				PoolType::LBP => weights::lbp::HydraWeight::<Runtime>::trade_execution_sell(),
+				PoolType::Stableswap(_) => Weight::zero(),
```
dmoka, line 454: instead of zero, how about putting some default weight of Omnipool or LBP, as a safeguard? 

Let's say that someone adds router support for XYK or stableswap, but forgets to change this, so the weight will remain zero. Although it has a small chance, we never know, better safe than sorry.

Also, what I have done on my branch is to add router integration tests returning "NotSupported" error for the AMMs not supported yet, and also also placing comments there to be sure that these weights are updated by the one who adds support for a new AMM.

runtime/hydradx/src/benchmarking/omnipool.rs
Roznovjak: These are benchmarks. The weights are used in the `AmmWeights` struct.

runtime/hydradx/src/assets.rs
Roznovjak: replaced by the omnipool weights.

runtime/hydradx/src/benchmarking/omnipool.rs
dmoka: Yes, but the similar `trade_execution_sell` and `trade_execution_buy` of LBP are placed to the LBP pallet `benchmark.rs` file. 

So why not putting these then into the omnipool pallet?

runtime/hydradx/src/benchmarking/route_executor.rs
dmoka: I see! I would not call it totally independent though, for example we have [this balance check](https://github.com/galacticcouncil/HydraDX-node/blob/e2702e877f9ace7bc0bc0d56356920b08d077269/pallets/route-executor/src/lib.rs#L201) after each trade in the route.

But yeah, this is probably a small weight, and if it would really complicates our life, the current solution is perfectly fine if we overestimate a bit.

runtime/hydradx/src/benchmarking/omnipool.rs
Roznovjak: We don't have benchmarks.rs in the omnipool pallet.

runtime/hydradx/src/benchmarking/route_executor.rs
Roznovjak: It's independent because there is no code path that depends on a specific AMM. `AMM::calculate_trade` and `AMM::execute_trade` is called for all AMMs. `ensure_that_user_spent_asset_in` is also called every time. The code that belongs to "overhead" doesn't care what AMM we use. It runs the same code for each AMM.

dmoka: Yeah, what I meant is that the loop depends on the number of AMM trades in the loop. Right now in the benchmark the `ensure_that_user_spent_asset_in` is called only once, as the route has one trade. 

But for example for a route with 5 trade, it would be called 5 times. We don't include this in our benchmark.

Roznovjak: Hmm. You are right. I didn't realize that. But we add the overhead weight for every trade, so the resulting weight is overestimated. I think we can update the benchmarks to calculate them with MaxRouteSize. But it would provide the overhead that is most accurate for trades with MaxRouteSize. The current implementation provides better weights for trades containing just one trade.

runtime/hydradx/src/benchmarking/omnipool.rs
mrq1911: dependencies reasons probably

runtime/hydradx/src/assets.rs
```diff
@@ -429,6 +431,55 @@ impl pallet_dca::Config for Runtime {
...
+
+		for trade in route {
+			let amm_weight = match trade.pool {
+				PoolType::Omnipool => weights::omnipool::HydraWeight::<Runtime>::trade_execution_sell(),
+				PoolType::LBP => weights::lbp::HydraWeight::<Runtime>::trade_execution_sell(),
+				PoolType::Stableswap(_) => weights::omnipool::HydraWeight::<Runtime>::trade_execution_sell(), // TODO: replace by stableswap weights
+				PoolType::XYK => weights::omnipool::HydraWeight::<Runtime>::trade_execution_sell(), // TODO: replace by XYK weights
+			};
+			weight.saturating_accrue(amm_weight);
+			weight.saturating_accrue(Self::sell_overhead_weight());
```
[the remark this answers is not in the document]
Roznovjak: Well, we iterate twice over the route, so there are 2 possible ways how to benchmark it.
1. What we do now. Benchmark it for just one trade and add overhead for every trade in a route. This is the most accurate for routes with one trade.
2. Benchmark it for `MaxRouteLength` and add it just once to the weight. This is the most accurate for trades with `MaxRouteLength` number of trades.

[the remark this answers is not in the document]
Roznovjak: There is also the third option: benchmark it with parameterized number of trades.

runtime/hydradx/src/assets.rs
```diff
@@ -431,6 +437,79 @@ impl pallet_dca::Config for Runtime {
 	type WeightInfo = weights::dca::HydraWeight<Runtime>;
 }
 
+// Provides weight info for the router
+pub struct AmmWeights;
+impl AmmWeights {
+	pub fn sell_overhead_weight() -> Weight {
+		weights::route_executor::HydraWeight::<Runtime>::sell_in_lbp()
```
mrq1911, line 444: it would be probably worth to comment what this does :) i can imagine looking at this in 6mounths being like .... what?

runtime/hydradx/src/benchmarking/omnipool.rs
```diff
@@ -433,6 +436,128 @@ runtime_benchmarks! {
...
+
+	}: {
+		assert!(<Omnipool as TradeExecution<RuntimeOrigin, AccountId, AssetId, Balance>>::calculate_sell(PoolType::Omnipool, token_id, stable_id, amount_sell).is_ok());
+		assert!(<Omnipool as TradeExecution<RuntimeOrigin, AccountId, AssetId, Balance>>::execute_sell(RawOrigin::Signed(seller.clone()).into(), PoolType::Omnipool, token_id, stable_id, amount_sell, buy_min_amount).is_ok());
+	}
+	verify {
+		assert!(<Runtime as pallet_omnipool::Config>::Currency::free_balance(stable_id, &seller) >= buy_min_amount);
+	}
+
+	trade_execution_buy {
```
mrq1911, line 500: better naming would be `router_execution_buy` the trade is very generic

mrq1911 (changes requested):

mrq1911 (approved):

PR #670, opened 2023-09-13
feat: stableswap oracle integration

enthusiastmartin:


runtime/adapters/src/lib.rs
```diff
@@ -789,3 +792,100 @@ where
...
+			OnActivityHandler::<Runtime>::on_trade(
+				STABLESWAP_SOURCE,
+				pool_id,
+				state.assets[idx],
+				0, // Correct
+				state.delta[idx],
+				state.shares,
+				state.after[idx],
+			)
+			.map_err(|(_, e)| e)?;
```
apopiak, line 878: This looks wrong?
It looks like you trigger the same asset pair twice per trade?

enthusiastmartin: removed

pallets/stableswap/src/types.rs
```diff
@@ -111,17 +118,47 @@ impl Default for Tradability {
...
+pub struct PoolState<AssetId> {
+	pub assets: Vec<AssetId>,
+	pub before: Vec<Balance>,
+	pub after: Vec<Balance>,
+	pub delta: Vec<Balance>,
+	pub issuance_before: Balance,
+	pub issuance_after: Balance,
+}
+
+pub trait StableswapHooks<AssetId> {
```
apopiak, line 135: docs

runtime/adapters/src/lib.rs
```diff
@@ -789,3 +792,89 @@ where
 		}
 	}
 }
+
+pub struct StableswapHooksAdapter<Runtime>(PhantomData<Runtime>);
```
apopiak, line 796: docs

runtime/adapters/src/lib.rs
```diff
@@ -789,3 +792,89 @@ where
...
+
+		Ok(())
+	}
+
+	fn on_liquidity_changed_weight(n: usize) -> Weight {
+		OnActivityHandler::<Runtime>::on_liquidity_changed_weight().saturating_mul(n as u64)
+	}
+
+	fn on_trade_weight(n: usize) -> Weight {
+		OnActivityHandler::<Runtime>::on_trade_weight().saturating_mul(2 * n as u64)
```
apopiak, line 878: unnecessary 2 * n

runtime/hydradx/src/assets.rs
```diff
@@ -299,7 +300,7 @@ impl pallet_ema_oracle::Config for Runtime {
 	type SupportedPeriods = SupportedPeriods;
 	/// With every asset trading against LRNA we will only have as many pairs as there will be assets, so
 	/// 20 seems a decent upper bound for the forseeable future.
-	type MaxUniqueEntries = ConstU32<20>;
+	type MaxUniqueEntries = ConstU32<40>;
```
apopiak, line 303: adjust comment

pallets/stableswap/src/lib.rs
```diff
@@ -712,14 +774,53 @@ pub mod pallet {
...
+						}
+					})
+					.collect(),
+				delta: assets
+					.into_iter()
+					.map(|v| {
+						if v == asset_in {
+							amount_in
+						} else if v == asset_out {
+							amount_out
```
apopiak, line 812: how would you distinguish positive from negative delta in the hook?

apopiak (commented):
Looks fine, just the question on how to tell positive from negative delta in the hook (or an explanation why that's unnecessary).
Missing docs, too.

runtime/hydradx/src/assets.rs
mrq1911: what is this change based on? do we have enough weight for that?

mrq1911 (changes requested):
pls rerun benchmarks

pallets/ema-oracle/src/lib.rs
```diff
@@ -387,11 +387,15 @@ impl<T: Config> OnTradeHandler<AssetId, Balance> for OnActivityHandler<T> {
...
+		// We assume that zero liquidity values are not valid and can be ignored.
+		if liquidity_a.is_zero() || liquidity_b.is_zero() {
+			log::warn!(
+				target: LOG_TARGET,
+				"Liquidity amounts should not be zero. Source: {source:?}, liquidity: ({liquidity_a},{liquidity_b})"
+			);
 			return Err((Self::on_trade_weight(), Error::<T>::OnTradeValueZero.into()));
 		}
+
 		let price = determine_normalized_price(asset_a, asset_b, liquidity_a, liquidity_b);
```
mrq1911, line 399: @apopiak @enthusiastmartin i dont think this works for stableswap as it just divides reserves to determine price, this works only for xyk and is not the way how stableswap determines spot price (neither LBP)

mrq1911: i think stableswap or LBP or any other amm than XYK has to be able to set price directly

runtime/hydradx/src/assets.rs
apopiak: It's based on the UI showing 16 assets (including LRNA) in the omnipool. Currently all pairs (including the new stableswap pairs) end up in the same accumulator and share this MaxEntries config. With a stablepool of 4 or 2 stablepools of 3 we would hit the cap already.
The weights should be fine if the benchmarks are rerun, just means fewer trades per block (and slightly higher fees).

pallets/ema-oracle/src/lib.rs
jak-pan: https://github.com/galacticcouncil/HydraDX-simulations/blob/01d560611efef167988247579eb897cfc505c47d/hydradx/model/amm/stableswap_amm.py#L111

jak-pan: You'll need a param. This can still be used if you think of it as nom/denom instead (of the result of the equation to get the price) it should be renamed then probably

pallets/stableswap/src/lib.rs
enthusiastmartin: you have the information which one is asset in and asset out. 

altough  in the adapter, we dont use that. we just provide information of delta_amount and state after the trade. so i belive oracle can work this out? 

it is basically similar to what it done in omnipool adapter, you provide deltas and reserves after trade.  It does not specify which one isnegative

pallets/xyk/src/lib.rs
```diff
@@ -459,8 +460,17 @@ pub mod pallet {
...
-				.map_err(|(_w, e)| e)?;
+			T::AMMHandler::on_liquidity_changed(
+				SOURCE,
+				asset_a,
+				asset_b,
+				amount_a,
+				amount_b,
+				liquidity_a,
+				liquidity_b,
+				Ratio::new(liquidity_a, liquidity_b),
```
enthusiastmartin, line 471: @apopiak 

the price is correctly calculated ? I was not sure whether it should be amounts or liquidity.

dmoka: according to the oracle tests, the liquidity should be used here. But Alex might know better...

math/src/stableswap/math.rs
```diff
@@ -611,15 +612,121 @@ pub(crate) fn normalize_value(amount: Balance, decimals: u8, target_decimals: u8
...
+		})?
+		.checked_mul_int(d)?;
+
+	let ann = calculate_ann(reserves.len(), amplification)?;
+
+	let (d, c, xi, n, ann, issuance) = to_u256!(d, c, reserves[asset_idx], n, ann, issuance);
+
+	let xann = xi.checked_mul(ann)?;
+	let p1 = d.checked_mul(xann)?;
+	let p2 = xi.checked_mul(c)?.checked_mul(n + 1)?;
```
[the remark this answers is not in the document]
enthusiastmartin: n is number of assets in pool. i will change just to be safe.

mrq1911 (approved):

pallets/xyk/src/lib.rs
apopiak: Sorry I didn't see this in time. Yes it should be liquidity for XYK

PR #677, opened 2023-09-21
feat: DCA benchmarking

Roznovjak:


pallets/dca/src/lib.rs
```diff
@@ -199,17 +191,27 @@ pub mod pallet {
 						}
 					}
 				}
+				match schedule.order {
```
dmoka, line 194: Note that on line 156 we already accrue the weights

```
				let weight_for_single_execution = Self::get_trade_weight(&schedule.order);
				weight.saturating_accrue(weight_for_single_execution);
```
Maybe this needs to go then?

pallets/dca/src/lib.rs
```diff
@@ -199,17 +191,27 @@ pub mod pallet {
...
+					Order::Buy { route, .. } => {
+						weight.saturating_accrue(T::AmmTradeWeights::buy_and_calculate_buy_trade_amounts_weight(&route))
+					}
+					Order::Sell { route, .. } => weight
+						.saturating_accrue(T::AmmTradeWeights::sell_and_calculate_sell_trade_amounts_weight(&route)),
+				}
 			}
 
 			weight
```
dmoka, line 203: (I comment here as I can not comment unchanged file in the review)

I expected that the [get_trade_weight()](https://github.com/galacticcouncil/HydraDX-node/blob/1de25855e854abe5ba596a1f4b72a0aba5da839f/pallets/dca/src/lib.rs#L1036) function will also change, as it is used to calculate the fee we deduct from the user at the beginning of each trade. 

Or why was it not changed?

runtime/hydradx/src/benchmarking/dca.rs
```diff
@@ -0,0 +1,340 @@
...
+	let token_amount = 200 * ONE;
+	<Currencies as MultiCurrencyExtended<AccountId>>::update_balance(0u32, &caller, token_amount as i128)?;
+
+	Ok(caller)
+}
+
+runtime_benchmarks! {
+	{Runtime, pallet_dca}
+
+	on_initialize_with_buy_trade{
```
dmoka, line 173: there is only an omnipool trade involved within this. The weight of this will be used to explicitly calculate the DCA execution fee in the DCA pallet. Don't we need to incorporate here more trade types such as stablepool?

integration-tests/Cargo.toml
```diff
@@ -184,3 +184,8 @@ std = [
     "hydradx-runtime/std",
     "pallet-staking/std",
 ]
+
+# we don't include integration tests when benchmarking feature is enabled
+runtime-benchmarks = [
```
dmoka, line 189: can you please write a brief summary of what was the main change about these feature flags? I mean, how we intend to use them from now on. Thanks!

pallets/dca/src/lib.rs
```diff
@@ -199,17 +191,27 @@ pub mod pallet {
...
 		}
 	}
 
 	#[pallet::config]
-	pub trait Config: frame_system::Config + pallet_route_executor::Config {
```
dmoka, line 209: nice, no tight coupling anymore!

pallets/dca/src/lib.rs
```diff
@@ -413,7 +417,7 @@ pub mod pallet {
 		/// Emits `Scheduled` and `ExecutionPlanned` event when successful.
 		///
 		#[pallet::call_index(0)]
-		#[pallet::weight(<T as Config>::WeightInfo::schedule())]
+		#[pallet::weight(<T as Config>::WeightInfo::schedule() + <T as Config>::AmmTradeWeights::calculate_buy_trade_amounts_weight(schedule.order.get_route()))]
```
dmoka, line 420: strange to see here buy specific info, namely `calculate_buy_trade_amounts_weight`

What if we schedule a DCA with sell?

pallets/lbp/src/benchmarking.rs
```diff
@@ -139,19 +139,94 @@ benchmarks! {
...
 	}: _(RawOrigin::Signed(caller.clone()), asset_out, asset_in, amount, max_limit)
 	verify{
 		assert_eq!(T::MultiCurrency::free_balance(asset_out, &caller), 999998100000000);
-		assert_eq!(T::MultiCurrency::free_balance(asset_in, &caller), 999998772262335);
-		assert_eq!(T::MultiCurrency::free_balance(asset_in, &fee_collector), 1000000000455474);
+		assert_eq!(T::MultiCurrency::free_balance(asset_in, &caller), 999998851241411);
+	}
+
+	multi_trade_execution_sell {
+		let c in 0..1;	// if c == 1, calculate_sell is executed
```
dmoka, line 159:
```suggestion
		let is_calculate_sell_executed in 0..1;	// if c == 1, calculate_sell is executed
```

pallets/lbp/src/benchmarking.rs
```diff
@@ -139,19 +139,94 @@ benchmarks! {
...
+		let e in 0..1;	// if e == 1, execute_sell is executed
```
dmoka, line 160:
```suggestion
		let is_execute_sell_executed in 0..1;	// if e == 1, execute_sell is executed
```

if you agree with these, you can do similar renamings in other benchmark tests

runtime/hydradx/src/benchmarking/route_executor.rs
```diff
@@ -16,326 +16,180 @@
...
+	System::set_block_number(2u32);
+	Ok(())
 }
 
 runtime_benchmarks! {
-	{ Runtime, pallet_route_executor}
+	{Runtime, pallet_route_executor}
 
+	// Calculates the weight of LBP trade. Used in the calculation to determine the weight of the overhead.
+	multi_trade_execution_sell_in_lbp {
```
dmoka, line 100: can we find a better prefix for this, instead of `multi_trade`? By reading this I expected there are multiple trade with different pools involved. Or what `multi_trade` means in this context?

dmoka (commented):
Added some remarks.

On branch `feat/stableswap-oracle` we also have several changes relating to DCA and route execution, so would be nice to combine these two branches and generate the benchmark all together, instead of doing it 2 times.

runtime/hydradx/src/assets.rs
```diff
@@ -428,9 +481,147 @@ impl pallet_dca::Config for Runtime {
...
+				PoolType::XYK => weights::omnipool::HydraWeight::<Runtime>::trade_execution_buy(), // TODO: replace by XYK weights + AMMHandler::on_trade_weight()
+			};
+			weight.saturating_accrue(amm_weight);
+			weight.saturating_accrue(Self::buy_overhead_weight());
+		}
+
+		weight
+	}
+
+	fn calculate_buy_trade_amounts_weight(route: &[Trade<AssetId>]) -> Weight {
```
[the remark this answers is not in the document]
Roznovjak: We don't need it. DCA `schedule` extrinsic needs just this one.

pallets/dca/src/lib.rs
Roznovjak: The weight on line 156 is for DCA part, and the second weight is for the router. We use dummy router in the benchmarks so we need to add additional weight.

pallets/dca/src/lib.rs
Roznovjak: We don't need to change it because it now benchmarks only DCA code. The router and AMMs are benchmarked separately.

runtime/hydradx/src/benchmarking/dca.rs
Roznovjak: Trade type used in the benchmarks doesn't matter because we use dummy router that does nothing.

pallets/dca/src/lib.rs
Roznovjak: Hmm. We don't call the router for sell. Only for buy. This can be optimized to add the weight only if trade == buy.

pallets/lbp/src/benchmarking.rs
Roznovjak: Unfortunately these parameters have to be named with only one char.

runtime/hydradx/src/benchmarking/route_executor.rs
Roznovjak: You are right that the naming is not the best. I will try to find a better name. I prefixed it with multi because these benchmarks are parameterized and can return weight for different combinations of calls (just the first one or the second one, both calls together, the first one called two times). Maybe we can rename it to something like `calculate_and_execute_buy_in_lbp(c, e)`

runtime/hydradx/src/assets.rs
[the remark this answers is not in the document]
dmoka: What is the reason that it needs only this one?

pallets/dca/src/lib.rs
dmoka: I wonder how much we gain with that optimization. If you think we would gain a lot of saved ref time, we could optimize it, otherwise just leave.

pallets/lbp/src/benchmarking.rs
dmoka: too bad! :D Next time!

runtime/hydradx/src/benchmarking/route_executor.rs
dmoka: that's more intention revealing, let's go for it!

integration-tests/src/router.rs
```diff
@@ -6,7 +6,11 @@ use crate::polkadot_test_net::*;
 use std::convert::Into;
```
dmoka, line 6: How about adding an integration test to include all pool types?

Roznovjak: Do you mean to test a trade with all pool types?

pallets/dca/src/lib.rs
mrq1911: i dont understand why is even here, the route shouldn't change weight of the schedule extrinsic as the execution is not even done here

mrq1911 (approved):

pallets/dca/src/lib.rs
Roznovjak: Because the extrinsic calls `Self::get_amount_in_for_buy()` which calls `Router::calculate_buy_trade_amounts()`

Issue #684, opened 2023-10-04
Forbid XYK pools containing LRNA


mrq1911:
add call & transact filter calls containing LRNA as one of the params for `XYK:::create_pool`

PR #685, opened 2023-10-05
feat: forbid XYK pools containing LRNA

Roznovjak:
Note: Call filter is called from Transact filter, so there is no need to update Transact filter


mrq1911:
fixes #684

PR #692, opened 2023-10-19
feat: add support for POL withdrawal to omnipool

enthusiastmartin:
This PR adds the two following extrinsics:

- withdraw_protocol_liquidity - adds to support to withdraw protocol liquidity given known price at the time
- remove_token - removes tokens from omnipool given that only protocol owns shares

This PR also changes remove_liquidity
- it skips price check and circuit breaker check if withdrawal is safe. Withdrawal is considered safe when asset tradability state is set to only REMOVE_LIQUIDITY or REMOVE_LIQUIDITY | ADD_LIQUIDITY.  Basically, when trading is disabled.


pallets/omnipool/src/lib.rs
```diff
@@ -1496,6 +1527,138 @@ pub mod pallet {
...
+		#[pallet::weight(<T as Config>::WeightInfo::withdraw_protocol_liquidity())]
+		#[transactional]
+		pub fn withdraw_protocol_liquidity(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+			amount: Balance,
+			price: (Balance, Balance),
+			dest: T::AccountId,
+		) -> DispatchResult {
+			ensure_root(origin.clone())?;
```
mrq1911, line 1541: this should be autority origin instead, and same for remove token

mrq1911 (changes requested):

pallets/omnipool/src/lib.rs
```diff
@@ -1946,4 +2124,18 @@ impl<T: Config> Pallet<T> {
 		);
 		Ok(())
 	}
+
+	pub fn process_hub_amount(amount: Balance, dest: &T::AccountId) -> DispatchResult {
+		if amount > Balance::zero() {
+			// If transfers fails and the amount is less than ED, it failed due to ED limit, so we simply burn it
+			if let Err(e) = T::Currency::transfer(T::HubAssetId::get(), &Self::protocol_account(), dest, amount) {
+				if amount < 400_000_000u128 {
+					T::Currency::withdraw(T::HubAssetId::get(), &Self::protocol_account(), amount)?;
```
mrq1911, line 2133: if you burn LRNA shouldn't you also update imbalance?

mrq1911 (approved):

PR #695, opened 2023-10-30
feat: runtime-mock and invariants

vgantchev:
This PR adds:
* runtime-mock - TestExternalities of the HydraDX runtime with rich data setup. This is used by the fuzzer and can be used by local (runtime integration) tests
* invariants - these are implemented as runtime panics! in try-runtime (function-level invariants) and try_state (system-level invariants)


pallets/omnipool/src/lib.rs
```diff
@@ -1286,6 +1301,8 @@ pub mod pallet {
 
 			Self::update_imbalance(state_changes.delta_imbalance)?;
 
+			assert!(asset_in_state.reserve < new_asset_in_state.reserve);
```
vgantchev, line 1304: @enthusiastmartin is this line kept outside of `try-runtime` feature on purpose?

Cargo.lock
```diff
@@ -77,19 +77,19 @@ version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
 dependencies = [
- "getrandom 0.2.11",
```
vgantchev, line 80: @enthusiastmartin did you run `cargo update`? 

Was this necessary for something, shall we keep the version changes or revert?

vgantchev: restored old Cargo.lock to fixb uild

runtime-mock/src/registry.rs
```diff
@@ -0,0 +1,58 @@
+use hydradx_runtime::RuntimeCall;
```
vgantchev, line 1: Registry mock is incompatible with the new asset registry

pallets/omnipool/src/lib.rs
enthusiastmartin: should be removed. i think i was testing something and left it there accidentally.

mrq1911 (approved):
fix tests & LGTM

PR #696, opened 2023-11-02
refactor: remove dependency on primitives from pallet-lbp

Roznovjak:
This PR removes dependency on primitives from pallet-lbp so lbp can be reused in Basilisk repo.
AssetPair is also removed from primitives because it's not used anymore.


pallets/lbp/src/lib.rs
```diff
@@ -36,7 +37,7 @@ use frame_system::ensure_signed;
 use hydra_dx_math::types::LBPWeight;
 use hydradx_traits::{AMMTransfer, AssetPairAccountIdFor, CanCreatePool, LockedBalance, AMM};
 use orml_traits::{MultiCurrency, MultiCurrencyExtended, MultiLockableCurrency};
-use primitives::{asset::AssetPair, Amount, AssetId, Balance};
+use primitives::Amount;
```
enthusiastmartin, line 40: the title is "remove dependcy on primitivies" and yet, there is still ?!

Roznovjak: :D yeah. The same thing in the previous PR https://github.com/galacticcouncil/HydraDX-node/pull/693/files#diff-3abe38ebc0eac7c197e17a967785f9c8b2acc2eda9863ee50d5dfb8109d4fde1

enthusiastmartin: missed that as it was not something that changed. 

that should be removed too as it is the purpose of the PR.

Roznovjak: Updated. Also for XYK.

enthusiastmartin (approved):

PR #698, opened 2023-11-08
feat: dont allow remove vote if referendum is finished

enthusiastmartin:
When a referendum is finished and staking position exists - it is no longer possible to remove vote.

It also fixes an issue where user does not get action in some specific scenarios.


pallets/staking/src/lib.rs
```diff
@@ -282,6 +282,9 @@ pub mod pallet {
 		/// Signer is not an owner of the staking position.
 		Forbidden,
 
+		/// Remove vote is not allowed when referendum is finished and staking position exists.
+		RemoveVoteNotAllowed,
```
mrq1911, line 286: name is too generic, just call it `StakingPositionsExists`

integration-tests/src/staking.rs
```diff
@@ -955,3 +955,60 @@ fn staking_should_assign_less_action_points_when_portion_of_staking_lock_is_vest
...
+		)
+		.unwrap()
+		.unwrap();
+		let stake_voting = pallet_staking::Pallet::<hydradx_runtime::Runtime>::get_position_votes(stake_position_id);
+		assert!(!stake_voting.votes.is_empty());
+		let (ref_vote_idx, vote) = stake_voting.votes[0];
+		assert_eq!(ref_vote_idx, 0);
+		assert_eq!(
+			vote,
+			pallet_staking::types::Vote::new(2 * UNITS, pallet_staking::types::Conviction::None)
```
mrq1911, line 1003: you should be able to remove vote if the conviction timer runs out even if the position exists, please test that this is possible

integration-tests/src/staking.rs
```diff
@@ -955,3 +955,60 @@ fn staking_should_assign_less_action_points_when_portion_of_staking_lock_is_vest
 		assert_eq!(position.get_action_points(), 50_u128);
 	});
 }
+
+#[test]
+fn staking_should_not_allow_to_remove_vote_when_referendum_is_finished_and_staking_position_exists() {
```
mrq1911, line 960: also test that you can remove vote after you unstake

mrq1911 (changes requested):

integration-tests/src/staking.rs
mrq1911: therefore this test should fail

pallets/staking/src/lib.rs
enthusiastmartin: that's little bit confusing imo. there is already similar error that staking position exists but used in different context.

i believe it should explicitly state that remove vote is not allowed.

integration-tests/src/staking.rs
enthusiastmartin: reworked, added more tests with various scenarios

pallets/democracy/src/traits.rs
```diff
@@ -1,9 +1,15 @@
 use crate::{AccountVote, ReferendumIndex};
 use frame_support::dispatch::DispatchResult;
 
+pub enum Status<B> {
```
martinfridrich, line 4: Is this used? I run tests without it and all tests passed.

martinfridrich (approved):

pallets/democracy/src/traits.rs
enthusiastmartin: leftover.will delete

integration-tests/src/staking.rs
```diff
@@ -955,3 +957,357 @@ fn staking_should_assign_less_action_points_when_portion_of_staking_lock_is_vest
...
+		let stake_position_id = pallet_staking::Pallet::<hydradx_runtime::Runtime>::get_user_position_id(
+			&sp_runtime::AccountId32::from(BOB),
+		)
+		.unwrap()
+		.unwrap();
+		assert_noop!(
+			Democracy::remove_vote(hydradx_runtime::RuntimeOrigin::signed(BOB.into()), r),
+			pallet_staking::Error::<hydradx_runtime::Runtime>::RemoveVoteNotAllowed
+		);
+		assert_ok!(Democracy::unlock(
```
mrq1911, line 1032: why this succeeds?

pallets/democracy/src/lib.rs
```diff
@@ -1266,12 +1267,26 @@ impl<T: Config> Pallet<T> {
...
+
+							should_lock
 						}
 					}
-					None => {} // Referendum was cancelled.
-				}
+					None => false, // Referendum was cancelled.
+				};
+
 				votes.remove(i);
```
mrq1911, line 1288: wait, you still remove vote but not lock?

i guess it is always supposed to be reverted, otherwise user could get stuck. can you pls add test that user can in fact remove vote after conviction expires?

integration-tests/src/staking.rs
mrq1911: i guess it always succeeds, but just do nothing

pallets/democracy/src/lib.rs
enthusiastmartin: no, that's not correct. if you dont have staking positions, it behaves like it used to.  meaning, you can always remove your vote.

integration-tests/src/staking.rs
enthusiastmartin: calling unlock checks and figures out if you can unlock or not. and it always returns success ( errors only if not signed or the lookup of target fails).

PR #702, opened 2023-11-15
fix: copy evm-utility module from acala

dmoka:
## Description

Copy the evm-utility module used in currency precompiles that implements the ERC20 standard.

We copy because referencing Acala takes a lot of space, and increased build time too

## Checklist:


runtime/hydradx/src/evm/evm-utility/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "module-evm-utility"
+version = "2.21.0"
+authors = ["Acala Developers"]
+edition = "2021"
+
+[dependencies]
+sha3 = { version = "0.10.6", default-features = false }
+
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v1.0.0", default-features = false }
```
mrq1911, line 10: please adjust dependency so it is consistent we rest

enthusiastmartin (approved):
updated the dependencies and we can merge

PR #709, opened 2023-12-05
feat: support eth address and remote accounts

dmoka:
## 1. Support ETH address in XCM
We want to support MultiLocation with Ethereum (Account20) address, so people can send xcm to ethereum address of our chain, where we handle the conversion ourselves

## 2. Allow remote accounts
Allowing to have remote accounts on Hydra for all purposes a native account could be used. Especially useful for when we do not want to deteriorate UX by switching networks in the UI.

Currently, there is no common ecosystem wide approach to derive accounts from xcm-origins. This PR adds a converter both used by

- [Kusama Asset-Hub](https://github.com/polkadot-fellows/runtimes/blob/b8df4b06c78fb2fb2d0f081c76e30569f12363c6/system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs#L91)
- [Centrifuge](https://github.com/centrifuge/centrifuge-chain/blob/93ace412bc269522916cc191caede54733921fa0/runtime/centrifuge/src/xcm.rs#L322)
- [Moonbeam](https://github.com/moonbeam-foundation/moonbeam/blob/3d6eb4e1de334c3259631ac15373e130c48c00af/runtime/moonbeam/src/xcm_config.rs#L111)

The converter was introduced by https://github.com/paritytech/polkadot/pull/7329 in order to create a general purpose conversion. The idea behind this converter is that a user User1 from ChainA has the ability to have a unique account on any other chain in the ecosystem. This so called remote account can only be controlled from User1 from ChainA.


runtime/hydradx/src/xcm.rs
```diff
@@ -306,8 +309,41 @@ pub type LocationToAccountId = (
...
+			} => {
+				let account_32 = ExtendedAddressMapping::into_account_id(H160::from(key));
+				Ok(account_32)
+			}
+			_ => Err(location),
+		}
+	}
+
+	//TODO: fix this reverse
+	fn reverse(who: AccountId) -> Result<MultiLocation, AccountId> {
```
dmoka, line 338: wondering if this is correct, of we should just throw Err in this case

green-jay: iirc according to @mrq1911 we should use last 20 bytes from 32-byte address to determine it

green-jay (approved):
works like a charm, LGTM

PR #714, opened 2023-12-08
feat: xcm exchange asset

dmoka:
## Description

We enable AssetExchange xcm message instruction in hydra.

It also means that we needed to change the `XcmFilter` of pallet_xcm from `Nothing` to `AllowTransferAndSwap`, which is a more permissive in terms of allowed xcm instructions.

## Checklist:


runtime/hydradx/src/xcm.rs
```diff
@@ -210,7 +216,7 @@ impl pallet_xcm::Config for Runtime {
 	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
 	type XcmRouter = XcmRouter;
 	type ExecuteXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
-	type XcmExecuteFilter = Nothing;
+	type XcmExecuteFilter = AllowTransferAndSwap<MaxXcmDepth, MaxNumberOfInstructions, RuntimeCall>;
```
mrq1911, line 219: why is this change necessary? i dont think we are planning to execute xcm messages locally and xcm::execute extrinsic is call filtered anyway currently

mrq1911 (changes requested):
did you try to exchange something from some sibling parachain using chopstics?

runtime/hydradx/src/xcm.rs
apopiak: We want it in combination with changing the call filter to allow UIs to interact with the chain through `pallet_xcm::execute`. This is part of the XCM exchange asset stuff.
Concretely you could imagine our UI swapping tokens on Hydra and then sending them to the destination in one extrinsic/with one XCM.

jak-pan (commented):
Any problems with this?

PR #715, opened 2023-12-09
feat: evm import queue enabled after upgrade

mrq1911:
the current node rejects blocks without frontier log - all before the upcoming upgrade to 21

this PR introduces a filter since when should be log expected


node/res/hydradx.json
```diff
@@ -24,6 +24,7 @@
   },
   "relay_chain": "polkadot",
   "para_id": 2034,
+  "evm_since": 4010000,
```
mrq1911, line 27: TBD when EVM runtime upgrade is enacted and active

vgantchev (approved):

PR #727, opened 2023-12-22
feat!: XCM Rate Limiter Integration for Polkadot v1.1

apopiak:
## Description

Port #638 to Polkadot v1.1

## Checklist:


integration-tests/src/lib.rs
```diff
@@ -21,4 +21,19 @@ mod staking;
 mod transact_call_filter;
 mod vesting;
 mod xcm_defer;
+mod xcm_rate_limiter;
+
+#[macro_export]
+macro_rules! assert_balance {
```
Roznovjak, line 27: Why do we want to have these macros here? I'm just curious. I wouldn't expect any code in this file.

Roznovjak (approved):

PR #758, opened 2024-02-09
fix: staking slashing

martinfridrich:
## Description
Staking rewards calculation update and introduction of min. slash.

### DIFF
--- a/math/src/staking/math.rs
+++ b/math/src/staking/math.rs
@@ -37,9 +37,9 @@ pub fn calculate_slashed_points(
 	stake_weight: u8,
 ) -> Option<Balance> {
 	let stake_weighted = current_stake.checked_mul(stake_weight.into())?;
-    let p = stake_increase.checked_mul(points)?;
+	let p = stake_increase.checked_mul(points)?;
 
-    p.checked_div(stake_weighted)?.min(points).into()
+	p.checked_div(stake_weighted)?.min(points).into()
 }
 
 /// Function calculates period number from block number and period size.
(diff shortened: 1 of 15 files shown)


math/src/staking/math.rs
```diff
@@ -37,9 +37,9 @@ pub fn calculate_slashed_points(
 	stake_weight: u8,
 ) -> Option<Balance> {
 	let stake_weighted = current_stake.checked_mul(stake_weight.into())?;
-	FixedU128::checked_from_rational(stake_increase, stake_weighted)?
-		.min(FixedU128::one())
-		.checked_mul_int(points)
+	let p = stake_increase.checked_mul(points)?;
```
poliwop, line 40: The math here is correct and I believe decreases rounding error.

In Omnipool calculations when we do this we convert to U256 so that when we multiply the two U128s we do not get overflow. We do not appear to be doing that here. Did we consider the magnitudes of the quantities involved and decide that since the points quantity is low we do not need to convert to U256? Won't we run into issues as people continue to accrue points over time?

poliwop (commented):
This review only covers the change to calculate_slashed_points in staking/math.rs. The math looks correct, but see my question about whether we should be converting to U256.

pallets/staking/src/lib.rs
```diff
@@ -443,43 +446,34 @@ pub mod pallet {
...
 					position.reward_per_stake = staking.accumulated_reward_per_stake;
 
 					let points =
 						Self::get_points(position, current_period, created_at).ok_or(Error::<T>::Arithmetic)?;
 					let slash_points =
 						math::calculate_slashed_points(points, position.stake, amount, T::CurrentStakeWeight::get())
-							.ok_or(Error::<T>::Arithmetic)?;
+							.ok_or(Error::<T>::Arithmetic)?
+							.max(T::MinSlash::get(&payable_percentage))
+							.min(points);
```
enthusiastmartin, line 476: i would move this into the math function.

pallets/staking/src/lib.rs
```diff
@@ -443,43 +446,34 @@ pub mod pallet {
...
+					.ok_or(Error::<T>::Arithmetic)?;
+
+					if !rewards.is_zero() {
+						let pot = Self::pot_account_id();
+						T::Currency::transfer(T::NativeAssetId::get(), &pot, &who, rewards)?;
+
+						position.accumulated_locked_rewards = position
+							.accumulated_locked_rewards
+							.checked_add(rewards)
+							.ok_or(Error::<T>::Arithmetic)?;
```
enthusiastmartin, line 464: is this correct? 

to me, it looks like you send the `rewards` to `who` and then you add the same amount to locked rewards ? 

Doesnt this mean that it has more accumulated rewards than he should ?

martinfridrich: This is correct. Increase stake locks rewards, so we send new rewards to user and then we locked them here: https://github.com/galacticcouncil/HydraDX-node/blob/0b9b0f1bba50d8e2cf4f6bb79c1aceb2c843d58b/pallets/staking/src/lib.rs#L496

enthusiastmartin (approved):

PR #760, opened 2024-02-12
fix: router Executed event

dmoka:
## Checklist:

### DIFF
--- a/pallets/route-executor/Cargo.toml
+++ b/pallets/route-executor/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-route-executor'
-version = '1.3.2'
+version = '2.0.0'
 description = 'A pallet to execute a route containing a sequence of trades'
 authors = ['GalacticCouncil']
 edition = '2021'
--- a/pallets/route-executor/Cargo.toml
+++ b/pallets/route-executor/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-route-executor'
-version = '1.3.1'
+version = '1.3.2'
 description = 'A pallet to execute a route containing a sequence of trades'
 authors = ['GalacticCouncil']
 edition = '2021'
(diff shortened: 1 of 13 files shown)


pallets/route-executor/src/lib.rs
```diff
@@ -268,18 +270,17 @@ pub mod pallet {
...
 				T::Currency::reducible_balance(asset_in, &who, Preservation::Preserve, Fortitude::Polite);
 
 			let trade_amounts = Self::calculate_buy_trade_amounts(&route, amount_out)?;
 
-			let last_trade_amount = trade_amounts.last().ok_or(Error::<T>::RouteCalculationFailed)?;
-			ensure!(
-				last_trade_amount.amount_in <= max_amount_in,
-				Error::<T>::TradingLimitReached
-			);
+			let first_trade = trade_amounts.last().ok_or(Error::<T>::RouteCalculationFailed)?;
```
mrq1911, line 282: last() is first? please add test to ensure you have this right

dmoka: the `calculate_buy_trade_amounts` calculates the trades from the end as we have the `amount_out` specified. So the last of that list will be the `first_trade`, hence the renaming.

we cover this with test.

pallets/route-executor/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-route-executor'
-version = '1.3.1'
+version = '1.3.2'
```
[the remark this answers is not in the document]
dmoka: bumping it to 2 as it is breaking change.

pallets/route-executor/src/lib.rs
mrq1911: but funcionally you didnt change anything, just renamed `last_trade_amount` to `first_trade`

mrq1911 (approved):

pallets/route-executor/src/lib.rs
mrq1911: the problem is in the asset in vs route asset in inconsistency

PR #763, opened 2024-02-15
feat: evm fee in other currencies

enthusiastmartin:
This PR adds support to pay evm fee in account's fee payment see which user sets using multi-transaction-payment-pallet.


runtime/adapters/src/price.rs
```diff
@@ -19,3 +28,76 @@ where
...
+			return (0,Weight::zero());
+		};
+		let account_balance = I::reducible_balance(from_currency, account, Preservation::Preserve, Fortitude::Polite);
+
+		if from_currency == to_currency {
+			return (account_balance, T::DbWeight::get().reads(2));
+		}
+		// We get the weight from the ema-oracle weights to get price
+		// Weight * 2 because we are reading from the storage twice ( from_currency/lrna and lrna/to_currency)
+		// TODO: it could really be a part of the PriceOracle trait?!
```
dmoka, line 56: todo

runtime/adapters/src/price.rs
```diff
@@ -19,3 +28,76 @@ where
...
+		// if this gets removed, the constraint on T and ema-oracle is not necessary
+		let price_weight = pallet_ema_oracle::Pallet::<T>::get_price_weight()
+			.saturating_mul(2)
+			.saturating_add(T::DbWeight::get().reads(2));
+		let Some(price) = P::price(&[Trade {
```
dmoka, line 61: what if we have a fee asset in other pool like stablepool? If we want so, we could get the route from onchain.

The same is true for `ConvertToCurrencyUsingOracle`

runtime/hydradx/src/evm/evm_fee.rs
```diff
@@ -0,0 +1,184 @@
...
+impl OnUnbalanced<EvmPaymentInfo<EmaPrice>> for DepositEvmFeeToTreasury {
+	// this is called for substrate-based transactions
+	fn on_unbalanceds<B>(amounts: impl Iterator<Item = EvmPaymentInfo<EmaPrice>>) {
+		Self::on_unbalanced(amounts.fold(EvmPaymentInfo::default(), |i, x| x.merge(i)))
+	}
+
+	// this is called from pallet_evm for Ethereum-based transactions
+	// (technically, it calls on_unbalanced, which calls this when non-zero)
+	fn on_nonzero_unbalanced(payment_info: EvmPaymentInfo<EmaPrice>) {
+		//TODO: perhaps better to avoid using something multi payment pallet - it is fine for now
```
dmoka, line 175: todo

runtime/hydradx/src/evm/runner.rs
```diff
@@ -0,0 +1,238 @@
+use crate::evm::WethAssetId;
```
dmoka, line 1: since most lines are copied from frontier, maybe we could acknowledge it in a comment?

traits/src/lib.rs
```diff
@@ -256,3 +256,15 @@ pub trait AMMPosition<AssetId, Balance> {
 		shares_amount: Balance,
 	) -> Result<(Balance, Balance), Self::Error>;
 }
+
+// Provides account's fee payment asset
```
dmoka, line 260:
```suggestion
/// Provides account's fee payment asset
```

traits/src/lib.rs
```diff
@@ -256,3 +256,15 @@ pub trait AMMPosition<AssetId, Balance> {
...
+pub trait FeePaymentCurrency<AccountId> {
+	type AssetId;
+	fn get(a: &AccountId) -> Option<Self::AssetId>;
+}
+
+// Provides account's fee paymentasset balance in a given currency
```
dmoka, line 266:
```suggestion
/// Provides account's fee paymentasset balance in a given currency
```

runtime/hydradx/src/evm/mod.rs
```diff
@@ -142,11 +135,31 @@ impl pallet_evm::Config for crate::Runtime {
 	type FeeCalculator = crate::DynamicEvmFee;
 	type FindAuthor = FindAuthorTruncated<Aura>;
 	type GasWeightMapping = pallet_evm::FixedGasWeightMapping<Self>;
-	type OnChargeTransaction = TransferEvmFees<DealWithFees>;
+	type OnChargeTransaction = evm_fee::TransferEvmFees<
+		evm_fee::DepositEvmFeeToTreasury,
+		crate::MultiTransactionPayment, // Get account's fee payment asset
+		WethAssetId,
+		ConvertToCurrencyUsingOracle<
```
mrq1911, line 142: why is this necessary? `OraclePriceAdapter` should do everything you need same as for multi currency payment, including the specification of `Period` which is `SHORT` by default, not last block. `SHORT` is used by multicurrency as well

runtime/hydradx/src/evm/mod.rs
```diff
@@ -142,11 +135,31 @@ impl pallet_evm::Config for crate::Runtime {
...
+		FungibleCurrencies<crate::Runtime>, // Multi currency support
+	>;
 	type OnCreate = ();
 	type PrecompilesType = precompiles::HydraDXPrecompiles<Self>;
 	type PrecompilesValue = PrecompilesValue;
-	type Runner = pallet_evm::runner::stack::Runner<Self>;
+	type Runner = WrapRunner<
+		Self,
+		pallet_evm::runner::stack::Runner<Self>, // Evm runner that we wrap
+		hydradx_adapters::price::FeeAssetBalanceInCurrencyConvertor<
```
mrq1911, line 155: another convertor ? why you cannot reuse the same as in `OnChargeTransaction`

runtime/hydradx/src/evm/runner.rs
mrq1911: yeah, i would add it so we can upgrade it as well when frontier runner changes

runtime/hydradx/src/evm/mod.rs
mrq1911: what about `AssetFeeOraclePriceProvider` ? :)

mrq1911 (changes requested):

runtime/adapters/src/price.rs
mrq1911: why are we always writing new price adapter, this should be resolved 10 times already

enthusiastmartin: i will change to this to use the exiting price with route. makes sense

runtime/hydradx/src/evm/mod.rs
enthusiastmartin: agree, i can simplify and reuse.

runtime/hydradx/src/evm/mod.rs
```diff
@@ -130,6 +122,8 @@ parameter_types! {
 	/// The amount of gas per storage (in bytes): BLOCK_GAS_LIMIT / BLOCK_STORAGE_LIMIT
 	/// The current definition of BLOCK_STORAGE_LIMIT is 40 KB, resulting in a value of 366.
 	pub GasLimitStorageGrowthRatio: u64 = 366;
+
+	pub const OracleEvmPeriod: OraclePeriod = OraclePeriod::LastBlock;
```
mrq1911, line 126: this should be the same as period used to determine substrate multipayment fees (short)

mrq1911 (changes requested):

mrq1911 (approved):

PR #765, opened 2024-02-20
fix: adjust router for insufficient asset

dmoka:
Since we introduced insufficient assets, the balance check won't work properly as we can have special case with selling fee payment asset for shitcoin, so we decided to remove it.

Furthermore we wanna disable route setting for insufficient asset

## TODO:
- [x] Rebenchmark pallet


integration-tests/src/dca.rs
```diff
@@ -2853,6 +2854,8 @@ mod with_onchain_route {
 
 		TestNet::reset();
 		Hydra::execute_with(|| {
+			let ps = frame_system::Pallet::<Runtime>::providers(&AccountId32::from(ALICE));
```
mrq1911, line 2857: what those ps variables do?

pallets/route-executor/src/lib.rs
```diff
@@ -310,6 +316,7 @@ pub mod pallet {
 				asset_in,
 				user_balance_of_asset_in_before_trade,
 				first_trade.amount_in,
+				existential_deposit, //TODO: add test for buy then fix it,
```
mrq1911, line 319: TODO

pallets/route-executor/src/lib.rs
```diff
@@ -454,16 +461,23 @@ impl<T: Config> Pallet<T> {
...
 
-			ensure!(
-				user_balance_of_asset_in_before_trade - spent_amount == user_balance_of_asset_in_after_trade,
-				Error::<T>::InvalidRouteExecution
-			);
+		let expected_user_balance_of_asset_in_after_trade = user_balance_of_asset_in_before_trade
+			.checked_sub(&spent_amount)
+			.ok_or(Error::<T>::InvalidRouteExecution)?;
+
+		if expected_user_balance_of_asset_in_after_trade < existential_deposit {
```
mrq1911, line 473: this could be simply just part of `ensure!` predicate

mrq1911 (changes requested):
pls bump versions

integration-tests/src/dca.rs
dmoka: it is still draft, about to removing them and clean the PR x)

dmoka:
dont merge it yet, we will rework to get the ED from asset registry

pallets/route-executor/src/lib.rs
```diff
@@ -454,16 +466,21 @@ impl<T: Config> Pallet<T> {
...
-				Error::<T>::InvalidRouteExecution
-			);
-		}
+		let expected_user_balance_of_asset_in_after_trade = user_balance_of_asset_in_before_trade
+			.checked_sub(&spent_amount)
+			.ok_or(Error::<T>::InvalidRouteExecution)?;
+
+		//If the user had leftover less than ED then it is wiped out, hence we can't check the balance precisely
+		ensure!(
+			expected_user_balance_of_asset_in_after_trade < existential_deposit
```
jak-pan, line 480: We can also check if it's 0

dmoka: You would check if the `user_balance_of_asset_in_after_trade` is equal 0?

dmoka: what is the reason for checking it for 0?

mrq1911 (approved):
Rebench and LGTM

PR #770, opened 2024-02-23
chore: tips and elections benchmarking

Roznovjak:


jak-pan (approved):
We didn't do this before? (We had custom settings even b4)

Roznovjak:
> We didn't do this before? (We had custom settings even b4)

We probably missed this long time ago and didn't notice until now. Every time we do rebenchmarking, we just rebenchmark the weights in the `runtime/weights` directory.

PR #782, opened 2024-03-05
feat: Manual oracle activation

Roznovjak:


pallets/ema-oracle/src/lib.rs
```diff
@@ -149,10 +152,17 @@ pub mod pallet {
...
 		OnTradeValueZero,
+		OracleNotFound,
 	}
 
 	#[pallet::event]
-	pub enum Event<T: Config> {}
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Oracle was added to the whitelist.
+		OracleAdded { source: Source, assets: (AssetId, AssetId) },
```
mrq1911, line 162: when you add event like this, i would expect it being fired every time oracle is created, not just when you modify whitelist

pallets/ema-oracle/src/lib.rs
```diff
@@ -232,7 +247,43 @@ pub mod pallet {
...
+			})?;
+
+			Self::deposit_event(Event::OracleAdded { source, assets });
+
+			Ok(())
+		}
+
+		#[pallet::call_index(1)]
+		#[pallet::weight(<T as Config>::WeightInfo::remove_oracle())]
+		pub fn remove_oracle(origin: OriginFor<T>, source: Source, assets: (AssetId, AssetId)) -> DispatchResult {
```
mrq1911, line 270: i would expect this function would remove actual oracle as well, not just item from whitelist

mrq1911 (changes requested):
please document new extrinsics

pallets/ema-oracle/src/lib.rs
```diff
@@ -243,7 +294,10 @@ impl<T: Config> Pallet<T> {
 		assets: (AssetId, AssetId),
 		oracle_entry: OracleEntry<BlockNumberFor<T>>,
 	) -> Result<(), ()> {
-		if !T::OracleWhitelist::contains(&(src, assets.0, assets.1)) {
+		if !(T::OracleWhitelist::contains(&(src, assets.0, assets.1))
+			|| WhitelistedAssets::<T>::get().contains(&(src, (assets.0, assets.1))))
```
mrq1911, line 298: i dont like the way this is tacked on, from configuration of the pallet i would expect that only `OracleWhitelist` controls it, nothing else, therefore this list should be only contains implementation (default ()) so if you override it in config, it wont be utilized

mrq1911: so for our case you would basically implement extension in the runtime config, not other way around

mrq1911 (approved):

PR #783, opened 2024-03-12
feat: min liquidity limit in omnipool::add_token

Roznovjak:


pallets/omnipool/src/lib.rs
```diff
@@ -468,8 +468,11 @@ pub mod pallet {
 
 			let amount = T::Currency::free_balance(asset, &Self::protocol_account());
 
+			let ed = T::AssetRegistry::existential_deposit(asset);
+			let minimum_pool_liquidity = ed.checked_mul(20).ok_or(ArithmeticError::Overflow)?;
```
enthusiastmartin, line 472: saturating_mul is enough here.

if the max value is reached, you would not be able to have a balance anyway.

pallets/omnipool/src/lib.rs
```diff
@@ -468,8 +468,11 @@ pub mod pallet {
...
+
 			ensure!(
-				amount >= T::MinimumPoolLiquidity::get() && amount > 0,
+				amount >= minimum_pool_liquidity && amount > 0,
```
enthusiastmartin, line 475: if check that ed> 0 then there is no need to check that amount > 0.

pallets/omnipool/src/lib.rs
```diff
@@ -468,10 +468,10 @@ pub mod pallet {
 
 			let amount = T::Currency::free_balance(asset, &Self::protocol_account());
 
-			ensure!(
-				amount >= T::MinimumPoolLiquidity::get() && amount > 0,
-				Error::<T>::MissingBalance
-			);
+			let ed = T::AssetRegistry::existential_deposit(asset).unwrap_or(u128::MAX);
```
enthusiastmartin, line 471: maybe it would be better to return an error here.

because the error returned would be potentially misleading ( it would return MissingBalance).

enthusiastmartin (approved):

Issue #785, opened 2024-03-13
Add signedExtension fee payment check for batch txs


jak-pan:
Currently there is no way to switch fee payment asset and execute tx in one go. We were suggested to add a check for first item in batch to be able to change fee payment asset in this way.

PR #788, opened 2024-03-14
feat: allow set_currency in batch TX

Roznovjak:
Allow `set_currency` in `batch`, `batch_all` and `force_batch` if it's the first TX.
Fixes: #785


pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -378,18 +379,33 @@ where
...
+		} else if let Some(pallet_utility::pallet::Call::<T>::batch_all { calls }) = call.is_sub_type() {
+			let first_call = &calls[0];
+			let currency = match first_call.is_sub_type() {
+				Some(Call::set_currency { currency }) => *currency,
+				_ => Pallet::<T>::account_currency(who),
+			};
+			currency
+		} else {
+			Pallet::<T>::account_currency(who)
 		};
```
mrq1911, line 409: this can be simplified
```suggestion
let currency = if let Some(Call::set_currency { currency }) = call.is_sub_type() {
    *currency
} else {
    let calls = match call {
        Call::batch { calls } | Call::batch_all { calls } => calls,
        _ => return Pallet::<T>::account_currency(who),
    };
    let first_call = &calls[0];
    match first_call.is_sub_type() {
        Some(Call::set_currency { currency }) => *currency,
        _ => Pallet::<T>::account_currency(who),
    }
};
```

mrq1911: you can even add force batch to it easily

jak-pan (approved):

PR #794, opened 2024-03-22
feat: apply 10% discount on HDX fee

dmoka:
## Checklist:

### DIFF
--- a/pallets/transaction-multi-payment/src/lib.rs
+++ b/pallets/transaction-multi-payment/src/lib.rs
@@ -494,7 +494,7 @@ where
 		if let Some(paid) = already_withdrawn {
 			// Calculate how much refund we should return
 			let (currency, refund, fee, tip) = match paid {
-				PaymentInfo::Native(mut paid_fee) => {
+				PaymentInfo::Native(paid_fee) => {
 					let corrected_fee = NATIVE_FEE_DISCOUNT_MULTIPLIER.mul_floor(corrected_fee);
 					(
 						T::NativeAssetId::get().into(),
--- a/pallets/transaction-multi-payment/src/lib.rs
+++ b/pallets/transaction-multi-payment/src/lib.rs
@@ -488,7 +488,6 @@ where
 		tip: Self::Balance,
 		already_withdrawn: Self::LiquidityInfo,
 	) -> Result<(), TransactionValidityError> {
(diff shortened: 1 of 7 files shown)


jak-pan:
Don't completely understand where this comes from, but I would not do discount on our base tx fees. If we want to "discount" fees for HDX I'd make other more expensive. But would first confirm we even want this.

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -453,13 +458,20 @@ where
...
 
-		let converted_fee =
+		let mut converted_fee =
 			convert_fee_with_price(fee, price).ok_or(TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
 
+		//Apply 10% discount on HDX
+		if currency == T::NativeAssetId::get() {
+			let fee_without_tip = converted_fee.saturating_sub(tip);
+			let fee_without_tip_discounted = NATIVE_FEE_DISCOUNT_MULTIPLIER.mul_floor(fee_without_tip);
+			converted_fee = fee_without_tip_discounted.saturating_add(tip);
```
enthusiastmartin, line 468: where did the tip go ? 

it seems that it is not withdrawn, and when it comes to correcting it , you add it back to corrected fee. 

although there seems to be a test to ensure issuance invariant - according to the code - i would say, it mints the tip ?! 

i would suggest to investigate bit further.

enthusiastmartin: i would suggest to add a simple test and call only pre_dispatch to see what happens with the tip?!

mrq1911: yeah, we have to make sure the treasury receives tip and that it is not burned @dmoka

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -480,17 +492,23 @@ where
 		tip: Self::Balance,
 		already_withdrawn: Self::LiquidityInfo,
 	) -> Result<(), TransactionValidityError> {
+		//TODO: add issuance tests for evm
```
mrq1911, line 495: todo

mrq1911 (changes requested):

dmoka:
Closing it for now, as we don't want to do this at the moment, we also want keep up the demand for LRNA.

PR #796, opened 2024-04-02
feat: add spot price provider for specified route

dmoka:
Explanation for spot price calculations, including fees.

We calculate spot price based on sell trade as it should yield very similar result like buy, and it also simplifies the implementation

### XYK:
- Since in the trade the `amount_out` is reduced by `fee`, it makes `asset_out` more expensive, so the asset_in/asset_out spot price should be increased. So divide spot-price-without-fee by `1-fee` to reflect correct amount out after the fee deduction

### LBP:
1. When distributed asset is bought, it is very similar to XYK, so the `fee` is deducted from `asset_out`, so divide spot-price-without-fee by `1-fee`
2. When accumulated asset is bought, fee does not change spot price as user receives the whole `amount_out` calculated based on whole `amount_in`. So the result should be the same to `spot_price_without_fee`

### Omnipool
1. When selling LRNA, no protocol fee is involved. Fee is taken from `asset_out`, so we need to increase the asset_in/asset_out spot price with dividing by `1-asset_fee` to reflect correct `amount_out` after the fee deduction
2. When trading any other assets, Both [protocol fee](https://github.com/galacticcouncil/HydraDX-node/blob/8abf856afa839b798c4e39e38af2b2901f55739b/math/src/omnipool/math.rs#L40C6-L40C27) and [asset fee](https://github.com/galacticcouncil/HydraDX-node/blob/8abf856afa839b798c4e39e38af2b2901f55739b/math/src/omnipool/math.rs#L53) reduce the asset_out amount received, both making the asset_in/asset_out spot price higher. So we increase the spot price with dividing by `(1-protocol_fee)*(1-asset_fee)` to reflect correct amount out after the fee deduction

### Stableswap
For stableswap there is no formula for trades involving share assets, so we need to do a trade with small amount.

1. When no share involved in direct trade,  `amount_out` is reduced by `fee` in SELL, making `asset_out` more expensive, so the asset_in/asset_out spot price should be increased. So divide spot-price-without-fee by `1-fee` to reflect correct amount out after the fee deduction
2. When share asset is sold, then we execute a small buy with specifying exact stable asset as amount_out, which under the hood leads to `withdraw_asset_amount`. We do this to prevent too low shares in calculations
3. When share asset is bought, then we execute a small sell with specifying exact stable asset as amount_in, which under the hood leads to `remove_liquidity_one_asset`. We do this to prevent too low shares in calculations

TODOs:
- [x] Lower the amounts in the tests so only the slippage can lead to differences
- [x] Rework providers to include fees
- [x] Use math calculation instead of dry-run trade for stableswap spot price
- [x] Add missing spot price providers for all AMMs
- [x] Add functionality to get spot price for given route
- [x] write tests for new lbp spot price provider


mrq1911 (changes requested):

pallets/route-executor/src/lib.rs
```diff
@@ -851,3 +853,31 @@ impl<T: Config> RouteProvider<T::AssetId> for Pallet<T> {
 		}
 	}
 }
+impl<T: Config> RouteSpotPriceProvider<T::AssetId> for Pallet<T> {
+	fn spot_price(route: &[Trade<T::AssetId>]) -> Option<FixedU128> {
```
mrq1911, line 857: it should be always clarified that the spot_price is with fee

pallets/route-executor/src/lib.rs
```diff
@@ -851,3 +853,31 @@ impl<T: Config> RouteProvider<T::AssetId> for Pallet<T> {
...
+			match spot_price_result {
+				Ok(spot_price) => prices.push(spot_price),
+				Err(_) => return None,
+			}
+		}
+		if prices.is_empty() {
+			return None;
+		}
+
+		let nominator = prices.iter().try_fold(U512::from(1u128), |acc, price| {
```
mrq1911, line 871: you can have a bad time when the route is more than 4 hops. you should probably aggregate if after every 4 hops

pallets/stableswap/src/trade_execution.rs
```diff
@@ -197,4 +199,79 @@ impl<T: Config> TradeExecution<T::RuntimeOrigin, T::AccountId, T::AssetId, Balan
...
+					Ok(spot_price_with_fee)
+				} else {
+					let amount_in = T::MinTradingLimit::get();
+
+					let assets = vec![AssetAmount {
+						asset_id: asset_a,
+						amount: amount_in,
+					}];
+
+					let share_amount = Self::calculate_shares(pool_id, &assets)
```
mrq1911, line 265: this case is not calculated by the math

pallets/stableswap/src/trade_execution.rs
```diff
@@ -197,4 +199,79 @@ impl<T: Config> TradeExecution<T::RuntimeOrigin, T::AccountId, T::AssetId, Balan
 			_ => Err(ExecutorError::NotSupported),
 		}
 	}
+
+	fn calculate_spot_price_with_fee(
+		pool_type: PoolType<T::AssetId>,
+		asset_a: T::AssetId,
+		asset_b: T::AssetId,
+	) -> Result<FixedU128, ExecutorError<Self::Error>> {
+		match pool_type {
```
mrq1911, line 208: would be ideal to have this whole method covered by single math function so it doesn't have to be reimplemented in SDK as well

mrq1911 (changes requested):

mrq1911 (approved):

PR #803, opened 2024-04-15
feat: OTC offchain worker

Roznovjak:


traits/src/router.rs
```diff
@@ -3,9 +3,14 @@ use frame_support::dispatch::DispatchResultWithPostInfo;
...
 use frame_support::weights::Weight;
 use scale_info::TypeInfo;
+use sp_arithmetic::FixedU128;
 use sp_std::vec;
 use sp_std::vec::Vec;
 
+pub trait RouteSpotPriceProvider<AssetId> {
+	fn spot_price(route: &[Trade<AssetId>]) -> Option<FixedU128>;
+}
+
```
enthusiastmartin, line 13: not a fan of new price provider trait. 

not sure what is the requirement here - but i believe you can use PriceProvider trait. 

i feel we have some underlying issue due to the fact that we always need to create new price provider traits.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+
+	/// Sort open OTCs orders and save a list in the offchain storage.
+	fn sort_otcs(block_number: BlockNumberFor<T>) {
+		log::debug!(
+			target: "offchain_worker::sort_otcs",
+			"sort_otcs()");
+
+		// acquire offchain worker lock.
+		let lock_expiration = Duration::from_millis(LOCK_DURATION);
+		let mut lock = StorageLock::<'_, Time>::with_deadline(OFFCHAIN_WORKER_LOCK, lock_expiration);
```
mrq1911, line 409: shouldn't you return the lock after sorting finishes? you want to make sure the array is not sorted again sooner than in 5 sec? if you want to make sure the otcs are sorted only once per block, you should cache it based on block height & hash - same block should yield the same order right?

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+	/// Try to find the correct amount to close the arbitrage opportunity.
+	fn try_find_trade_amount(otc_id: OrderId, route: Vec<Trade<AssetIdOf<T>>>) -> Option<Balance> {
+		let otc = <pallet_otc::Orders<T>>::get(otc_id).unwrap();
+
+		// use binary search to determine the correct sell amount
+		let mut sell_amt = otc.amount_in; // start by trying to fill the whole order
+		let mut sell_amt_up = sell_amt;
+		let mut sell_amt_down = 0; // TODO: set to some min trade amount
+
+		let iters = if !otc.partially_fillable { 1 } else { NUM_OF_ITERATIONS };
```
mrq1911, line 496: please flip the condition so there is no negation

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+pub const NAMED_RESERVE_ID: NamedReserveIdentifier = *b"otcorder";
+
+// value taken from https://github.com/substrate-developer-hub/recipes/blob/master/pallets/ocw-demo/src/lib.rs
+pub const UNSIGNED_TXS_PRIORITY: u64 = 100;
+
+pub const OFFCHAIN_WORKER_DATA: &[u8] = b"hydradx/otc-settlements/data/";
+pub const OFFCHAIN_WORKER_DATA_LAST_UPDATE: &[u8] = b"hydradx/otc-settlements/data-last-update/";
+pub const OFFCHAIN_WORKER_LOCK: &[u8] = b"hydradx/otc-settlements/lock/";
+pub const OFFCHAIN_WORKER_MAX_ITERATIONS: &[u8] = b"hydradx/otc-settlements/max-iterations/";
+pub const LOCK_DURATION: u64 = 5_000; // 5 seconds
```
mrq1911, line 69: these constants needs to be more descriptive, especially when they are used only in one place

mrq1911: something like SORTED_ORDERS_LOCK and FILL_SEARCH_ITERATIONS would be better

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+
+// Re-export pallet items so that they can be accessed from the crate namespace.
+pub use pallet::*;
+
+pub type Balance = u128;
+pub type NamedReserveIdentifier = [u8; 8];
+
+pub const PALLET_ID: PalletId = PalletId(*b"otcsettl");
+
+pub const NAMED_RESERVE_ID: NamedReserveIdentifier = *b"otcorder";
```
mrq1911, line 60: this is not used. also it is already defined in otc pallet

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+
+		/// Weight information for the extrinsics.
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn offchain_worker(block_number: BlockNumberFor<T>) {
+			// limit the cases when the offchain worker run
+			if sp_io::offchain::is_validator() {
```
mrq1911, line 122: why you limit it tho? i assume only validators (collators) can actually settle it with unsigned tx?

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+			trade_amount_out: Balance,
+		},
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Otc order not found
+		OrderNotFound,
+		/// Execution requirements not met
+		InvalidConditions,
```
mrq1911, line 179: this is way very vague of error, what condition was not met?

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+}
+
+impl<T: Config> Pallet<T> {
+	pub fn account_id() -> T::AccountId {
+		PALLET_ID.into_account_truncating()
+	}
+
+	/// Ensure that the profit is more than some minimum amount.
+	fn ensure_min_profit(asset: T::AssetId, _profit: Balance) -> DispatchResult {
+		let _min_amount = <T as Config>::ExistentialDeposits::get(&asset)
```
mrq1911, line 234: this approach doesn't really work for insufficient assets as their existential deposit is 1

mrq1911 (changes requested):

mrq1911 (changes requested):
test it with insufficient assets as well

pallets/otc-settlements/src/lib.rs
Roznovjak: additional spam prevention, if you run it as a non validator.

pallets/otc-settlements/src/lib.rs
Roznovjak: You are right, thanks. I started from a copy of the otc pallet.

pallets/otc-settlements/src/lib.rs
Roznovjak: The lock guard should be dropped at the end of `sort_otcs()` and the lock is unlocked. Also the list is updated every time there is a newer block (`try_update_last_block_storage()`).

mrq1911 (approved):

mrq1911:
merge #796 1st

mrq1911 (changes requested):
merge https://github.com/galacticcouncil/HydraDX-node/pull/796 1st

pallets/otc-settlements/src/lib.rs
Roznovjak: We get EDs from ED provider, this was supposed to handle cases when ED == 1 (insufficient assets)

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,536 @@
...
+					otc_price.saturating_sub(router_price_after)
+				} else {
+					router_price_after.saturating_sub(otc_price)
+				}
+			};
+
+			let price_precision = T::PricePrecision::get()
+				.checked_mul(&otc_price)
+				.ok_or(ArithmeticError::Overflow)?;
+			if price_diff > price_precision {
```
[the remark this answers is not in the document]
Roznovjak: it's ok as it is

pallets/otc-settlements/README.md
```diff
@@ -0,0 +1,16 @@
+# OTC Settlements pallet
+## Description
+The pallet provides implementation of the offchain worker for closing existing arbitrage opportunities between OTC 
+orders and the Omnipool.
+Two main parts of this pallet are methods to find the correct amount in order to close an existing arbitrage opportunity 
+and an extrinsic. The extrinsic is mainly called by the offchain worker as unsigned extrinsic, but can by also called
```
dmoka, line 6:
```suggestion
and an extrinsic. The extrinsic is mainly called by the offchain worker as unsigned extrinsic, but can be also called 
```

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+		ensure!(_profit >= _min_expected_profit, Error::<T>::TradeAmountTooLow);
+
+		Ok(())
+	}
+
+	/// Executes two trades: asset_a -> OTC -> asset_b, and asset_b -> Router -> asset_a.
+	///
+	/// If the OTC order is partially fillable, the extrinsic fails if the existing arbitrage
+	/// opportunity is not closed after the trade.
+	/// If the OTC order is not partially fillable, fails if there is no profit after the trade.
```
jak-pan, line 252: I guess it should also fail in this case with partially fillable

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+impl<T: Config> Pallet<T> {
+	pub fn account_id() -> T::AccountId {
+		PALLET_ID.into_account_truncating()
+	}
+
+	/// Ensure that the profit is more than some minimum amount.
+	fn ensure_min_profit(otc_amount_in: Balance, _profit: Balance) -> DispatchResult {
+		let _min_expected_profit = T::MinProfitPercentage::get().mul_floor(otc_amount_in);
+
+		// In the benchmark we doesn't make any trade, so this check would fail.
```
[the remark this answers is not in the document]
jak-pan:
```suggestion
		// We use benchmark of the router.sell (or any other swap mechanism that will be used) to get worst case scenario instead of trying to do this here. We disable this check because we don't do swap here.
```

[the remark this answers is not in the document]
jak-pan: Okay so pls update the comment to make it more precise.

pallets/otc-settlements/src/lib.rs
Roznovjak: Yes, it also fails for partially fillable orders if there is no profit. I will update the comment.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + pallet_otc::Config + SendTransactionTypes<Call<Self>> {
+		/// The overarching event type.
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+		/// Named reservable multi currency.
+		type Currency: MultiCurrency<Self::AccountId, CurrencyId = AssetIdOf<Self>, Balance = Balance>;
```
enthusiastmartin, line 93: i would rather depends on frame suppor traits such Transfer,Inspect - whatever you need. instead of orml.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		#[pallet::constant]
+		type ProfitReceiver: Get<Self::AccountId>;
+
+		/// Minimum profit in terms of percentage.
+		#[pallet::constant]
+		type MinProfitPercentage: Get<Perbill>;
+
+		/// Determines when we consider an arbitrage as closed.
+		#[pallet::constant]
+		type PricePrecision: Get<FixedU128>;
```
enthusiastmartin, line 110: is this ok with various decimals ?

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+	fn ensure_min_profit(otc_amount_in: Balance, _profit: Balance) -> DispatchResult {
+		let _min_expected_profit = T::MinProfitPercentage::get().mul_floor(otc_amount_in);
+
+		// In the benchmark we calculate the overhead of extrinsic and we doesn't make any trade.
+		// We disable this check because otherwise it would fail.
+		#[cfg(not(feature = "runtime-benchmarks"))]
+		// tell the binary search algorithm to find higher values
+		ensure!(_profit >= _min_expected_profit, Error::<T>::TradeAmountTooLow);
+
+		Ok(())
```
enthusiastmartin, line 246: nitpick: i believe what you can do to make it more readable
```
                // In the benchmark we calculate the overhead of extrinsic and we doesn't make any trade.
		// We disable this check because otherwise it would fail.
                if cfg!(feature = "runtime-benchmarks") {
                    return Ok(());
                }
                
		let min_expected_profit = T::MinProfitPercentage::get().mul_floor(otc_amount_in);
		// tell the binary search algorithm to find higher values
		ensure!(profit >= min_expected_profit, Error::<T>::TradeAmountTooLow);
		Ok(())
```

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+
+		// get initial account balances
+		let asset_a_balance_before = <T as Config>::Currency::free_balance(asset_a, &pallet_acc);
+		let asset_b_balance_before = <T as Config>::Currency::free_balance(asset_b, &pallet_acc);
+
+		<T as Config>::Currency::deposit(asset_a, &pallet_acc, amount)?;
+
+		if !otc.partially_fillable {
+			ensure!(otc.amount_out == amount, Error::<T>::NotPartiallyFillable);
+		}
```
enthusiastmartin, line 268: you can move this above, so no need to query the free and balances and make deposit.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+
+		ensure!(
+			route
+				== T::Router::get_route(AssetPair {
+					asset_in: asset_b,
+					asset_out: asset_a,
+				}),
+			Error::<T>::InvalidRoute
+		);
```
enthusiastmartin, line 277: same here. you can do this first and then query balance and make deposit

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		} else {
+			log::debug!(
+			target: "offchain_worker::settle_otc",
+				"calling fill order");
+			pallet_otc::Pallet::<T>::fill_order(RawOrigin::Signed(pallet_acc.clone()).into(), otc_id)?;
+		};
+
+		let otc_amount_out = <T as Config>::Currency::free_balance(asset_b, &pallet_acc)
+			.checked_sub(asset_b_balance_before)
+			.unwrap();
```
enthusiastmartin, line 300: why not use saturaing_sub?

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+			route.clone(),
+		)
+		.map_err(|_| Error::<T>::TradeAmountTooHigh)?;
+		// There are 3 possible types of error:
+		// min trade limit not reached - we start with the largest possible amount, so we can't increase it more.
+		// max trade limit reached - we are interested in this one. We can decrease the amount and try again.
+		// some other error - we can't handle this one properly.
+
+		// // Compare OTC and Router price
+		let router_price_after = T::Router::spot_price_with_fee(&route).unwrap();
```
enthusiastmartin, line 323: why not to return an error instead of unwraping ?

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+				ensure!(router_price_after <= otc_price, Error::<T>::TradeAmountTooHigh);
+				ensure!(router_price_after >= otc_price, Error::<T>::TradeAmountTooLow);
+			}
+		}
+
+		let asset_a_balance_after_router_trade = <T as Config>::Currency::free_balance(asset_a, &pallet_acc);
+
+		let profit = asset_a_balance_after_router_trade
+			.checked_sub(amount)
+			.ok_or(ArithmeticError::Overflow)?;
```
enthusiastmartin, line 354: nitpick: underflow.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		let pallet_acc = Self::account_id();
+
+		let otc = <pallet_otc::Orders<T>>::get(otc_id).ok_or(Error::<T>::OrderNotFound)?;
+		let (asset_a, asset_b) = (otc.asset_in, otc.asset_out);
+
+		// get initial account balances
+		let asset_a_balance_before = <T as Config>::Currency::free_balance(asset_a, &pallet_acc);
+		let asset_b_balance_before = <T as Config>::Currency::free_balance(asset_b, &pallet_acc);
+
+		<T as Config>::Currency::deposit(asset_a, &pallet_acc, amount)?;
```
enthusiastmartin, line 264: just to understand: why we need to deposit and withdraw?

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+			if let Ok(_guard) = lock.try_lock() {
+				let sorted_otcs = StorageValueRef::persistent(OFFCHAIN_WORKER_DATA);
+
+				let mut list = vec![];
+				for (otc_id, otc) in <pallet_otc::Orders<T>>::iter() {
+					let otc_price = FixedU128::checked_from_rational(otc.amount_out, otc.amount_in);
+
+					let route = T::Router::get_route(AssetPair {
+						asset_in: otc.asset_out,
+						asset_out: otc.asset_in,
```
enthusiastmartin, line 419: this is confusing. i would suggest to add a comment why out is in and in is out.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+			.get::<Vec<SortedOtcsStorageType>>()
+			.unwrap_or_default()
+			.unwrap_or_default();
+
+		for (otc_id, otc_price, router_price_before, _price_diff) in sorted_otcs.iter() {
+			log::debug!(
+			target: "offchain_worker::settle_otcs",
+				"test OTC id {:?} ", otc_id);
+
+			if router_price_before > otc_price {
```
enthusiastmartin, line 455: just curious - if you ignoring this case, why put it in the list at all ? 

on line 426, where you add it to this list-  you can already remove this one.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+					amount: sell_amt,
+					route,
+				};
+				let _ = SubmitTransaction::<T, Call<T>>::submit_unsigned_transaction(call.into());
+			}
+		}
+	}
+
+	/// Try to find the correct amount to close the arbitrage opportunity.
+	fn try_find_trade_amount(otc_id: OrderId, route: Vec<Trade<AssetIdOf<T>>>) -> Option<Balance> {
```
enthusiastmartin, line 486: you can use slice instead of vec, than there is not need to copy the route where you use this function.

you can also pass the otc by reference, since you already have it , so no need to query storage again inside this function

pallets/otc-settlements/src/lib.rs
Roznovjak: so you don't need to have an asset needed for OTC swap. At the beginning of the TX, we borrow you some asset so you can close arbitrage opportunity. At the end of the execution you have more than what we borrowed you, so we withdraw the amount that was given to you and the rest goes to the treasury.

pallets/otc-settlements/src/lib.rs
Roznovjak: Thanks. Updated.

pallets/otc-settlements/src/lib.rs
Roznovjak: It's ok. We use this param together with `Price` type.

pallets/otc-settlements/src/lib.rs
enthusiastmartin: you can have integer underflow. it is used eg in balances pallet too. but fine with me.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,568 @@
...
+				ensure!(router_price_after <= otc_price, Error::<T>::TradeAmountTooHigh);
+				ensure!(router_price_after >= otc_price, Error::<T>::TradeAmountTooLow);
+			}
+		}
+
+		let asset_a_balance_after_router_trade = <T as Config>::Currency::balance(asset_a, &pallet_acc);
+
+		let profit = asset_a_balance_after_router_trade
+			.checked_sub(amount)
+			.ok_or(ArithmeticError::Overflow)?;
```
enthusiastmartin, line 363: i think this is incorrect. 

if there was something already in the account prior to calling settle otc - it would be included in the profit.

Roznovjak: Thanks. Fixed.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+	/// Ensure that the profit is more than some minimum amount.
+	fn ensure_min_profit(otc_amount_in: Balance, profit: Balance) -> DispatchResult {
+		// In the benchmark we calculate the overhead of extrinsic and we doesn't make any trade.
+		// We disable this check because otherwise it would fail.
+		if cfg!(feature = "runtime-benchmarks") {
+			return Ok(());
+		}
+
+		let min_expected_profit = T::MinProfitPercentage::get().mul_floor(otc_amount_in);
+		// tell the binary search algorithm to find higher values
```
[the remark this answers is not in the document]
Roznovjak: I will update the comment. The comment is in the correct place. If the `ensure` fails, we return the `TradeAmountTooLow` error which tells the algorithm to search for higher value.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+		let (asset_a, asset_b) = (otc.asset_in, otc.asset_out);
+
+		if !otc.partially_fillable {
+			ensure!(otc.amount_out == amount, Error::<T>::NotPartiallyFillable);
+		}
+
+		ensure!(
+			route
+				== T::Router::get_route(AssetPair {
+					asset_in: asset_b,
```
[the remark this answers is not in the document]
Roznovjak: thanks. I will add a comment.

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+		///
+		#[pallet::call_index(0)]
+		#[pallet::weight(<T as Config>::WeightInfo::settle_otc_order()
+			.saturating_add(<T as Config>::RouterWeightInfo::sell_weight(route))
+		.saturating_add(<T as Config>::RouterWeightInfo::get_route_weight())
+		.saturating_add(<T as Config>::RouterWeightInfo::calculate_spot_price_with_fee_weight(route))
+		.saturating_add(<T as pallet_otc::Config>::WeightInfo::fill_order().max(<T as pallet_otc::Config>::WeightInfo::partial_fill_order()))
+		)]
+		pub fn settle_otc_order(
+			_origin: OriginFor<T>,
```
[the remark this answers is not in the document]
Roznovjak: Yes it has to be there.

dmoka (approved):

pallets/otc-settlements/src/lib.rs
```diff
@@ -0,0 +1,591 @@
...
+	/// opportunity is not closed after the trade and if there is no profit after the trade.
+	/// If the OTC order is not partially fillable, fails only if there is no profit after the trade.
+	pub fn settle_otc(otc_id: OrderId, amount: Balance, route: Vec<Trade<AssetIdOf<T>>>) -> DispatchResult {
+		let pallet_acc = Self::account_id();
+
+		let otc = <pallet_otc::Orders<T>>::get(otc_id).ok_or(Error::<T>::OrderNotFound)?;
+		let (asset_a, asset_b) = (otc.asset_in, otc.asset_out);
+
+		if !otc.partially_fillable {
+			ensure!(otc.amount_out == amount, Error::<T>::NotPartiallyFillable);
```
enthusiastmartin, line 293: i believe there is an issue here.

you compare amount with asset out  ( asset_b) but when profit is calculated - you substrasct the same amount from asset_a

enthusiastmartin (changes requested):

pallets/otc-settlements/src/lib.rs
Roznovjak: There should be `amount_in` in the comparison. Fixed. Thanks

mrq1911 (approved):

PR #819, opened 2024-05-16
fix: claim LM rewards when less than ED

dmoka:
Solves this https://github.com/galacticcouncil/HydraDX-node/issues/776

- [x] rebenchmark both omni and xyk LM


pallets/liquidity-mining/src/lib.rs
```diff
@@ -1182,7 +1189,19 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 							farm_entry.updated_at = current_period;
 
 							let pot = Self::pot_account_id().ok_or(Error::<T, I>::ErrorGetAccountId)?;
-							T::MultiCurrency::transfer(global_farm.reward_currency, &pot, &who, rewards)?;
+							let ed = T::AssetRegistry::existential_deposit(global_farm.reward_currency)
+								.ok_or(Error::<T, I>::NoExistentialDepositForAsset)?;
```
martinfridrich, line 1193: This should be `InconsistentState` error - you can't have asset without ED. `T::AssetRegistry::existential_deposit` returns `None` only if asset is not registered and something is very, very wrong if this happen

integration-tests/src/omnipool_liquidity_mining.rs
```diff
@@ -538,6 +538,202 @@ fn withdraw_shares_should_work_when_deposit_exists() {
 	});
 }
 
+#[test]
+fn withdraw_shares_should_should_return_reward_to_user_when_bigger_than_ed() {
```
martinfridrich, line 542: Typo in the  name `...should_should...` also I would use "send" instead of "return" but  that's detail .
I don't think this test is correct if I get the name correctly.
`Charlie` already have HDX balance so this test passes without ED lowering on line 547.

dmoka: fixed typos.

Yeah it was indeed not correct the thing was that in transfer the ED for HDX is retrieved from pallet balances, not from asset registry, so this test was indeed incorrect.

Inreased the yield reward percentage, not it should be fine.

PR #822, opened 2024-05-20
feat: add omnipool liquidity limits

dmoka:
Solves this: https://github.com/galacticcouncil/HydraDX-node/issues/808


pallets/omnipool/src/lib.rs
```diff
@@ -423,6 +423,10 @@ pub mod pallet {
 		ZeroAmountOut,
 		/// Existential deposit of asset is not available.
 		ExistentialDepositNotAvailable,
+		/// Liquidity limit is not reached while removing liquidity
+		LiquidityLimitNotReached,
+		/// Shares limit is not reached while adding liquidity
+		SharesLimitNotReached,
```
enthusiastmartin, line 429: I would just create one error called `SlippageLimit`. 

i believe there is no real need to have more detailed error message.  we have similar thing in stableswap, dca has something similar too. it would be good to be consistent accross amms and similar things.

enthusiastmartin (changes requested):

enthusiastmartin (changes requested):
To avoid code duplication, i'd suggest to call the new version of add_ and remove_ liquidity from the original add_liquidity and remove_liquidity with min limits.

pallets/omnipool/src/lib.rs
```diff
@@ -426,6 +426,8 @@ pub mod pallet {
 		ZeroAmountOut,
 		/// Existential deposit of asset is not available.
 		ExistentialDepositNotAvailable,
+		/// Slippage protection
+		SlippageLimitReached,
```
enthusiastmartin, line 430: Just called SlippageLimit.  The `reached` does not make much sense.

pallets/omnipool/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-omnipool"
-version = "4.2.2"
+version = "4.2.3"
```
enthusiastmartin, line 3: actually would do 4.3 version. since we are adding new api.

enthusiastmartin (changes requested):

enthusiastmartin (approved):

PR #834, opened 2024-06-03
fix: disable ED charging and refund for insufficient assets

dmoka:
Solves this: https://github.com/galacticcouncil/HydraDX-node/issues/820

- [x] rebenchmark router, update behchmark, then rebenchmark DCA with new weights
- [ ] @mrq1911 find out the the XCM issue that where we should also not charge ED 
- [x] considering moving the SkipEd storage to asset registry to have a more generic solution
- [x] rebenchmark router and orml-tokens once everything resolved


martinfridrich:
orml-tokens needs be rebenchmarked also, you introduced new storage ops into each transfer

pallets/route-executor/src/lib.rs
```diff
@@ -575,6 +612,45 @@ impl<T: Config> Pallet<T> {
...
+				|| !T::InspectRegistry::is_sufficient(trade.asset_out))
+		{
+			*skip_ed_disabling = true;
+			//We optimize to set the state for middle trades only once at the first middle trade, then we change no state till the last trade
+			match trade_index {
+				0 => SkipEd::<T>::put(SkipEdState::SkipEdLock),
+				trade_index if trade_index.saturating_add(1) == route_length => {
+					SkipEd::<T>::put(SkipEdState::SkipEdUnlock)
+				}
+				1 => SkipEd::<T>::put(SkipEdState::SkipEdLockAndUnlock),
```
mrq1911, line 632: shouldn't this be a default, not just 1 for all the cases where you are in the middle of the route?

pallets/route-executor/src/lib.rs
```diff
@@ -54,6 +54,13 @@ pub use pallet::*;
 
 pub const MAX_NUMBER_OF_TRADES: u32 = 5;
 
+#[derive(Debug, Encode, Decode, Copy, Clone, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+pub enum SkipEdState {
+	SkipEdLock,
```
mrq1911, line 59: i think you can remove SkipEd prefix from the enum names

pallets/route-executor/src/weights.rs
```diff
@@ -18,10 +18,10 @@
 
 //! Autogenerated weights for `pallet_route_executor`
 //!
-//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 32.0.0
-//! DATE: 2024-05-24, STEPS: `10`, REPEAT: `30`, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
+//! DATE: 2024-06-05, STEPS: `10`, REPEAT: `30`, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! WORST CASE MAP SIZE: `1000000`
-//! HOSTNAME: `bench-bot`, CPU: `Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz`
+//! HOSTNAME: `dmoka-msi-pc`, CPU: `AMD Ryzen 9 5900X 12-Core Processor`
```
mrq1911, line 24: the benchmark should be ran on reference machine

pallets/route-executor/src/lib.rs
```diff
@@ -575,6 +612,45 @@ impl<T: Config> Pallet<T> {
...
+		route_length: usize,
+		trade_index: usize,
+		trade: Trade<T::AssetId>,
+	) {
+		if route_length > 1
+			&& (!T::InspectRegistry::is_sufficient(trade.asset_in)
+				|| !T::InspectRegistry::is_sufficient(trade.asset_out))
+		{
+			*skip_ed_disabling = true;
+			//We optimize to set the state for middle trades only once at the first middle trade, then we change no state till the last trade
```
mrq1911, line 626: does this really do anything? i do not see any performance gain, just complexity increase

pallets/route-executor/src/lib.rs
```diff
@@ -318,6 +351,10 @@ pub mod pallet {
 				)?;
 			}
 
+			if skip_ed_disabling {
+				SkipEd::<T>::kill();
```
mrq1911, line 355: you should make sure the SkipEd storage is always killed after the last hop is processed, so there is no impact on the on-chain storage

mrq1911 (changes requested):

pallets/route-executor/src/lib.rs
martinfridrich: We would have to set another state for routes with only 1 trade if this would be default storage value.
This state is necessary only when you are in the middle of the route

pallets/route-executor/src/lib.rs
mrq1911: but there really shoudln't be any additional op as this item is never stored on chain, it is just used in memory during route processing

mrq1911: i would rather annotate it somehow so the bechmark knows instead of trying to confuse it with more complex code

martinfridrich: ok, we can change it.
Yes, I know the real write won't happen but you have to do everything else storage vise and I read it's worth to not to do.

dmoka: all right, we will only add this overhead weight if insufficient asset is involved, already fixed.

dmoka: given the fact that the storage value is the cheapest to modify and here we talk about just one write, I don't see much value in complicating the already complex benchmark of route executor.

dmoka:
> orml-tokens needs be rebenchmarked also, you introduced new storage ops into each transfer

done!

pallets/route-executor/src/lib.rs
martinfridrich: I think only change in benchmarks is that we need trigger this branch of code by using insufficient asset. But this needs to be done anyway

I would say this will be always counted as 1 read and 1 write in benchmarks as:
```
* Multiple reads of the same value count as one read.
* Multiple writes of the same value count as one write.
* Multiple reads of the same value, followed by a write to that value, count as one read and one write.
* A write followed by a read only counts as one write.
```

My concern with performance was/is with storage overlay access as described here: 
https://substrate.stackexchange.com/a/2625

If we don't think complexity is worth it, we can do storage write on every iteration

dmoka: Yes, what I am doing is that I add the overhead in case of insufficient asset: 

https://github.com/galacticcouncil/hydration-node/blob/2396e44f731d42fcf9af2a4204ecd9d2a5001115/runtime/hydradx/src/assets.rs#L973-L978

As for killing the storage, it always incorporated. As the reading/writing to a storage value is the cheapest (comparing to other storage types), I think we are fine with the current solution. Let me know otherwise.

integration-tests/src/router.rs
```diff
@@ -973,6 +974,530 @@ mod omnipool_router_tests {
...
+				//ED should be refunded to alice as she sold all her asset, minus the 10% extra
+				assert_eq!(alice_balance_before_trade, alice_balance_after_trade + extra_ed_charge);
+
+				TransactionOutcome::Commit(DispatchResult::Ok(()))
+			});
+		});
+	}
+
+	#[test]
+	fn should_no_ed_charging_disabled_when_only_one_route_with_insufficient_assets() {
```
martinfridrich, line 1294: I wouldn't know what this test is about just by this name

integration-tests/src/router.rs
```diff
@@ -973,6 +974,530 @@ mod omnipool_router_tests {
...
+					hydradx_runtime::RuntimeOrigin::signed(ALICE.into()),
+					insufficient_asset_1,
+					insufficient_asset_2,
+					amount_to_sell,
+					0,
+					trades
+				),);
+
+				//ED for insufficient_asset_1 is refunded, but ED for insufficient_asset_2 is charged plus extra 10%
+				assert_balance!(ALICE.into(), HDX, 1000 * UNITS - 1 * ed - extra_ed_charge);
```
martinfridrich, line 1366:
```suggestion
				assert_balance!(ALICE.into(), HDX, 1000 * UNITS - ed - extra_ed_charge);
```

integration-tests/src/router.rs
```diff
@@ -973,6 +974,530 @@ mod omnipool_router_tests {
...
+					Trade {
+						pool: PoolType::XYK,
+						asset_in: insufficient_asset3,
+						asset_out: insufficient_asset4,
+					},
+				];
+
+				assert_ok!(Currencies::deposit(insufficient_asset1, &ALICE.into(), 1500 * UNITS,));
+
+				let ed = UNITS * 11 / 10;
```
martinfridrich, line 1481: I would use `InsufficientEDinHDX::get()` everywhere instead of `let ed = UNITS * 11 / 10` so we don't have to change tests if will change calculation

pallets/route-executor/src/lib.rs
martinfridrich: I think @mrq1911 ment:
`SkipEdState {
	Lock,
	LockAndUnlock,
	Unlock,
}`

"ed" is still redundant in the usage of this enum
I would seggest:
`
SkipED {
        Lock,
	LockAndUnlock,
	Unlock,
}
`
"state"/"handling" in name doesn't add much in my opinion

pallets/route-executor/src/lib.rs
mrq1911: we really have to use storage to have some sort of context stored during the execution, mby we could use some other way to share context without touching storage layer, but i am not aware of any currently. that would be good way to optimize this.

pallets/route-executor/src/weights.rs
mrq1911: why was this resolved?

dmoka: ah yeah, I just updated the one that is used in production so at location `runtime/hydrardx/src/weights`

Anyway, I will update this with the new weights too.

integration-tests/src/router.rs
dmoka: ah yeah, fixed!

`ed_charging_should_not_be_disabled_when_only_one_trade_with_insufficient_assets`

martinfridrich (approved):

mrq1911 (approved):

PR #835, opened 2024-06-06
refactor: use genesis builder

Roznovjak:
Fixes: #813. This PR removes deprecated method `from_genesis` and replaces it with `builder`.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx"
-version = "13.0.1"
+version = "13.1.0"
 description = "HydraDX node"
 authors = ["GalacticCouncil"]
 edition = "2021"
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx"
-version = "13.0.0"
+version = "13.0.1"
 description = "HydraDX node"
 authors = ["GalacticCouncil"]
 edition = "2021"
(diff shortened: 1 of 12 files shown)


node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx"
-version = "13.0.0"
+version = "13.0.1"
```
dmoka, line 3: very small thing, but maybe it can be a minor version bump, and not a patch.

node/src/chain_spec/mod.rs
```diff
@@ -115,108 +113,115 @@ pub fn parachain_genesis(
...
-	RuntimeGenesisConfig {
-		system: SystemConfig {
-			// Add Wasm runtime to storage.
-			..Default::default()
-		},
-		session: SessionConfig {
-			keys: initial_authorities
-				.0
+) -> serde_json::Value {
+	serde_json::json!({
```
dmoka, line 117: Could we write any meaningful test for this, so next time when we change it we have some safety net? Or what ensures that it still works? Or you tested it manually?

dmoka (approved):
Added 2 comment, nothing critical.

node/src/chain_spec/mod.rs
Roznovjak: ChainSpec consists of 3 parts: client spac, genesis config and wasm code. `from_genesis()` returns `RuntimeGenesisConfig` type and I could't find a way how to use in tests( it's not printable, comparable,...). When I tried to work with `ChainSpec` type I had issues with the wasm, it's not build deterministically so it can produce different results every build. So we can't just compare the new one with the old one. Also the structure is a little bit different. But I tried to at least manually compare everything except wasm code and the content is the same.

PR #836, opened 2024-06-07
feat: opengov staking

enthusiastmartin:
Preparation of staking pallet for opengov. 

Once opengov is introduced, do:

- [x] Add Integration tests
- [x] update benchmarks in conviction-voting pallet to include worst case
- [x] migration of existing votes  ( clearing existing staking votes )


pallets/staking/src/migration.rs
```diff
@@ -0,0 +1,76 @@
...
+
+			let weight = T::DbWeight::get().reads_writes(1, 1);
+			let ev = PositionVotes::<T>::clear(u32::MAX, None);
+			assert!(ev.maybe_cursor.is_none(), "PositionVotes storage is not empty");
+
+			let pv = ProcessedVotes::<T>::clear(u32::MAX, None);
+			assert!(pv.maybe_cursor.is_none(), "ProcessedVotes storage is not empty");
+
+			// Clear the storage
+			weight
```
martinfridrich, line 57: I don't think this weight it is correct. `clear` is doing delete for each item up to `limit`.

pallets/staking/src/integrations/conviction_voting.rs
```diff
@@ -0,0 +1,235 @@
...
+					PositionVotes::<T>::mutate(position_id, |voting| {
+						voting.votes.remove(vote_idx);
+					});
+				}
+			}
+			Ok(())
+		});
+	}
+
+	fn remove_vote_locks_if_needed(who: &T::AccountId, ref_index: ReferendumIndex) -> Option<Balance> {
```
martinfridrich, line 131: This function doesn't do what it says in the name, I would rename it.

martinfridrich (approved):

PR #852, opened 2024-06-27
feat: support assets with xyk dot pools as fee asset

dmoka:
This PR is about:
- using any non-fee payment assets* as fee in DCA https://github.com/galacticcouncil/HydraDX-node/issues/844
- being able to use non-fee payment assets as general fee payment currency

*A non-fee payment asset is any asset whose price is not calculated within the multi-payment pallet. For example we also want to support WUD which is a sufficient asset but not tracked in multi pallets.

TODO:
- [x] adjust benchmark
- [x] rebench on reference machine
- [x] maual test
- [x] check if dca works with buynig insuff, so HDX should be payed, check itnegration test about it
- [x] add support for insufficient general fee payment currency
- [x] refactor so the pallets don't depend on XYK and AssetReg
- [x]  rebenchmark
- [x]  manual test
- [x] modify the check and if check if asset is feee asset, if not  then we check for the XYK pool
- [x] adjust  `ExtrinsicBaseWeight`
- [x] rebenchmark `withdraw_fee` (used for `ExtrinsicBaseWeight`) on reference machine
- [x] manual test the fee value

(Benchmark withdraw_fee and check if it is not bigger than `ExtrinsicBaseWeight`. If so, then adjust. "The base weight contains all the shit for validating tx and doing all the on initialize and other stuff for every transaction so we should probably increase it anyway... But there should be a guide on how to bench this from shawn somewhere I think")


traits/src/fee.rs
```diff
@@ -0,0 +1,28 @@
...
+		dest: &AccountId,
+		from: AssetId,
+		into: AssetId,
+		amount: Amount,
+		max_limit: Amount,
+	) -> DispatchResult;
+
+	fn pool_trade_fee(swap_amount: Amount) -> Result<Amount, DispatchError>;
+
+	fn get_amount_in_for_out(
```
enthusiastmartin: so it is consistent with other naming. 
```suggestion
	fn calculate_in_given_out(
```

traits/src/fee.rs
```diff
@@ -0,0 +1,28 @@
...
+	fn buy(
+		origin: &AccountId,
+		dest: &AccountId,
+		from: AssetId,
+		into: AssetId,
+		amount: Amount,
+		max_limit: Amount,
+	) -> DispatchResult;
+
+	fn pool_trade_fee(swap_amount: Amount) -> Result<Amount, DispatchError>;
```
enthusiastmartin, line 21: why result is amount ?! should be Permill ? or more generic PerThing ?

dmoka: Here I am calculating the pool trade fee which is the ExchangeFeeRate multiplied by an amount.

We could return only the ExchangeFeeRate, but then the fee calculation needs to happen where this function is used, but that would require xyk specific code, such as:

[https://github.com/galacticcouncil/hydration-node/blob/6cab9f7e235c7a050e8c30163f006e565225a641/runtime/hydradx/src/assets.rs#L1755-L1760](https://github.com/galacticcouncil/hydration-node/blob/6cab9f7e235c7a050e8c30163f006e565225a641/runtime/hydradx/src/assets.rs#L1755-L1760)

enthusiastmartin: i would rename it to calculate fee amount.

pallets/dca/src/lib.rs
```diff
@@ -1102,6 +1155,19 @@ impl<T: Config> Pallet<T> {
 		Ok(amount)
 	}
 
+	fn get_fee_in_dot(fee_amount_in_native: Balance) -> Result<Balance, DispatchError> {
```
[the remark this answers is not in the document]
dmoka: All right, the renaming is reasonable, but let's keep the parameter name `fee_amount_in_native` to prevent future confusion. This will help that only native amounts are used in the future when someone will call this function.

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -598,6 +619,8 @@ where
 	<T as frame_system::Config>::RuntimeCall: IsSubType<Call<T>> + IsSubType<pallet_utility::pallet::Call<T>>,
 	<T as pallet_utility::Config>::RuntimeCall: IsSubType<Call<T>>,
 	BalanceOf<T>: FixedPointOperand,
+	BalanceOf<T>: From<MC::Balance>,
+	BalanceOf<T>: From<u128>,
```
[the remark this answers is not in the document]
dmoka: good catch, removed!

pallets/dca/src/lib.rs
```diff
@@ -943,14 +950,36 @@ impl<T: Config> Pallet<T> {
...
+			//The DOT we need to buy is calculated the same way how we convert weight to insufficient fee
+			let pool_trade_fee = T::SwappablePaymentAssetSupport::calculate_fee_amount(fee_amount_in_sold_asset)?;
+
+			//Since there is a trade fee involved in xyk buy swap, we need to unallocate that, together with amount_in
+			let effective_amount_in = fee_amount_in_sold_asset
+				.checked_add(pool_trade_fee)
+				.ok_or(ArithmeticError::Overflow)?;
+			Self::unallocate_amount(schedule_id, schedule, effective_amount_in)?;
+
+			let fee_in_dot = Self::convert_to_polkadot_native_asset(Self::weight_to_fee(weight_to_charge))?;
```
enthusiastmartin, line 973: do you need to do this again ? 

should not this be equal to fee_amount_in_sold_asset?

pallets/dca/src/lib.rs
```diff
@@ -1092,16 +1139,36 @@ impl<T: Config> Pallet<T> {
...
 		let amount = if asset_id == T::NativeAssetId::get() {
 			asset_amount
-		} else {
+		} else if T::SwappablePaymentAssetSupport::is_transaction_fee_currency(asset_id) {
 			let price = T::NativePriceOracle::price(asset_id).ok_or(Error::<T>::CalculatingPriceError)?;
 
 			multiply_by_rational_with_rounding(asset_amount, price.n, price.d, Rounding::Up)
 				.ok_or(ArithmeticError::Overflow)?
+		} else {
+			let fee_amount_in_dot = Self::convert_to_polkadot_native_asset(asset_amount)?;
```
enthusiastmartin, line 1148: this does not seem right. 

it tries to convert asset_amount to DOT, by calling convert_to_polkadot.  and that functions uses hdx/dot price. but asset amount here is not HDX. 
or am i missing something ?

pallets/dca/src/lib.rs
dmoka: `fee_amount_in_sold_asset` is different from `fee_in_dot`.

Because the `amount_in` (so DCA fee asset) is not necessarly DOT.

But we need to calculate the `fee_in_dot`, so we know how much we should do a buy trade for, within the FeeAsset-DOT xyk pool

pallets/dca/src/lib.rs
dmoka: maybe the variable naming could be improved, so `asset_amount` could be renamed to `native_amount`

The name of the function `convert_native_amount_to_currency`, so this function is supposed to be used for converting native amount (so HDX amount) to some currency.

enthusiastmartin: ah ok. i haven't noticed that it already converts native amount

enthusiastmartin (approved):

PR #853, opened 2024-07-02
feat: enable metadata hash verification

Roznovjak:


bkchr:
@Roznovjak how are you building the runtime? Why do you think that it did not set the env variable?

Roznovjak:
> @Roznovjak how are you building the runtime? Why do you think that it did not set the env variable?

I'm building it with `cargo build --release -p hydradx-runtime --features=metadata-hash`.
After the build, `echo $RUNTIME_METADATA_HASH` is empty.

bkchr:
> After the build, `echo $RUNTIME_METADATA_HASH` is empty.

The env variable is not only set for the `wasm-builder` in the build process. It will not be set afterwards.

Roznovjak:
> > After the build, `echo $RUNTIME_METADATA_HASH` is empty.
> 
> The env variable is not only set for the `wasm-builder` in the build process. It will not be set afterwards.

Ah. Thanks.

mrq1911 (approved):

PR #863, opened 2024-07-10
feat: OTC fees

Roznovjak:


pallets/otc/src/lib.rs
```diff
@@ -291,14 +305,17 @@ pub mod pallet {
 			let who = ensure_signed(origin)?;
 			let order = <Orders<T>>::get(order_id).ok_or(Error::<T>::OrderNotFound)?;
 
-			Self::execute_order(&order, &who, order.amount_in, order.amount_out)?;
+			let fee = Self::calculate_fee(order.amount_out);
```
enthusiastmartin, line 308: why not to have the fee calculation inside the execute_order ?  you would not need extra parameter and there would be only one place to calculate it.

i get you need to have the fee amount in the event, but the execute order can return the amount.

Roznovjak: Initially it was there, but I didn't like returning `fee` from `execute_order()`. It doesn't make a lot of sense. But I can change it

enthusiastmartin: or you can move the deposit event into execute order as well.

Roznovjak: sounds good, thanks

Roznovjak: in the latest version, we need fee in some check prior calling execute_order, so I will leave it as it is

vgantchev (approved):
lgtm

enthusiastmartin (approved):

Issue #883, opened 2024-07-29
Add sell all functions to router and AMMs


jak-pan:
Similar to https://github.com/open-web3-stack/open-runtime-module-library/blob/23c8363d53fe8580f636a7f820c54b994d1b96f7/tokens/src/lib.rs#L522 this would allow us to create simpler UX and pass through functionality with XCM programs allowing better composability in similar sense that buy_for works for fee payment.

dmoka:
We start with sell_all for router which solves the Max button issue.

PR #888, opened 2024-08-07
feat: add sell_all for route executor

dmoka:
Fixes https://github.com/galacticcouncil/hydration-node/issues/883

- [x] test out on chopsticks


mrq1911 (changes requested):
benchmarks

mrq1911 (approved):
bechmark is not necessary as there is no additional read

jak-pan (approved):
I assume the native balance test is also a fee payment asset and in this case LGTM

dmoka:
> I assume the native balance test is also a fee payment asset and in this case LGTM

@jak-pan  Although it is true that the fee payment asset is indeed HDX in the native balance test, it does not have any impact as in integration tests the tx fees are not deducted directly. Anyway, I tested out sell_all on chopsticks, and working fine, first the fee is deducated then the rest is sold in `sell_all`

dmoka:
https://github.com/galacticcouncil/hydration-ui/pull/1042

PR #891, opened 2024-08-15
feat: custom conviction voting pallet for opengov

enthusiastmartin:
This PR uses conviction voting pallet from forked sdk repository rather than ac opy in hydration repo.


enthusiastmartin:
no longer relevant

PR #907, opened 2024-09-15
feat: ICE

enthusiastmartin:
Initial ICE implementation.

TODO:

- [ ] docs, comments, readme
- [ ] benchmark
- [x] tests implementation (stubs)
- [ ] add math test for bying lrna
- [x] solution weight calculation
- [x] improve solution validation
- [ ] dont allow `submit_solution` to be tipped
- [ ] dont allow `submit_solution` in batch
- [x] slash proposer in case of invalid solution
- [ ] router address allowed to buy lrna ( needed to set route) - it should not be allowed to.
- [ ] on_success/on_failure hooks
- [x] solver using convex optimization


jak-pan:
@enthusiastmartin can you pls make build work in CI so guys can get a binary?

pallets/ice/src/tests/engine.rs
```diff
@@ -0,0 +1,146 @@
+use super::*;
+use crate::engine::ICEEngine;
```
jak-pan, line 2: Just a nitpick... Intent composing engine engine ? 😄

enthusiastmartin:
ICE parked.

PR #910, opened 2024-09-20
feat: unified trade events

### LINKED ISSUE #755: Unify all trading events into the one generic one
```
Swapped { 
    who: AccountId, 
    pool: PoolType, 
    operation: ( Buy | Sell ), 
    asset_in: AssetId, 
    asset_out: AssetId, 
    amount_in: Balance, 
    amount_out: Balance,
    fee_rate: Permill
    batch: Option<u256> 
}
```

batch is nonce that router is setting on start of execution and resetting to `None` after 

add all trading events in the Omnipool, Stableswap, XYK & OTC with this one
depreciate all the old trading related fees (will be removed in future)

Roznovjak:
Fixes #755

Additional todos:
- [x] Create galactic council branch from batch 3, create PR.
- [x] Create a patch 4
- [x] Create PR for master https://github.com/paritytech/polkadot-sdk/pull/6517
- [x] integration-tests like MM
- [x] organize effort for manual testing

TODO:
- [x] First merge https://github.com/galacticcouncil/polkadot-sdk/pull/9
- [x] Use the galactic council polkadot sdk patch in dependnecie
- [x] Add stack management to utils pallet
- [x] rebench utility pallet
- [x] Use stack in deposit trade event.


pallets/lbp/src/lib.rs
```diff
@@ -1111,6 +1127,18 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, BalanceOf<T>> for Pallet<T
 			fee_amount: transfer.fee.1,
 		});
 
+		pallet_trade_event::Pallet::<T>::deposit_trade_event(
+			transfer.origin.clone(),
```
jak-pan, line 1133: I am thinking to add second account because of the Intents... We might have 2 accounts trading directly. cc @enthusiastmartin wdyt?

jak-pan: Also easier to track when if have multiple pools with same assets

pallets/lbp/src/lib.rs
```diff
@@ -1244,6 +1273,19 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, BalanceOf<T>> for Pallet<T
...
+			pallet_trade_event::PoolType::LBP,
+			pallet_trade_event::TradeOperation::Buy,
+			transfer.assets.asset_in,
+			transfer.assets.asset_out,
+			transfer.amount,
+			transfer.amount_b,
+			vec![transfer.fee],
+			None,
```
jak-pan, line 1288: batch id?

pallets/omnipool/src/lib.rs
```diff
@@ -359,6 +363,12 @@ pub mod pallet {
 
 		/// Asset's weight cap has been updated.
 		AssetWeightCapUpdated { asset_id: T::AssetId, cap: Permill },
+
+		/// Amount of the Hub asset has been updated.
+		HubAmountUpdated {
+			hub_amount_in: Balance,
+			hub_amount_out: Balance,
```
jak-pan, line 370: batch-id? (I would probably rename it to event-id or something similar

pallets/omnipool/src/lib.rs
```diff
@@ -1031,185 +1041,7 @@ pub mod pallet {
...
-			});
-
-			#[cfg(feature = "try-runtime")]
-			Self::ensure_trade_invariant(
-				(asset_in, asset_in_state, new_asset_in_state),
-				(asset_out, asset_out_state, new_asset_out_state),
-			);
-
-			Ok(())
+			Self::do_sell(origin, asset_in, asset_out, amount, min_buy_amount, None)
```
jak-pan, line 1045: batch id?

pallets/lbp/src/lib.rs
enthusiastmartin: not sure if the second account is good idea.. it is just confusing to me - what does the second account mean really ?  then you would need to amounts too. 

if you want to capture the matched trade in ICE - why not just fire 2 trade events for each account. The source of this wont be any AMM, but something else - eg  ICED.

enthusiastmartin:
what if we call the new pallet something more generic ... eg. amm-support. 

it might be useful for other things, not only one specific trade event.

pallets/trade-event/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+	#[pallet::error]
+	pub enum Error<T> {}
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Trade executed.
+		Swapped {
+			who: T::AccountId,
+			pool: PoolType<AssetId>,
```
enthusiastmartin, line 59: after discussion with @jak-pan , we suggest:

```suggestion
			swapper: T::AccountId,  
			filler: T::AccountId, // ICE acc, Omnipool acc, stablepool acc, xyk account etc..
			type: PoolType<AssetId>,
```

traits/src/router.rs
```diff
@@ -60,12 +60,19 @@ pub trait RouteProvider<AssetId> {
 	}
 }
 
+#[derive(Encode, Decode, Clone, Copy, Debug, Eq, PartialEq, TypeInfo, MaxEncodedLen)]
+pub enum TradeOperation {
+	Sell,
+	Buy,
```
enthusiastmartin, line 66:
```suggestion
	ExactIn,
        ExactOut,
```

pallets/trade-event/src/lib.rs
Roznovjak: Who is `filler` in OTC?

Roznovjak: `type` is a keyword, cannot be used

jak-pan: > Who is `filler` in OTC?

the one filling order

traits/src/router.rs
dmoka: What is the reason behind the suggestion? Sell/Buy sounds like better operation names.

ExactIn and ExactOut are always known implicit knowledge of Sell and Buys repsectively.

traits/src/router.rs
```diff
@@ -167,6 +183,7 @@ pub trait TradeExecution<Origin, AccountId, AssetId, Balance> {
 		asset_out: AssetId,
 		amount_in: Balance,
 		min_limit: Balance,
+		event_id: Option<IncrementalId>,
```
dmoka, line 186: this can be removed, we don't need it anymore

pallets/omnipool/src/lib.rs
dmoka: why does this event `HubAmountUpdated` need the batch/event id?

pallets/trade-event/src/lib.rs
dmoka: isn't it the one who owns the order?

Cargo.toml
```diff
@@ -3,47 +3,48 @@ panic = 'unwind'
...
-    'pallets/asset-registry',
-    'runtime/adapters',
-    'pallets/collator-rewards',
-    'pallets/transaction-pause',
-    'pallets/ema-oracle',
-    'pallets/liquidity-mining',
-    'pallets/currencies',
+  'node',
+  'runtime/hydradx',
+  'pallets/omnipool',
```
jak-pan, line 8: Was this formatted wrongly b4?

integration-tests/src/referrals.rs
jak-pan, line 1: I am missing swapped events (and probably fees in the event shown correctly?) when somebody uses referral code.

integration-tests/src/router.rs
```diff
@@ -152,13 +153,70 @@ mod router_different_pools_tests {
...
+						20000000,
+						LBP::pool_data(LBP::get_pair_id(pallet_lbp::types::AssetPair::new(DAI, LRNA)))
+							.unwrap()
+							.fee_collector,
+					)],
+					operation_id: vec![ExecutionType::Router(0)],
+				}
+				.into(),
+				pallet_amm_support::Event::Swapped {
+					swapper: BOB.into(),
```
jak-pan, line 175: I guess we'll have 2 here. Also this taste is buying LRNA from LBP which is confusing as you can't do that.

jak-pan: We're also minssing all of the fees here.

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+	pub trait Config: frame_system::Config {
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+	}
+
+	#[pallet::storage]
+	/// Next available incremental ID
+	#[pallet::getter(fn incremental_id)]
+	pub(super) type IncrementalId<T: Config> = StorageValue<_, IncrementalIdType, ValueQuery>;
+
+	// TODO:
```
jak-pan, line 82: TODO?

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+	#[pallet::getter(fn id_stack)]
+	pub(super) type IdStack<T: Config> = StorageValue<_, ExecutionIdStack, ValueQuery>;
+
+	#[pallet::error]
+	pub enum Error<T> {
+		MaxStackSizeReached,
+		EmptyStack,
+	}
+
+	// on initialize - clear operation_id stack if not empty
```
jak-pan, line 94: Why the events have comments above the macros and storages have docs in between?

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+	/// Returns next incremental ID and updates the storage.
+	pub fn next_incremental_id() -> Result<IncrementalIdType, DispatchError> {
+		IncrementalId::<T>::try_mutate(|current_id| -> Result<IncrementalIdType, DispatchError> {
+			let inc_id = *current_id;
+			*current_id = current_id.checked_add(1).ok_or(ArithmeticError::Overflow)?;
+			Ok(inc_id)
+		})
+	}
+
+	#[allow(clippy::too_many_arguments)]
```
jak-pan, line 125: Just set a global setting

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+		});
+	}
+}
+
+impl<T: Config> ExecutionTypeStack<IncrementalIdType> for Pallet<T> {
+	fn push(execution_type: ExecutionType<IncrementalIdType>) -> DispatchResult {
+		IdStack::<T>::try_mutate(|stack| -> DispatchResult {
+			stack
+				.push(execution_type)
+				.map_err(|_| Error::<T>::MaxStackSizeReached.into())
```
jak-pan, line 153: Shouldn't we cap this somehow?

pallets/circuit-breaker/src/tests/mock.rs
```diff
@@ -86,6 +86,7 @@ frame_support::construct_runtime!(
 		Omnipool: pallet_omnipool,
 		Tokens: orml_tokens,
 		CircuitBreaker: pallet_circuit_breaker,
+		AmmSupport: pallet_amm_support,
```
jak-pan, line 89: This should probably be renamed to just support or swap or chain support or smth as it's not just used by amms

jak-pan (changes requested):
Sirs pls cross check this spec with our design of swapped event it should be compatible to map this https://www.erc7683.org/spec (i.e. amount of inputs, swapper, filler etc.) 
I will cross-check it later again.

What do we do with XCM Exchange or XCM Execute? I guess we should also map that (Also through batches and stuff)

Cargo.toml
dmoka: i think yes, it was 8 instead of 4 spaces

dmoka: i think yes, it should be 4 spaces instead of 8. Or am i using some wrong fmt config?

integration-tests/src/referrals.rs
dmoka: added swapped event assert in test `trading_in_omnipool_should_use_asset_rewards_when_set`, pls double check

integration-tests/src/router.rs
dmoka: yes, fixed, pls double check

pallets/amm-support/src/lib.rs
dmoka: as no relation to any task here, I remove it.

pallets/amm-support/src/lib.rs
dmoka: i guess this gotta be a todo comment. Added on init where clearing the stack.

pallets/amm-support/src/lib.rs
dmoka: We have it capped to 10, by using this const: https://github.com/galacticcouncil/hydration-node/blob/39a2d9a478368b7dd90146b7d387789ac6a5424a/pallets/amm-support/src/lib.rs#L43

traits/src/router.rs
```diff
@@ -68,6 +105,31 @@ pub enum PoolType<AssetId> {
 	Omnipool,
 }
 
+#[derive(Encode, Decode, Clone, Copy, Debug, Eq, PartialEq, TypeInfo, MaxEncodedLen)]
+pub enum AssetType<AssetId, NFTId> {
+	Fungible(AssetId),
+	NFT(NFTId),
```
enthusiastmartin, line 111: why not call it explicitly - Nonfungible ?

pallets/omnipool/src/lib.rs
```diff
@@ -232,6 +235,9 @@ pub mod pallet {
 
 		/// Oracle price provider. Provides price for given asset. Used in remove liquidity to support calculation of dynamic withdrawal fee.
 		type ExternalPriceOracle: ExternalPriceProvider<Self::AssetId, EmaPrice, Error = DispatchError>;
+
+		/// Operation id provider for unified events
+		type AmmUnifiedEventSupport: IncrementalIdProvider<IncrementalIdType> + ExecutionTypeStack<IncrementalIdType>;
```
enthusiastmartin, line 240: this does not make much sense.

you added direct dependency on pallet_amm_support - yet you require to provide something that implements these traits. and that something is set again to AmmSupportPallet in the runtime.

pallets/amm-support/src/lib.rs
enthusiastmartin: note : it can be done a much easier way ( let's discuss on call )

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+
+	fn get() -> Vec<ExecutionType<IncrementalIdType>> {
+		IdStack::<T>::get().get()
+	}
+
+	fn clear() {
+		IdStack::<T>::kill();
+	}
+}
+impl<T: Config> IncrementalIdProvider<IncrementalIdType> for Pallet<T> {
```
enthusiastmartin, line 182: Nope.

Note: note for myself.

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+	#[pallet::storage]
+	/// Next available incremental ID
+	#[pallet::getter(fn incremental_id)]
+	pub(super) type IncrementalId<T: Config> = StorageValue<_, IncrementalIdType, ValueQuery>;
+
+	#[pallet::storage]
+	/// Next available incremental ID
+	#[pallet::getter(fn id_stack)]
+	pub(super) type IdStack<T: Config> = StorageValue<_, ExecutionIdStack, ValueQuery>;
+
```
enthusiastmartin, line 87: explanation would be cool.. as it is on clear based on naming

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+		},
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+			let mut weight: Weight = Weight::zero();
+			weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
+
+			<Self as ExecutionTypeStack<IncrementalIdType>>::clear();
```
enthusiastmartin, line 117: nope. 

note for myself.

runtime/hydradx/src/system.rs
```diff
@@ -305,9 +310,32 @@ impl pallet_utility::Config for Runtime {
...
+	type BatchPostHook = PopBatchExecutionTypeForUnifiedEvent;
 	type WeightInfo = weights::pallet_utility::HydraWeight<Runtime>;
 }
 
+pub struct PushBatchExecutionTypeForUnifiedEvent;
+
+impl BatchPreHook for PushBatchExecutionTypeForUnifiedEvent {
+	fn on_batch_start() -> DispatchResult {
+		let next_event_id = AmmSupport::next_id().map_err(|_| ArithmeticError::Overflow)?;
+		AmmSupport::push(ExecutionType::Batch(next_event_id))?;
```
enthusiastmartin, line 323: here we can clearly see why I added my previous "nope" comment ( where it implements the same interface for Pallet).

first point:
If you try to use interface of AmmSupport - you will get these options next_id(), next_increment_id(), get() (some id)  and probabyl few others as well ( from the storage gettters).

since there seem to be 2 types of ids afaik - it becomes very messy and unclear what to actually use.

second point - why the next_id and push has to be outside the support pallet - it is a logic that should be hidden inside the pallet imo.

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {}
+}
+
+impl<T: Config> Pallet<T> {
+	/// Returns next incremental ID and updates the storage.
+	pub fn next_incremental_id() -> Result<IncrementalIdType, DispatchError> {
+		IncrementalId::<T>::try_mutate(|current_id| -> Result<IncrementalIdType, DispatchError> {
+			let inc_id = *current_id;
+			*current_id = current_id.checked_add(1).ok_or(ArithmeticError::Overflow)?;
```
enthusiastmartin, line 132: use overflowing. so when reaches the end - it starts from 0 rather than make all calls failing due to this.

traits/src/router.rs
dmoka: since we removed, it is not applicable anymore

integration-tests/src/cross_chain_transfer.rs
```diff
@@ -110,7 +110,7 @@ fn rococo_should_receive_asset_when_sent_from_hydra() {
 	Rococo::execute_with(|| {
 		assert_eq!(
 			hydradx_runtime::Balances::free_balance(AccountId::from(BOB)),
-			2_999_989_698_923 // 3 * HDX - fee
+			2_999_989_606_492 // 3 * HDX - fee
```
jak-pan, line 113: Why did the fee change?

pallets/circuit-breaker/src/tests/mock.rs
jak-pan: What did we decide?

integration-tests/src/dca.rs
```diff
@@ -588,6 +682,106 @@ mod omnipool {
...
+						operation_id: vec![
+							ExecutionType::DCA(schedule_id, 0),
+							ExecutionType::Router(1),
+							ExecutionType::Omnipool(2)
+						],
+					}),
+					RuntimeEvent::AmmSupport(pallet_amm_support::Event::Swapped {
+						swapper: ALICE.into(),
+						filler: Omnipool::protocol_account(),
+						filler_type: pallet_amm_support::types::Filler::Omnipool,
```
jak-pan, line 731: Would be nice if this had more complex route since it's integration test but I guess we can do that separately.

integration-tests/src/dca.rs
```diff
@@ -201,6 +202,99 @@ mod omnipool {
...
+				last_two_swapped_events,
+				vec![
+					RuntimeEvent::AmmSupport(pallet_amm_support::Event::Swapped {
+						swapper: ALICE.into(),
+						filler: Omnipool::protocol_account(),
+						filler_type: pallet_amm_support::types::Filler::Omnipool,
+						operation: pallet_amm_support::types::TradeOperation::ExactOut,
+						inputs: vec![Asset::new(HDX, 140421094367051)],
+						outputs: vec![Asset::new(LRNA, 70210545436437)],
+						fees: vec![Fee::new(LRNA, 35105272718, Omnipool::protocol_account()),],
```
jak-pan, line 234: What happens with referrals, should we see it here?

jak-pan: (Just OT question, doesn't fit context)

integration-tests/src/dca.rs
```diff
@@ -201,6 +202,99 @@ mod omnipool {
...
+						operation_id: vec![
```
jak-pan, line 235: I am inclined to rename this to something in line of `operation_stack` does it ring more bells?

integration-tests/src/exchange_asset.rs
```diff
@@ -120,6 +124,86 @@ fn hydra_should_swap_assets_when_receiving_from_acala_with_sell() {
...
+					swapper: TempAccountForXcmAssetExchange::get(),
+					filler: Omnipool::protocol_account(),
+					filler_type: pallet_amm_support::types::Filler::Omnipool,
+					operation: pallet_amm_support::types::TradeOperation::ExactIn,
+					inputs: vec![pallet_amm_support::types::Asset::new(ACA, 50000000000000),],
+					outputs: vec![pallet_amm_support::types::Asset::new(LRNA::get(), 49180327868852)],
+					fees: vec![Fee::new(LRNA::get(), 24590163934, Omnipool::protocol_account()),],
+					operation_id: vec![
+						ExecutionType::Xcm(topic_id, 0),
+						ExecutionType::XcmExchange(1),
```
jak-pan, line 147: No topic here?

integration-tests/src/exchange_asset.rs
```diff
@@ -120,6 +124,86 @@ fn hydra_should_swap_assets_when_receiving_from_acala_with_sell() {
...
+		let last_two_swapped_events = &last_swapped_events[last_swapped_events.len() - 2..];
+		let topic_id = [
+			88, 96, 17, 161, 1, 110, 221, 66, 135, 183, 36, 117, 203, 64, 121, 121, 144, 130, 127, 130, 11, 90, 144,
+			184, 162, 244, 249, 80, 95, 193, 189, 200,
+		];
+		pretty_assertions::assert_eq!(
+			last_two_swapped_events,
+			vec![
+				RuntimeEvent::AmmSupport(pallet_amm_support::Event::Swapped {
+					swapper: TempAccountForXcmAssetExchange::get(),
```
jak-pan, line 138: Can we get original acc or just this temp acc

integration-tests/src/referrals.rs
```diff
@@ -319,6 +325,36 @@ fn trading_in_omnipool_should_use_asset_rewards_when_set() {
...
+				}
+				.into(),
+				pallet_amm_support::Event::Swapped {
+					swapper: BOB.into(),
+					filler: Omnipool::protocol_account(),
+					filler_type: Filler::Omnipool,
+					operation: TradeOperation::ExactIn,
+					inputs: vec![Asset::new(LRNA, 1205165959)],
+					outputs: vec![Asset::new(DAI, 26663424573622008)],
+					fees: vec![Fee::new(DAI, 70524156750724, Omnipool::protocol_account())],
```
jak-pan, line 352: Looks like we're missing referral fees here

integration-tests/src/router.rs
```diff
@@ -2522,16 +3032,35 @@ mod lbp_router_tests {
...
+					amount: 9_980_000_000_000,
+					sale_price: received_amount_out,
+					fee_asset: HDX,
+					fee_amount: 20_000_000_000,
+				}
+				.into(),
+				pallet_amm_support::Event::Swapped {
+					swapper: BOB.into(),
+					filler: LBP::get_pair_id(pallet_lbp::types::AssetPair::new(DAI, HDX)),
+					filler_type: pallet_amm_support::types::Filler::LBP,
```
jak-pan, line 3049: Shouldn't we have shareToken here?

math/src/stableswap/math.rs
```diff
@@ -129,7 +129,7 @@ pub fn calculate_shares<const D: u8>(
 	amplification: Balance,
 	share_issuance: Balance,
 	fee: Permill,
-) -> Option<Balance> {
+) -> Option<(Balance, Vec<Balance>)> {
```
jak-pan, line 132: This looks like breaking change.

jak-pan: UI should be notified updated and ready for this.

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+			fees: Vec<Fee<T::AccountId>>,
+			operation_id: Vec<ExecutionType>,
+		},
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+			let mut weight: Weight = Weight::zero();
+			weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
```
jak-pan, line 92: do we read | save anything?

pallets/amm-support/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+mod tests;
+
+pub mod types;
+
+// Re-export pallet items so that they can be accessed from the crate namespace.
+pub use pallet::*;
+
+const LOG_TARGET: &str = "runtime::amm-support";
+
+pub const MAX_STACK_SIZE: u32 = 10;
```
jak-pan, line 38: What is the max size of the event at 10? Maybe we could increase to 16?

pallets/amm-support/src/lib.rs
jak-pan: Q: Do we need to read or save anything? 

We should have unique block in which something is executed and then execution id can be per block and as such we might wipe it and have 0 reads and writes here? Also not sure to what we configured this but it could be u16 or u32 if we do so

pallets/amm-support/src/types.rs
```diff
@@ -0,0 +1,68 @@
...
+		Self { asset, amount }
+	}
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, Copy, Debug, Eq, PartialEq, TypeInfo, MaxEncodedLen)]
+pub enum ExecutionType {
+	Router(IncrementalIdType),
+	DCA(ScheduleId, IncrementalIdType),
+	ICE(IncrementalIdType),
```
jak-pan, line 54: no ICE yet?

pallets/otc-settlements/src/tests.rs
```diff
@@ -292,11 +294,28 @@ fn existing_arb_opportunity_should_trigger_trade_when_correct_amount_can_be_foun
...
-		}
-		.into()]);
+		expect_events(vec![
+			Event::Executed {
+				asset_id: HDX,
+				profit: 17_736_110_470_326,
+			}
+			.into(),
+			pallet_amm_support::Event::Swapped {
+				swapper: OtcSettlements::account_id(),
```
jak-pan, line 304: This is flipped. You are filling existing swap (order)

pallets/otc/src/lib.rs
```diff
@@ -325,13 +347,29 @@ pub mod pallet {
...
 				order_id,
-				who,
+				who: who.clone(),
 				amount_in: order.amount_in,
 				amount_out: order.amount_out,
 				fee,
 			});
+
+			pallet_amm_support::Pallet::<T>::deposit_trade_event(
+				who,
```
jak-pan, line 360: flipped

integration-tests/src/cross_chain_transfer.rs
dmoka: It is coming from using `WirthUniqueTopic` for Router

integration-tests/src/dca.rs
dmoka: I like this, lemme rename it!

jak-pan (changes requested):
Some nitpicks and some missing stuff (LBP, OTC flipped accounts and possible rename)

integration-tests/src/exchange_asset.rs
dmoka: the API of AssetExchange does the full xcm instructions, so can't have the topic ID easily

```
		_origin: Option<&Location>,
		give: AssetsInHolding,
		want: &Assets,
		maximal: bool,
```

What we could maybe do is to read the topic from our stack storage, as at this point, the `Xcm([u8; 32], IncrementalIdType)` should be there. Not sure if it is always there, otherwise we could do use Optional type. WDYT?

pallets/amm-support/src/lib.rs
dmoka: bounded vec stored in storage somehow works only with u32, experienced this strange behaviour also in other pallets, not sure why...

Bumped to 16 anyway

integration-tests/src/exchange_asset.rs
dmoka: the API of AssetExchange does the full xcm instructions, so can't have the topic ID easily

		_origin: Option<&Location>,
		give: AssetsInHolding,
		want: &Assets,
		maximal: bool,
What we could maybe do is to read the topic from our stack storage, as at this point, the Xcm([u8; 32], IncrementalIdType) should be there. Not sure if it is always there, otherwise we could do use Optional type. WDYT?

integration-tests/src/exchange_asset.rs
dmoka: unfortunately only the temp acc as the actual trade happens by that account (on behalf of user), that's how we could implement asset exchange back then.

pallets/amm-support/src/lib.rs
dmoka: as discussed, we leave it as it is for now, then after testing it out by indexer on the test evn, we could do this optimizastion.

pallets/omnipool/src/lib.rs
```diff
@@ -2017,18 +2125,40 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// Calls `on_trade_fee` hook and ensures that no more than the fee amount is transferred.
-	fn process_trade_fee(trader: &T::AccountId, asset: T::AssetId, amount: Balance) -> DispatchResult {
```
dmoka, line 2020: test

integration-tests/src/router.rs
dmoka: there is no shareToken in LBP. Source: https://github.com/galacticcouncil/hydration-node/blob/1be939a7a4022585a9fa37042361282ae1c89028/pallets/lbp/src/lib.rs#L945

pallets/omnipool/src/lib.rs
```diff
@@ -2017,18 +2124,48 @@ impl<T: Config> Pallet<T> {
...
+		amount: Balance,
+	) -> Result<Vec<Fee<T::AccountId>>, DispatchError> {
 		let account = Self::protocol_account();
 		let original_asset_reserve = T::Currency::free_balance(asset, &account);
 
 		// Let's give little bit less to process. Subtracting one due to potential rounding errors
 		let allowed_amount = amount.saturating_sub(Balance::one());
-		let used = T::OmnipoolHooks::on_trade_fee(account.clone(), trader.clone(), asset, allowed_amount)?;
+		let fees = T::OmnipoolHooks::on_trade_fee(account.clone(), trader.clone(), asset, allowed_amount)?;
+		let additional_fees_total: Balance = fees
```
enthusiastmartin, line 2138: it is not really additional fee. so i would suggest to change naming here. eg "taken_fee"

pallets/omnipool/src/lib.rs
```diff
@@ -2017,18 +2124,48 @@ impl<T: Config> Pallet<T> {
...
+			.clone()
+			.into_iter()
+			.filter_map(|opt| opt.map(|(balance, _)| balance))
```
enthusiastmartin, line 2141: since you filter out zero amounts later on, you can do it directly here

```suggestion
			.filter_map(|opt| opt.map(|(balance, _)| if balance.is_zero(){
			     None
			}
			else{
			    Some(balance)
			}))
```

something like that should work

pallets/omnipool/src/lib.rs
```diff
@@ -2017,18 +2124,48 @@ impl<T: Config> Pallet<T> {
...
+
+		let mut all_trade_fees = vec![Fee {
+			asset: asset.into(),
+			amount: normal_trade_fee,
+			recipient: Self::protocol_account(),
+		}];
+
+		let additional_fee_entries: Vec<Fee<T::AccountId>> = fees
+			.iter()
+			.filter_map(|opt| {
```
enthusiastmartin, line 2159: no need for filter map, just map

pallets/omnipool/src/lib.rs
```diff
@@ -2017,18 +2124,48 @@ impl<T: Config> Pallet<T> {
...
+			.sum();
+
 		let asset_reserve = T::Currency::free_balance(asset, &account);
 		let diff = original_asset_reserve.saturating_sub(asset_reserve);
 		ensure!(diff <= allowed_amount, Error::<T>::FeeOverdraft);
-		ensure!(diff == used, Error::<T>::FeeOverdraft);
-		Ok(())
+		ensure!(diff == additional_fees_total, Error::<T>::FeeOverdraft);
+
+		let normal_trade_fee = amount.saturating_sub(additional_fees_total);
```
enthusiastmartin, line 2149: normal does not say much. i'd suggest "protocol_fee_amount"

pallets/omnipool/src/lib.rs
```diff
@@ -2017,18 +2124,48 @@ impl<T: Config> Pallet<T> {
...
+
+		let mut all_trade_fees = vec![Fee {
+			asset: asset.into(),
+			amount: normal_trade_fee,
+			recipient: Self::protocol_account(),
+		}];
+
+		let additional_fee_entries: Vec<Fee<T::AccountId>> = fees
+			.iter()
```
enthusiastmartin, line 2158: no need to iter over references here. you can directly consume the fees. so you avoid clones and stuff.

but that requires my previous comment to filter out zeors previously.

pallets/omnipool/src/lib.rs
enthusiastmartin: but that should not really return 0 fee amounts anyway.

dmoka: i see your point, but here I sum the total, instead of converting them to some entries, so having the current implementation makes more sense imo.

pallets/omnipool/src/lib.rs
dmoka: since tkaen fee can contain optionals, and I wanna filter them out, filter map seems like the better option

pallets/omnipool/src/lib.rs
dmoka: since your previous comment does not take into consideration that we sum only total amount there, therefore I dont think we need to refactor this further, lmk otherwise.

integration-tests/src/dca.rs
dmoka: with the latest changes, we include both referrals an staking taken fees.

math/src/stableswap/math.rs
dmoka: crated branch for this [fix/fee-changes](https://github.com/galacticcouncil/HydraDX-wasm/tree/fix/fee-changes)

pallets/omnipool/src/lib.rs
```diff
@@ -1089,6 +1094,42 @@ pub mod pallet {
 				protocol_fee_amount: state_changes.fee.protocol_fee,
 			});
 
+			pallet_support::Pallet::<T>::add_to_context(ExecutionType::Omnipool)?;
```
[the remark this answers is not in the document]
dmoka: we have two swapped events emitted in case of a omnipool swap, namely AssetA -> HUB and HUB-> AssetB. In order to connect these two, we need the context.

pallets/support/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+		let next_id = IncrementalId::<T>::try_mutate(|current_id| -> Result<IncrementalIdType, DispatchError> {
+			let inc_id = *current_id;
+			*current_id = current_id.overflowing_add(1).0;
+			Ok(inc_id)
+		})?;
+
+		ExecutionContext::<T>::try_mutate(|stack| -> DispatchResult {
+			//We make it fire and forget, and it should fail only in test and when if wrongly used
+			debug_assert_ne!(stack.len(), MAX_STACK_SIZE as usize, "Stack should not be full");
+			if let Err(err) = stack.try_push(execution_type(next_id)) {
```
Roznovjak, line 137: There might be an issue with this. Imagine a scenario where we push 15 times to the stack. The stack is populated 10 times, and ignored 5 times. Then let's reduce the stack 7 times. Now the stack has 3 items, but it should have 8 items.

pallets/support/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+			filler_type,
+			operation,
+			inputs,
+			outputs,
+			fees,
+			operation_stack,
+		});
+	}
+
+	pub fn add_to_context<F>(execution_type: F) -> Result<IncrementalIdType, DispatchError>
```
Roznovjak, line 124: We want to have these methods infallible so I would change the return type to just `IncrementalIdType`, if possible. But I'm not sure when `try_mutate` can fail. This is just an idea.

pallets/dca/src/lib.rs
```diff
@@ -330,7 +331,9 @@ pub mod pallet {
 			who: T::AccountId,
 			block: BlockNumberFor<T>,
 		},
-		///The DCA trade is successfully executed
+		/// Deprecated. Use pallet_amm::Event::Swapped instead.
+		/// The DCA trade is successfully executed
+		//TODO: remove once we migrated completely to pallet_amm::Event::Swapped
```
jak-pan, line 336:
```suggestion
		// TODO: remove once we migrated completely to pallet_amm::Event::Swapped
```

pallets/lbp/Cargo.toml
```diff
@@ -66,5 +67,6 @@ std = [
     'frame-benchmarking/std',
```
jak-pan, line 67:
```suggestion
    "frame-benchmarking/std",
```

pallets/omnipool/src/lib.rs
```diff
@@ -305,6 +305,8 @@ pub mod pallet {
 			shares_removed: Balance,
 		},
 		/// Sell trade executed.
+		/// Deprecated. Replaced by pallet_broadcast::Swapped
+		//TODO: remove when completely migrated to new Swapped event
```
jak-pan, line 309:
```suggestion
		// TODO: remove when completely migrated to new Swapped event
```

pallets/omnipool/src/lib.rs
```diff
@@ -317,6 +319,8 @@ pub mod pallet {
 			protocol_fee_amount: Balance,
 		},
 		/// Buy trade executed.
+		/// Deprecated. Replaced by pallet_broadcast::Swapped
+		//TODO: remove when completely migrated to new Swapped event
```
jak-pan, line 323:
```suggestion
		// TODO: remove when completely migrated to new Swapped event
```

pallets/omnipool/src/lib.rs
```diff
@@ -1089,6 +1094,42 @@ pub mod pallet {
 				protocol_fee_amount: state_changes.fee.protocol_fee,
 			});
 
+			pallet_broadcast::Pallet::<T>::add_to_context(ExecutionType::Omnipool);
+
+			//Swapped event for AssetA to HubAsset
```
jak-pan, line 1099:
```suggestion
			// Swapped event for AssetA to HubAsset
```

pallets/omnipool/src/lib.rs
```diff
@@ -1089,6 +1094,42 @@ pub mod pallet {
...
+					*state_changes.asset_in.delta_hub_reserve,
+				)],
+				vec![Fee::new(
+					T::HubAssetId::get().into(),
+					state_changes.fee.protocol_fee,
+					Destination::Burned,
+				)],
+			);
+
+			//Swapped event for HubAsset to AssetB
```
jak-pan, line 1117:
```suggestion
			// Swapped event for HubAsset to AssetB
```

pallets/omnipool/src/lib.rs
```diff
@@ -1293,6 +1335,42 @@ pub mod pallet {
...
+					*state_changes.asset_in.delta_hub_reserve,
+				)],
+				vec![Fee::new(
+					T::HubAssetId::get().into(),
+					state_changes.fee.protocol_fee,
+					Destination::Burned,
+				)],
+			);
+
+			//Swapped even from HubAsset to AssetB
```
jak-pan, line 1358:
```suggestion
			// Swapped even from HubAsset to AssetB
```

pallets/omnipool/src/lib.rs
```diff
@@ -1810,6 +1888,20 @@ impl<T: Config> Pallet<T> {
 			protocol_fee_amount: state_changes.fee.protocol_fee,
 		});
 
+		//No protocol fee in case of selling hub asset
```
jak-pan, line 1891:
```suggestion
		// No protocol fee in case of selling hub asset
```

pallets/omnipool/src/lib.rs
```diff
@@ -1917,6 +2010,20 @@ impl<T: Config> Pallet<T> {
 			protocol_fee_amount: state_changes.fee.protocol_fee,
 		});
 
+		//No protocol fee in case of buying asset for hub asset
```
jak-pan, line 2013:
```suggestion
		// No protocol fee in case of buying asset for hub asset
```

pallets/omnipool/src/lib.rs
```diff
@@ -2017,18 +2124,45 @@ impl<T: Config> Pallet<T> {
...
-		let used = T::OmnipoolHooks::on_trade_fee(account.clone(), trader.clone(), asset, allowed_amount)?;
+		let taken_fee = T::OmnipoolHooks::on_trade_fee(account.clone(), trader.clone(), asset, allowed_amount)?;
+		let taken_fee_entries: Vec<Fee<T::AccountId>> = taken_fee
+			.iter()
+			.filter_map(|opt| {
+				opt.clone().map(|(balance, recipient)| {
+					Fee::new(asset.into(), balance, Destination::Account(recipient.clone()))
+				})
+			})
+			.filter(|fee| fee.amount > 0) //filter out when we zero percentage is configured for fees
```
jak-pan, line 2145:
```suggestion
			.filter(|fee| fee.amount > 0) // filter out when we zero percentage is configured for fees
```

pallets/otc/src/tests/mock.rs
```diff
@@ -57,6 +57,7 @@ frame_support::construct_runtime!(
 		 System: frame_system,
 		 OTC: otc,
 		 Tokens: orml_tokens,
+		 Broadcast:pallet_broadcast,
```
jak-pan, line 60:
```suggestion
		 Broadcast: pallet_broadcast,
```

pallets/stableswap/src/lib.rs
```diff
@@ -208,6 +211,8 @@ pub mod pallet {
 			fee: Balance,
 		},
 		/// Sell trade executed. Trade fee paid in asset leaving the pool (already subtracted from amount_out).
+		/// Deprecated. Replaced by pallet_broadcast::Swapped
+		//TODO: remove once we migrated completely to pallet_amm::Event::Swapped
```
jak-pan, line 215:
```suggestion
		// TODO: remove once we migrated completely to pallet_amm::Event::Swapped
```

pallets/lbp/Cargo.toml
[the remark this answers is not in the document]
Roznovjak: The code is already formatted. Maybe we will need to adjust some settings for the formatting.

jak-pan (approved):

PR #914, opened 2024-09-29
chore: lower minTotalFarmRewards for xyk liquidity mining

martinfridrich:
This PR lowers `MinTotalFarmRewards` for xyk liquidity mining.
This is safe to lower as farms' creation needs to go through governance.


martinfridrich:
yes, this is not covered by integration test.

PR #920, opened 2024-10-04
feat: add insufficient fee support for evm

dmoka:
Add missing insufficient fee asset support for EVM transactions/

Since for EVM, we use dispatch permit, that logic needed to be adjusted.

- [x] increase the cost of dispatch permit


traits/src/fee.rs
```diff
@@ -17,6 +17,12 @@ pub trait SwappablePaymentAssetTrader<AccountId, AssetId, Amount>: InspectTransa
 		asset_out_amount: Amount,
 	) -> Result<Amount, DispatchError>;
 
+	fn calculate_out_given_in(
+		asset_in: AssetId,
+		asset_in_amount: Amount,
+		asset_out: AssetId,
```
enthusiastmartin, line 23: keep the order consistent as always.

so asset_in, asset_out, amount

runtime/hydradx/src/evm/evm_fee.rs
```diff
@@ -87,10 +102,52 @@ where
...
+				else {
+					return Err(Error::<T>::WithdrawFailed);
+				};
+
+				let Some(eth_dot_price_as_fixed) = FixedU128::checked_from_rational(eth_dot_price.n, eth_dot_price.d)
+				else {
+					return Err(Error::<T>::WithdrawFailed);
+				};
+
+				let fee_in_dot: Balance = convert_fee_with_price(fee.unique_saturated_into(), eth_dot_price_as_fixed)
```
enthusiastmartin, line 129: why you need to do this ? 

isnt this result from the convert function on line 119 ?

dmoka: good catch, conver already returns this.

mrq1911 (changes requested):
include cost of swapping into dispatch permit

runtime/hydradx/src/evm/evm_fee.rs
```diff
@@ -88,10 +102,45 @@ where
 			return Ok(None);
 		}
 		let account_id = T::AddressMapping::into_account_id(*who);
-		let fee_currency = AC::get(&account_id);
-		let Some((converted, price)) = C::convert((EC::get(), fee_currency, fee.unique_saturated_into())) else {
-			return Err(Error::<T>::WithdrawFailed);
-		};
+		let account_fee_currency = AccountCurrency::get(&account_id);
+
+		let dot = DotAssetId::get();
```
enthusiastmartin, line 107: you dont need to read it here, as it is only used in some cases

traits/src/fee.rs
```diff
@@ -17,6 +17,12 @@ pub trait SwappablePaymentAssetTrader<AccountId, AssetId, Amount>: InspectTransa
 		asset_out_amount: Amount,
 	) -> Result<Amount, DispatchError>;
 
+	fn calculate_out_given_in(
+		asset_in: AssetId,
+		asset_out: AssetId,
+		asset_in_amount: Amount,
```
enthusiastmartin, line 23: nitpick: the generic here should Balance - amount is considered signed ( check primitives::Amount type). 

just to be consistent.

runtime/adapters/src/price.rs
```diff
@@ -55,28 +56,83 @@ where
 		// if this gets removed (eg. Convert returns weight), the constraint on T and ema-oracle is not necessary
 		price_weight.saturating_accrue(pallet_ema_oracle::Pallet::<T>::get_price_weight().saturating_mul(2));
 
+		//2 reads as we are checking if from and to assets are transaction fee currencies
+		let is_transaction_fee_currency_check_weight = T::DbWeight::get().reads(2);
+		price_weight.saturating_accrue(is_transaction_fee_currency_check_weight);
```
enthusiastmartin, line 61: no need to introduce a variable here.

runtime/adapters/src/price.rs
```diff
@@ -55,28 +56,83 @@ where
...
 			return (0, price_weight);
 		};
 		(converted, price_weight)
 	}
 }
 
-pub struct ConvertAmount<P>(sp_std::marker::PhantomData<P>);
+pub struct ConvertAmount<PriceProv, SwappablePaymentAssetSupport, DotAssetId>(
+	sp_std::marker::PhantomData<(PriceProv, SwappablePaymentAssetSupport, DotAssetId)>,
+);
```
enthusiastmartin, line 72: nitpick: this one is one me - it should be really be called "ConvertBalance" ( considering my other nitpick here)

enthusiastmartin (approved):

mrq1911 (approved):

PR #939, opened 2024-10-30
fix: erc20 edge cases error handling

mrq1911:


runtime/hydradx/src/evm/erc20_currency.rs
```diff
@@ -174,11 +174,25 @@ fn decode_integer(value: Vec<u8>) -> Option<U256> {
...
+}
+
 fn handle_result(result: CallResult) -> DispatchResult {
 	let (exit_reason, value) = result;
 	match exit_reason {
-		ExitReason::Succeed(ExitSucceed::Returned) => Ok(()),
-		ExitReason::Succeed(ExitSucceed::Stopped) => Ok(()),
+		ExitReason::Succeed(_) => {
+			if Some(false) == decode_bool(value) {
+				Err(DispatchError::Other("evm: erc20 transfer returned false"))
```
dmoka, line 191: Could we improve this error message a bit, relating to the domain? When can this happen? Maybe we could hint that.

dmoka (approved):

runtime/hydradx/src/evm/erc20_currency.rs
mrq1911: not sure what do you mean. i dont think we can make it more descriptive, this is literally describing this special case when we error out even though transactions succeed.

Frankcastleauditor (approved):

PR #953, opened 2024-11-26
feat: add more lm wrappers

dmoka:
Further dicusiions:
- add positions- in case of empty array, we dont do anything with farms

in case of remove, we need 

Add two more extrinsics:
- add_liquidity_with_limit_and_join_farms
- add_liquidity_stableswap_omnipool_and_join_farms

In order to make UX better on FE


integration-tests/src/omnipool_liquidity_mining.rs
```diff
@@ -1430,3 +1769,345 @@ pub fn expect_reward_claimed_events(e: Vec<RuntimeEvent>) {
...
+		(10000 * UNITS).try_into().unwrap(),
+	));
+
+	assert_ok!(<Currencies as MultiCurrencyExtended<_>>::update_balance(
+		HDX,
+		&account,
+		(10000 * UNITS).try_into().unwrap(),
+	));
+
+	//<Currencies as MultiCurrencyExtended<AccountId>>::update_balance(HDX, &treasury, 500_000_000_000_000_000_000i128)?;
```
jak-pan, line 1895: Remove commented out code

integration-tests/src/omnipool_liquidity_mining.rs
```diff
@@ -1430,3 +1769,345 @@ pub fn expect_reward_claimed_events(e: Vec<RuntimeEvent>) {
...
+			Some(name),
+			AssetKind::Token,
+			Balance::one(),
+			None,
+			None,
+			None,
+			None,
+		))
+	})?;
+	/*let name = b"ETH".to_vec().try_into().map_err(|_| "BoundedConvertionFailed")?;
```
jak-pan, line 1995: Remove commented out code

pallets/omnipool-liquidity-mining/src/benchmarks.rs
jak-pan, line 1: Why is this deleted?

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -996,6 +998,84 @@ pub mod pallet {
...
+		/// - `amount`: amount of the asset to be deposited into the liquidity mining.
+		/// - `min_shares_limit`: The min amount of delta share asset the user should receive in the position
+		///
+		/// Emits `LiquidityAdded` events from both pool
+		/// Emits `SharesDeposited` event for the first farm entry
+		/// Emits `SharesRedeposited` event for each farm entry after the first one
+		///
+		#[pallet::call_index(18)]
+		#[pallet::weight(<T as Config>::WeightInfo::add_liquidity_stableswap_omnipool_and_join_farms(farm_entries.len() as u32))]
+		pub fn add_liquidity_stableswap_omnipool_and_join_farms(
```
jak-pan, line 1057: I'm wondering if this should be here. But can't immediately think about where else would it be.

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -1125,4 +1205,64 @@ impl<T: Config> Pallet<T> {
...
+
+	#[require_transactional]
+	fn do_withdraw_shares(
+		origin: OriginFor<T>,
+		deposit_id: DepositId,
+		yield_farm_id: YieldFarmId,
+	) -> Result<Balance, DispatchError> {
+		let owner = Self::ensure_nft_owner(origin, deposit_id)?;
+
+		//NOTE: not tested - this should never fail.
```
jak-pan, line 1217: What do you mean by this comment? If user provides wrong deposit_id you'll get error. But it looks like it's handled correctly.

integration-tests/src/omnipool_liquidity_mining.rs
```diff
@@ -514,6 +523,336 @@ fn add_liquidity_and_join_farms_should_work_for_multiple_farms() {
...
+			hydradx_runtime::OmnipoolLiquidityMining::add_liquidity_with_limit_and_join_farms(
+				RuntimeOrigin::signed(CHARLIE.into()),
+				farms.try_into().unwrap(),
+				ETH,
+				1_000 * UNITS,
+				1_000 * UNITS + 1
+			),
+			pallet_omnipool::Error::<hydradx_runtime::Runtime>::SlippageLimit
+		);
+	});
```
jak-pan, line 702: We should also check events

runtime/hydradx/src/benchmarking/omnipool_liquidity_mining.rs
jak-pan, line 1: Why was this moved here?

pallets/omnipool-liquidity-mining/src/benchmarks.rs
dmoka: it has been replaced to hydradx/src/benchmarking because we need more dependencies (stableswap) to fulfull the benchmarking

pallets/omnipool-liquidity-mining/src/lib.rs
dmoka: yeah, creating a new pallet would be overkill, I feel this is OK to have it here.

pallets/omnipool-liquidity-mining/src/lib.rs
dmoka: These comments were added by @martinfridrich in several places in the past when the pallet was made. I think it is for explaining the reason for using defensive_ok and returning `InconsistentStateError`.

runtime/hydradx/src/benchmarking/omnipool_liquidity_mining.rs
dmoka: easier to include stableswap dependency. 

We also do the same for xyk_lm benchmark due to the same reason, so by placing it to here we make the code base more consistent too.

integration-tests/src/omnipool_liquidity_mining.rs
dmoka: added, pls double check

pallets/omnipool-liquidity-mining/src/lib.rs
martinfridrich: it's just comment why this error is not teste. It doesn't have to be there as `InconsistentStateError` represents error that can never happen. 
It's probably relic from before I added `InconsistentStateError` errors.

@jak-pan this storage read can't fail. If user provide wrong `deposit_id` it would fail on `ensure_nft_owner`
Fail of this storage read would mean that LM's position exists without underlying omnipool's position and it should never happen. 

@dmoka why did you add this function? I quickly grep-ed it and it looks like it's not used by this pallet and it's not `pub`

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -1007,6 +1009,83 @@ pub mod pallet {
...
+		/// This function allows user to add liquidity to stableswap pool,
+		/// then adding the stable shares as liquidity to omnipool
+		/// then use that omnipool shares to join multiple farms.
+		///
+		/// Parameters:
+		/// - `origin`: owner of the omnipool position to deposit into the liquidity mining.
+		/// - `farm_entries`: list of farms to join.
+		/// - `asset`: id of the asset to be deposited into the liquidity mining.
+		/// - `amount`: amount of the asset to be deposited into the liquidity mining.
+		/// - `min_shares_limit`: The min amount of delta share asset the user should receive in the position
```
martinfridrich, line 1059: This is wrong, fn bellow has different set of params

martinfridrich (commented):
I didn't do full review, just what i saw because I was mentioned

runtime/hydradx/src/benchmarking/omnipool_liquidity_mining.rs
```diff
@@ -0,0 +1,974 @@
...
+		OmnipoolLiquidityMining::redeposit_shares(RawOrigin::Signed(lp1.clone()).into(), 3, 4, deposit_id)?;
+		OmnipoolLiquidityMining::redeposit_shares(RawOrigin::Signed(lp1.clone()).into(), 5, 6, deposit_id)?;
+		OmnipoolLiquidityMining::redeposit_shares(RawOrigin::Signed(lp1.clone()).into(), 7, 8, deposit_id)?;
+		OmnipoolLiquidityMining::redeposit_shares(RawOrigin::Signed(lp1.clone()).into(), 9, 10, deposit_id)?;
+
+		set_period(400);
+
+	}: {
+		//We fire and forget as claim rewards is disabled
+		let _ = OmnipoolLiquidityMining::claim_rewards(RawOrigin::Signed(lp1).into(), deposit_id, 10);
```
martinfridrich, line 613: This is not true anymore, this benchmark should fail if extrinsic failed

pallets/omnipool-liquidity-mining/src/lib.rs
dmoka: as discussed, we keep it here.

pallets/omnipool-liquidity-mining/src/lib.rs
dmoka: hmm you are right, left behind, will remove, thanks for noting!

runtime/hydradx/src/benchmarking/omnipool_liquidity_mining.rs
dmoka: fixed

runtime/hydradx/src/benchmarking/omnipool_liquidity_mining.rs
```diff
@@ -0,0 +1,969 @@
...
+		let blocks_per_period = BlockNumberFor::<crate::Runtime>::from(100_u32);
+		let reward_currency = register_asset(b"REW".to_vec(), 10000 * ONE).map_err(|_| BenchmarkError::Stop("Failed to register asset"))?;
+		let owner = funded_account("owner", 0, &[REWARD_CURRENCY]);
+		let yield_per_period = Perquintill::from_percent(20);
+		let min_deposit = 1_000;
+		let price_adjustment = FixedU128::from(10_u128);
+
+	}: _(RawOrigin::Root,  G_FARM_TOTAL_REWARDS, planned_yielding_periods, blocks_per_period, REWARD_CURRENCY, owner, yield_per_period, min_deposit, FixedU128::one())
+
+	update_global_farm {
```
martinfridrich, line 343: This benchmark is not correct. `update_global_farm` is doing more work when farm is not empty

runtime/hydradx/src/benchmarking/omnipool_liquidity_mining.rs
```diff
@@ -0,0 +1,969 @@
...
+		set_period(100);
+		lm_deposit_shares(lp, global_farm_id, 2, position_id)?;
+
+		set_period(1000);
+
+	}:  {
+		OmnipoolLiquidityMining::create_yield_farm(RawOrigin::Signed(owner).into(), global_farm_id, ETH, FixedU128::one(), Some(LoyaltyCurve::default()))?;
+
+	} verify {
+	}
```
martinfridrich, line 406:
```suggestion
	}:  _(RawOrigin::Signed(owner).into(), global_farm_id, ETH, FixedU128::one(), Some(LoyaltyCurve::default()))
```

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -1007,6 +1011,46 @@ pub mod pallet {
...
+		/// Emits `SharesDeposited` event for the first farm entry
+		/// Emits `SharesRedeposited` event for each farm entry after the first one
+		///
+		#[pallet::call_index(16)]
+		#[pallet::weight(<T as Config>::WeightInfo::add_liquidity_stableswap_omnipool_and_join_farms(farm_entries.len() as u32))]
+		pub fn add_liquidity_stableswap_omnipool_and_join_farms(
+			origin: OriginFor<T>,
+			stable_pool_id: T::AssetId,
+			stable_asset_amounts: BoundedVec<AssetAmount<T::AssetId>, ConstU32<MAX_ASSETS_IN_POOL>>,
+			farm_entries: BoundedVec<(GlobalFarmId, YieldFarmId), T::MaxFarmEntriesPerDeposit>,
```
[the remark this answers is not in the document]
dmoka: We discussed that empty farm would do the job, we also do similar in router namely that when empty route specified, then we use onchain route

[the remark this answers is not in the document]
martinfridrich: yes, I remember discussion(if there wasn't another discussion I wasn't part of) but we discussed functionality, not implementation details, so yes, we discussed what should happen when user won't specify farms. 

I don't see a point about router, this pallet has nothing to do with router and we already have extrinsic(`join_farms`) that doesn't allow empty `farm_entries` list.
We should be consistent within the pallet first and not try to be partially consistent with something unrelated...

martinfridrich (approved):

Issue #954, opened 2024-11-30
Add more usable fee payment estimation runtime API


jak-pan:
It is currently super complicated to calculate fee payment amount as we have bunch of different rules for this. We should create runtime API where you will enter weight + accountId and get correct asset and amount, alternatively provide extrinsic and accountId and get correct values. This should not be hard as we have all the functions in the runtime to do this (when charging the fees) We could even add EVM calls to this to make it more simple.

P.S. these functions will be useful when implementing XCM fee estimation API and payment.

floor-licker:
I'd love to be assigned this issue, could get to work on it right away.

jak-pan:
> I'd love to be assigned this issue, could get to work on it right away.

Let's try

jak-pan:
> I'd love to be assigned this issue, could get to work on it right away.

Let me know if you want to take it and if you need more information.

sylvaincormier:
Hi @jak-pan,  
I’d be happy to take this issue on! 
Could you please assign it to me, 
and let me know if there’s any further detail or guidance you’d recommend before I get started? 
Thanks!

PR #962, opened 2024-12-12
Issue 954 fee estimation api

sylvaincormier:
Hi @jak-pan,

I’ve opened a draft PR that addresses #954 by introducing a runtime API for fee estimation. The API provides two key methods:

    estimate_fee_payment(weight, account_id): Returns the correct asset and fee amount based on the given weight and the account’s configured currency.
    estimate_fee_payment_for_extrinsic(encoded_extrinsic, account_id): Decodes the extrinsic, determines the applicable fees, and then provides the correct asset and amount.

I’ve also included extensive test coverage covering fallback scenarios (no price available), invalid extrinsic inputs, minimal weight fees, large extrinsics near max size, and account currency switching.

I made it a draft PR since there are only a few days left in the Kudos Carnival, and I wanted to ensure you have a chance to review and provide feedback before finalizing. Please let me know if any changes are needed. Thank you!

### DIFF
--- a/primitives/src/runtime_api.rs
+++ b/primitives/src/runtime_api.rs
@@ -1,0 +1,20 @@
+use crate::constants::transaction::MaxExtrinsicSize;
+use codec::Codec;
+use frame_support::BoundedVec;
+use sp_weights::Weight;
+
+sp_api::decl_runtime_apis! {
+	pub trait FeeEstimationApi<AccountId, AssetId, Balance>
+	where
+		AccountId: Codec,
+		AssetId: Codec,
+		Balance: Codec,
+	{
+		fn estimate_fee_payment(weight: Weight, account_id: AccountId) -> (AssetId, Balance);
+
+		fn estimate_fee_payment_for_extrinsic(
+			uxt: BoundedVec<u8, MaxExtrinsicSize>,
+			account_id: AccountId
+		) -> (AssetId, Balance);
+	}
+}
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -65,6 +65,7 @@ use sp_std::{convert::From, prelude::*};
 use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 // A few exports that help ease life for downstream crates.
+
 use frame_support::{construct_runtime, pallet_prelude::Hooks, weights::Weight};
 pub use hex_literal::hex;
 use orml_traits::MultiCurrency;
@@ -78,6 +79,12 @@ pub use primitives::{
 };
 use sp_api::impl_runtime_apis;
 pub use sp_consensus_aura::sr25519::AuthorityId as AuraId;
+use sp_runtime::traits::One;
+
+
+use frame_support::dispatch::GetDispatchInfo;
+use frame_support::BoundedVec;
+use primitives::constants::transaction::MaxExtrinsicSize;
 
 /// Opaque types. These are used by the CLI to instantiate machinery that don't need to know
 /// the specifics of the runtime. They can then be made to be agnostic over specific formats
@@ -440,19 +447,7 @@ use sp_core::OpaqueMetadata;
 use xcm_fee_payment_runtime_api::Error as XcmPaymentApiError;
 
 impl_runtime_apis! {
-	impl sp_api::Core<Block> for Runtime {
-		fn version() -> RuntimeVersion {
-			VERSION
-		}
 
-		fn execute_block(block: Block) {
-			Executive::execute_block(block)
-		}
-
-		fn initialize_block(header: &<Block as BlockT>::Header) -> ExtrinsicInclusionMode {
-			Executive::initialize_block(header)
-		}
-	}
 
 	impl sp_api::Metadata<Block> for Runtime {
 		fn metadata() -> OpaqueMetadata {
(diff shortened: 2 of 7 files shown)


primitives/src/runtime_api.rs
```diff
@@ -0,0 +1,20 @@
+use crate::constants::transaction::MaxExtrinsicSize;
+use codec::Codec;
+use frame_support::BoundedVec;
+use sp_weights::Weight;
+
+sp_api::decl_runtime_apis! {
+	pub trait FeeEstimationApi<AccountId, AssetId, Balance>
```
enthusiastmartin, line 7: Naming. 

I would go with slightly different naming .

TransactionFeeAPI

and methods:
- estimate_fee
- estimate_uxt_fee

I let @jak-pan share his thoughts here on naming,

primitives/src/runtime_api.rs
```diff
@@ -0,0 +1,20 @@
...
+	where
+		AccountId: Codec,
+		AssetId: Codec,
+		Balance: Codec,
+	{
+		fn estimate_fee_payment(weight: Weight, account_id: AccountId) -> (AssetId, Balance);
```
enthusiastmartin, line 13: The result of both methods should return an Option.

In the actual implementation it seems it does unwrap_or in multiple places , so it can potentially return incorrect result, not the one you expect.

I would change the return value to be a struct. This would  help with potential future extension if we want  to include additional details.

primitives/src/runtime_api.rs
```diff
@@ -0,0 +1,20 @@
...
+
+		fn estimate_fee_payment_for_extrinsic(
+			uxt: BoundedVec<u8, MaxExtrinsicSize>,
```
enthusiastmartin, line 16: No need to use boundedVec here and also no need for new constant.

Use the same as it is done in TransactionPaymentAPI.

	`uxt: <Block as BlockT>::Extrinsic,`

runtime/hydradx/src/lib.rs
```diff
@@ -450,19 +457,7 @@ use sp_core::OpaqueMetadata;
 use xcm_fee_payment_runtime_api::Error as XcmPaymentApiError;
 
 impl_runtime_apis! {
-	impl sp_api::Core<Block> for Runtime {
```
enthusiastmartin, line 453: nitpick: not sure what happened here but if possible, would you revert these changes

this PR should not touch/change anything else.

runtime/hydradx/src/lib.rs
```diff
@@ -1186,4 +1181,66 @@ impl_runtime_apis! {
...
+	impl primitives::runtime_api::FeeEstimationApi<Block, AccountId, AssetId, Balance> for Runtime {
+		fn estimate_fee_payment(weight: Weight, account_id: AccountId) -> (AssetId, Balance) {
+			// Get the currency configured for this account
+			let currency = MultiTransactionPayment::account_currency(&account_id);
+
+			// Convert weight to base fee using native asset calculations
+			let native_fee = TransactionPayment::weight_to_fee(weight);
+
+			// Get the price of the account's currency if applicable
+			let price = MultiTransactionPayment::price(currency).unwrap_or(Price::one());
```
enthusiastmartin, line 1206: this is not correct. if there is not price, it should not return anything, hence the comment about return an option.

runtime/hydradx/src/lib.rs
```diff
@@ -1186,4 +1181,66 @@ impl_runtime_apis! {
...
+
+			// Calculate the final fee in the account's currency
+			let fee_in_currency = price.checked_mul_int(native_fee).unwrap_or(native_fee);
```
enthusiastmartin, line 1209: again, if it fails - it is return incorrect info.

after you change, this should look something like :

	`price.checked_mul_int(fee).map(|f| f.max(One::one()))`
	
we dont want to see 0 fee. this would be consistent with other stuff.

runtime/hydradx/src/lib.rs
```diff
@@ -1186,4 +1181,66 @@ impl_runtime_apis! {
...
+
+		fn estimate_fee_payment_for_extrinsic(
+			uxt: BoundedVec<u8, MaxExtrinsicSize>,
+			account_id: AccountId,
+		) -> (AssetId, Balance) {
+			use codec::DecodeAll;
+
+			// Decode the extrinsic
+			let uxt = UncheckedExtrinsic::decode_all(&mut &uxt[..])
+				.expect("Invalid extrinsic provided");
```
enthusiastmartin, line 1222: this could cause panics easily.

there is not need to have boundev and and decode stuff. See my other comment to use differetn type for the paramter.

runtime/hydradx/src/lib.rs
```diff
@@ -1186,4 +1181,66 @@ impl_runtime_apis! {
...
+			let weight_fee = TransactionPayment::weight_to_fee(info.weight);
+
+			// Total native fee
+			let native_fee = len_fee.saturating_add(weight_fee);
+
+			// Convert to account's preferred currency
+			let currency = MultiTransactionPayment::account_currency(&account_id);
+			let price = MultiTransactionPayment::price(currency).unwrap_or(Price::one());
+			let fee_in_currency = price.checked_mul_int(native_fee).unwrap_or(native_fee);
+
```
enthusiastmartin, line 1241: no need to do this again - just call the other method.

enthusiastmartin (commented):
thank you for your submission. 

all-in-all, it could work but i have some suggestions - see comments below.

Other issue which I am not sure yet about is the place of the runtime api 

It wont be definitely in the primives. I will update on this one.

enthusiastmartin:
dormant for quite some time.

PR #976, opened 2024-12-18
feat: add staking migration

enthusiastmartin:
Add staking migration in prepare for opengov release.

This migration is clearing existing staking votes if there are any left.


enthusiastmartin:
no longer relevant

Issue #972, opened 2024-12-16
Omnipool dynamic fee adjustments


enthusiastmartin:
This includes multiple things to do:

- Multiblock dynamic fee adjustment
- mint H20 for asset fees
- use portion of H20 fees to buy HDX
- Remove imbalance

PR #981, opened 2025-01-03
feat: omnipool - remove imbalance

enthusiastmartin:
This is part of the #972 

due to the fact that this does lot of changes in api - It is done as separeate PR to make things clear regarding the fee changes. 

Note that is built on top of #910 PR to avoid complicated merge conflicts. Hence that should be merged first.


enthusiastmartin:
part of #987

PR #983, opened 2025-01-07
feat: rolling dca

mrq1911:


pallets/dca/src/lib.rs
```diff
@@ -914,6 +923,10 @@ impl<T: Config> Pallet<T> {
 		schedule: &Schedule<T::AccountId, T::AssetId, BlockNumberFor<T>>,
 		amount_to_unreserve: Balance,
 	) -> DispatchResult {
+		if schedule.is_rolling() {
```
dmoka, line 926: strange that this is the only place this guard clause needed to be added, and not in `replan_or_complete` function, where the schedule is completed when there is not enough left in the RemainingAmounts.

But based on the unit tests, seems like working, so nice! An integration test might be handy here.

pallets/dca/src/types.rs
```diff
@@ -34,6 +34,12 @@ pub struct Schedule<AccountId, AssetId, BlockNumber> {
 	pub order: Order<AssetId>,
 }
 
+impl<AccountId, AssetId, BlockNumber> Schedule<AccountId, AssetId, BlockNumber> {
+	pub fn is_rolling(&self) -> bool {
+		self.total_amount == 0
```
dmoka, line 39: cool modification without breaking API!

pallets/dca/src/lib.rs
dmoka, line 926: strange that this is the only place this guard clause needed to be added, and not in replan_or_complete function, where the schedule is completed when there is not enough left in the RemainingAmounts.

But based on the unit tests, seems like working, so nice! An integration test might be handy here.

pallets/dca/src/types.rs
```diff
@@ -34,6 +34,12 @@ pub struct Schedule<AccountId, AssetId, BlockNumber> {
 	pub order: Order<AssetId>,
 }
 
+impl<AccountId, AssetId, BlockNumber> Schedule<AccountId, AssetId, BlockNumber> {
+	pub fn is_rolling(&self) -> bool {
```
dmoka, line 38: cool modification without breaking API!

pallets/dca/src/lib.rs
mrq1911: the remaining amount is never decreased, therefore the DCA runs until it actually encounters any error which makes it terminate

mrq1911: rolling schedule cannot be completed as there is not any goal to complete

dmoka, line 926: I would also adjust the README, describing this rolling kind of schedules.

dmoka (approved):
Please adjsut README first, apart from that LGTM, good to go.

Issue #986, opened 2025-01-10
Add LocationToAccountId converter for GlobalConsensus(..)


green-jay:
### Description
Currently there is no way of translating foreign consensus xcm origins to AccountIds. There are two scenarios at hand where this will be useful:

**1. Transacting on Hydration from Kusama accounts**
E.g. Adding KSM to Omnipool from Kusama treasury. It will be assigned the ownership of the LP position so it can trustlessly control it. But also any signed origin, parachain, etc. should be able to operate on Hydration remotely from Kusama.

**2. BSX listing**
Similar case as above, BSX can be listed from Basilisk treasury and controlled remotely via XCM.

Will be done via `GlobalConsensusParachainConvertsFor` once we upgrade to newer SDK.

### Tasklist:
- [ ] GlobalConsensus(Kusama) is supported
### Future work:
- [ ] GlobalConsensus(Parachain(Root)) is supported

### Auxiliary info
Relevant configuration:
https://github.com/galacticcouncil/hydration-node/blob/4c3a2d736bf64fd61a3bb08e5b0d9731b1a95011/runtime/hydradx/src/xcm.rs#L414

We could use/inspire by [GlobalConsensusConvertsFor](https://github.com/paritytech/polkadot-sdk/blob/738282a2c4127f5e6a1c8d50235ba126b9f05025/polkadot/xcm/xcm-builder/src/location_conversion.rs#L387C12-L387C38) for Kusama and [GlobalConsensusParachainConvertsFor](https://github.com/polkadot-fellows/runtimes/blob/c1f8b7f0b282818ef527a9de2374be70b4377736/system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs#L101) for Parachains on Kusama.

Test snippet (failing currently, should pass):
```rust
#[test]
fn assert_kusama_root_account() {
    // Initialize the Externalities environment
    sp_io::TestExternalities::new_empty().execute_with(|| {
        let ksm_root_location = Location::new(2, [GlobalConsensus(Kusama)]);
        let ksm_root_account = GlobalConsensusConvertsFor::<UniversalLocation, AccountId>::convert_location(&ksm_root_location)
            .expect("Failed to convert location");

        // // Example treasury account, replace with the actual expected value
        let expected_account = AccountId::from_ss58check("5G4KKqSKDkiMGiPzCQY12dSB15aBikyNQJL9VDmbMH4SxiWD")
            .expect("Invalid SS58 address format");
[...]

PR #990, opened 2025-01-14
feat: Add Kusama xcm location converter

green-jay:
Solves #986


runtime/hydradx/src/tests.rs
```diff
@@ -363,3 +367,24 @@ mod xcm_fee_payment_api_tests {
...
+#[test]
+fn assert_kusama_root_account() {
+	// Initialize the Externalities environment
+	sp_io::TestExternalities::new_empty().execute_with(|| {
+		let ksm_root_location = Location::new(2, [GlobalConsensus(Kusama)]);
+		let ksm_root_account =
+			GlobalConsensusConvertsFor::<UniversalLocation, AccountId>::convert_location(&ksm_root_location)
+				.expect("Failed to convert location");
+
+		// // Example treasury account, replace with the actual expected value
```
dmoka, line 380: comment feels a bit redundant, I would just name the variable to `expected_treasury_account` and remove the comment.

runtime/hydradx/src/tests.rs
```diff
@@ -363,3 +367,24 @@ mod xcm_fee_payment_api_tests {
...
+		let expected_account = AccountId::from_ss58check("5G4KKqSKDkiMGiPzCQY12dSB15aBikyNQJL9VDmbMH4SxiWD")
+			.expect("Invalid SS58 address format");
+		assert_eq!(ksm_root_account, expected_account);
+
+		// // Example of a wrong account
```
dmoka, line 385:
```suggestion
```

runtime/hydradx/src/tests.rs
```diff
@@ -363,3 +367,24 @@ mod xcm_fee_payment_api_tests {
 		});
 	}
 }
+
+#[test]
+fn assert_kusama_root_account() {
```
dmoka, line 372: wondering if we could come up with the concrete use case in the test to justify the need for adding this `GlobalConsensusConverter`?

By doing so we dont need to couple our test to implementation details `GlobalConsensusConverter`, so it will become more flexible.

runtime/hydradx/src/xcm.rs
```diff
@@ -480,6 +481,9 @@ pub type LocationToAccountId = (
 	HashedDescription<AccountId, DescribeFamily<DescribeAllTerminal>>,
 	// Convert ETH to local substrate account
 	EvmAddressConversion<RelayNetwork>,
+	// Converts a location which is a top-level relay chain (which provides its own consensus) into a
+	// 32-byte `AccountId`.
+	GlobalConsensusConvertsFor<UniversalLocation, AccountId>,
```
dmoka, line 486: this seems like not Kusama specific, so I assume this will be triggered for Polkadot relay chain too?!

dmoka (approved):
LGTM, only small improvement in tests.

runtime/hydradx/src/xcm.rs
green-jay: no, GlobalConsensus junction is not used when referencing own parent, so this is used currently only with Parents: 2 and a relay chain root at the same time

runtime/hydradx/src/tests.rs
green-jay: Good suggestion but that would require a complex integration test that would mock P<>K bridge and performing a transact instruction execution remotely. I would keep it for later as now it is a pretty straight forward change inherited from polkadot sdk and we need the change to be merged asap

PR #994, opened 2025-01-22
feat: more concurent deciding referenda in root track

mrq1911:


runtime/hydradx/src/governance/tracks.rs
```diff
@@ -51,7 +51,7 @@ const TRACKS_DATA: [(u16, pallet_referenda::TrackInfo<Balance, BlockNumber>); 10
 		0,
 		pallet_referenda::TrackInfo {
 			name: "root",
-			max_deciding: 1,
+			max_deciding: 3,
```
jak-pan, line 54: I would do just 2, we need it only temporarily anyway. 

All non-upgrade things should be on different tracks.

+ needs re-benchmark of democracy and conviction voting. @enthusiastmartin maybe update some CONST for maximum tracks / votes?

enthusiastmartin: nope. staking and voting dont know anything about this.

jak-pan: https://github.com/galacticcouncil/hydration-node/blob/3b4eabee7d8bc97c06dfce7c04d10a5fede14f16/pallets/staking/src/benchmarks.rs#L65

https://github.com/galacticcouncil/hydration-node/blob/3b4eabee7d8bc97c06dfce7c04d10a5fede14f16/pallets/staking/src/types.rs#L148

https://github.com/galacticcouncil/hydration-node/blob/3b4eabee7d8bc97c06dfce7c04d10a5fede14f16/runtime/hydradx/src/assets.rs#L1468
-> this should go to governance part of the runtime config tbf

https://github.com/galacticcouncil/hydration-node/blob/master/runtime/hydradx/src/governance/old.rs#L119
This looks misconfigured?

https://github.com/galacticcouncil/hydration-node/blob/master/pallets/democracy/src/benchmarking.rs#L138

We might be missing this completely in the benchmarks of opengov?

vgantchev (approved):

Issue #989, opened 2025-01-14
Benchmark XCM weights in a better way


dmoka:
![image](https://github.com/user-attachments/assets/ce5fbab8-e6d0-44df-92a8-3513082cf183)

Link1: https://github.com/galacticcouncil/hydration-node/blob/02adadc5257790e0dc3350dd116933c26eb60ec9/runtime/hydradx/src/xcm.rs#L194

LInk2: https://github.com/galacticcouncil/hydration-node/blob/02adadc5257790e0dc3350dd116933c26eb60ec9/runtime/hydradx/src/xcm.rs#L122

Most parachain uses `FixedWeightBounds`, but many system chains use `WeightInfoBounds`

It seems like we should also utilize `WeightInfoBounds` to have more representative weights

dmoka:
You can use [pallet-xcm-benchmarks](https://github.com/paritytech/polkadot-sdk/tree/master/polkadot/xcm/pallet-xcm-benchmarks/). You need to [configure it in your runtime](https://github.com/paritytech/polkadot-sdk/blob/ddffa027d7b78af330a2d3d18b7dfdbd00e431f0/cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs#L1911) and then run benchmarks. You'll end up with [these files](https://github.com/paritytech/polkadot-sdk/tree/master/cumulus/parachains/runtimes/assets/asset-hub-westend/src/weights/xcm). The generic and fungible ones are autogenerated and the mod.rs you just copy and paste, it references the correct ones

PR #995, opened 2025-01-22
feat: xcm benchmarking

dmoka:
https://github.com/galacticcouncil/hydration-node/issues/989

TODO:
- [x] double check if everything is cheaper. If everything is chaper, we wait 2409, if not we are good to go
- [ ] UI should test it out and change dest fees by running chopstick matrix
- [ ] talk to moonbeam about this changes
- [x] regenerate benches on ref machine
- [x] fix integration tests


jak-pan:
How did you check the weights are lower than b4? Can you drop old bench results somewhere so we can confirm?

runtime/hydradx/src/lib.rs
```diff
@@ -1135,12 +1158,142 @@ impl_runtime_apis! {
...
 				fn get_asset() -> Asset {
 					Asset {
-						id: AssetId(Location::here()),
+						id: AssetId(PolkadotLocation::get()),
 						fun: Fungible(ExistentialDeposit::get()),
 					}
 				}
 			}
 
+			use primitives::constants::currency::UNITS;
```
jak-pan, line 1167: we should probably split this to separate file as it's getting crowded

dmoka: you remember, the last time when we tried to put the runtime apis into a separate file, we had a weird error on prod, so maybe it is just better to leave it in lib.rs...

dmoka:
> How did you check the weights are lower than b4? Can you drop old bench results somewhere so we can confirm?

Just did some testing. Not all extrinsic became cheaper. 

Here is the old weight for all xcm extrinsic: 

```
	pub const BaseXcmWeight: XcmWeight = XcmWeight::from_parts(100_000_000, 0);
```

This was used as an "overestimate". It seems it was not that good overestimate.

Here are the xcm extrinsics that became more expensive after the current rebenchmarking:

```
	withdraw_asset
	transfer_asset
	transfer_reserve_asset
	initiate_reserve_withdraw
	receive_teleported_asset
	deposit_asset
	deposit_reserve_asset
	report_holding
	report_error
	claim_asset
	subscribe_version
	unsubscribe_version
	query_pallet
	report_transact_status
	exchange_asset
```

And here are the ones that became cheaper:

```
	reserve_asset_deposited
	buy_execution
	query_response
	transact
	refund_surplus
	set_error_handler
	set_appendix
	clear_error
	descend_origin
	clear_origin
	trap
	burn_asset
	expect_asset
	expect_origin
	expect_error
	expect_transact_status
	expect_pallet
	clear_transact_status
	set_topic
	clear_topic
	unpaid_execution
	set_fees_mode
```

Based on this result, it seems we gotta wait for `2409` to not to break any UI?

runtime/hydradx/src/tests.rs
```diff
@@ -329,9 +329,15 @@ mod xcm_fee_payment_api_tests {
 				},
 			]);
 
-			assert_eq!(
-				Runtime::query_xcm_weight(VersionedXcm::from(xcm_message.clone())),
-				Ok(Weight::from_parts(200000000, 0))
+			let queried_weight = Runtime::query_xcm_weight(VersionedXcm::from(xcm_message.clone())).unwrap();
+			let ref_time = queried_weight.ref_time();
+			assert!(
```
Roznovjak, line 334: Let's not use any range. Assert that it returns `Ok` with some value should be enough.

runtime/hydradx/src/lib.rs
```diff
@@ -1135,12 +1158,142 @@ impl_runtime_apis! {
...
+
+			parameter_types! {
```
Roznovjak, line 1169: I would place `PriceForParentDelivery` here

Roznovjak (commented):
pallet_xcm uses `type WeightInfo = weights::pallet_xcm::HydraWeight<Runtime>;`, is that correct?

Roznovjak:
> pallet_xcm uses `type WeightInfo = weights::pallet_xcm::HydraWeight<Runtime>;`, is that correct?

Yes, it's correct. Ignore this.

Roznovjak:
> > How did you check the weights are lower than b4? Can you drop old bench results somewhere so we can confirm?
> 
> Just did some testing. Not all extrinsic became cheaper.
> 
> Here is the old weight for all xcm extrinsic:
> 
> ```
> 	pub const BaseXcmWeight: XcmWeight = XcmWeight::from_parts(100_000_000, 0);
> ```
> 
> This was used as an "overestimate". It seems it was not that good overestimate.
> 
> Here are the xcm extrinsics that became more expensive after the current rebenchmarking:
> 
> ```
> 	withdraw_asset 89.25
> 	transfer_asset 204.92
> 	transfer_reserve_asset 583%
> 	initiate_reserve_withdraw 326.35%
> 	receive_teleported_asset  18446744073609.55% (Weight::MAX as not used)
> 	deposit_asset  57.39%
> 	deposit_reserve_asset   455.64%
> 	report_holding 318.42%
> 	report_error 280.14%
> 	claim_asset 39.10%
> 	subscribe_version 382.13%
> 	unsubscribe_version  4.51%
> 	query_pallet 295.17%
> 	report_transact_status   280.23%
> 	exchange_asset:  3848.85%
> ```
> 
> And here are the ones that became cheaper:
> 
> ```
> 	reserve_asset_deposited 98.01%
> 	buy_execution  98.71%
> 	query_response 65.93%
> 	transact 89.38%
> 	refund_surplus 95.54%
> 	set_error_handler 98.76%
> 	set_appendix 98.78%
> 	clear_error 98.79%
> 	descend_origin 98.76%
> 	clear_origin 98.81%
> 	trap 98.83%
> 	burn_asset 74.84%
> 	expect_asset 92.23%
> 	expect_origin 98.82%
> 	expect_error 98.80%
> 	expect_transact_status 98.57%
> 	expect_pallet 86.65%
> 	clear_transact_status 98.73%
> 	set_topic 98.80%
> 	clear_topic 98.80%
> 	unpaid_execution  98.77%
> 	set_fees_mode 98.82%
> ```
> 
> Based on this result, it seems we gotta wait for `2409` to not to break any UI, that's what we discussed last time?

How did you get there results? I don't see any change in pallet-xcm weight file.

runtime/hydradx/src/tests.rs
dmoka: What is the issue with using range?

The problem with specific assertion is that that it will fail every time we rebenchmark. By having a range, it will fail only if the weight significantly changes, which is good, I mean, we wanna be aware of that.

dmoka:
> How did you get there results? I don't see any change in pallet-xcm weight file.

@Roznovjak it is not about changing the weight of extrinsics of pallet-xcm, it is about the weights associated with the XCM instructions. Check weights/xcm/mod.rs for the details.

Roznovjak (approved):

wigy-opensource-developer:
@dmoka @jak-pan Could any parachain open a HRMP channel with Hydration since [this was merged](https://github.com/galacticcouncil/hydration-node/pull/995/changes#diff-461d0970333463649bc70d7ee0ba155275ef00bb196f31743f62f6661682fc7aR103)? If yes, how?

PR #1003, opened 2025-01-27
fix: remove stableswap ed check

dmoka:
We should not do this check as we can have accounts with balance less than `T::MinPoolLiquidity` then the following check prevents the shares withdrawal

https://github.com/galacticcouncil/hydration-node/blob/ea4e622fe5adb0dda12dd4000ac6b55849ff7649/pallets/stableswap/src/lib.rs#L574-L578

TODO:
- [x] check on chopsticks the wasm if works

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4919,7 +4919,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "282.0.0"
+version = "283.0.0"
 dependencies = [
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-parachain-system",
@@ -9195,7 +9195,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-stableswap"
-version = "4.2.0"
+version = "4.3.0"
 dependencies = [
  "bitflags 1.3.2",
  "frame-benchmarking",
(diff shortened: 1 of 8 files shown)


jak-pan:
I think this check was supposed to check the shares left after withdrawal to stop you from being locked out as there is similar check in XYK if I remember correctly. 

However, this won't prevent it as you can just send shares away. This means we can end up in a state where there is enough shares in a pool but there is not enough shares in anybody's account to actually withdraw anything. I think the solution to this is to tie the minimal pool balances/shares to existential deposit.

This should be viable thing after we implement #1007 

However this number probably shouldn't change. I would consider setting it to the fallback values so we don't end up in the same situation.

dmoka:
> However this number probably shouldn't change. I would consider setting it to the fallback values so we don't end up in the same situation.

So you mean that instead of removing the checks, we should lower the MinPoolLiquidity so to 1_000 from 1_000_000?

jak-pan:
> > However this number probably shouldn't change. I would consider setting it to the fallback values so we don't end up in the same situation.
> 
> So you mean that instead of removing the checks, we should lower the MinPoolLiquidity so to 1_000 from 1_000_000?

No. :D I don't know how this was deduced. We should remove this check and I made a new issue for the problem this was probably supposed to fix (but in a wrong way here)

enthusiastmartin (approved):

PR #1014, opened 2025-02-03
fix: unified events max callstack

dmoka:
- Making the broadcast pallet event stack changing functions fallable, so we prevent wront usages.

- Whitelisting broadcast pallet Stack storage and clearing it up in the end of the block, to save space and improve performance


jak-pan (approved):

pallets/broadcast/src/lib.rs
```diff
@@ -61,18 +62,18 @@ pub mod pallet {
...
-	/// After the stack is full, we start to increase the overflow count,
-	/// so we how many times we can ignore the removal from the context.
-	#[pallet::storage]
-	pub(super) type OverflowCount<T: Config> = StorageValue<_, u32, ValueQuery>;
-
 	#[pallet::error]
-	pub enum Error<T> {}
+	pub enum Error<T> {
+		///The execution context has reached its maximum size
+		ExecutionContextFull,
```
jak-pan, line 73:
```suggestion
		ExecutionCallStackOverflow,
```

pallets/broadcast/src/lib.rs
```diff
@@ -61,18 +62,18 @@ pub mod pallet {
...
+		///Cannot remove from the execution context as it is empty
+		ExecutionContextEmpty,
```
jak-pan, line 75:
```suggestion
		ExecutionCallStackUnderflow,
```

pallets/broadcast/src/lib.rs
```diff
@@ -61,18 +62,18 @@ pub mod pallet {
...
-	/// To handle the overflow of increasing the execution context.
-	/// After the stack is full, we start to increase the overflow count,
-	/// so we how many times we can ignore the removal from the context.
-	#[pallet::storage]
-	pub(super) type OverflowCount<T: Config> = StorageValue<_, u32, ValueQuery>;
-
 	#[pallet::error]
-	pub enum Error<T> {}
+	pub enum Error<T> {
+		///The execution context has reached its maximum size
```
jak-pan, line 72:
```suggestion
		///The execution context call stack has reached its maximum size
```

pallets/broadcast/src/lib.rs
```diff
@@ -61,18 +62,18 @@ pub mod pallet {
...
+		ExecutionContextFull,
+		///Cannot remove from the execution context as it is empty
```
jak-pan, line 74:
```suggestion
		///Execution context call stack is empty
```

pallets/broadcast/src/lib.rs
jak-pan: Reasoning : If you get this error you could  infer what you are doing wrong

pallets/broadcast/src/lib.rs
jak-pan: This is probably bigger issue than overflow which could happen... This can happen only if you fuk up the impl. Maybe we should do something extra here???

jak-pan (changes requested):
Just naming otherwise LGTM can merge

pallets/broadcast/src/lib.rs
dmoka: nice one!

pallets/broadcast/src/lib.rs
dmoka: Before I added debug_assert! for us, so if we fuk up the impls then the developers know it, if tested. But then the fuzzer started failing on these as it just called this function, triggering the debug assert

dmoka: We could add some logging, so we get notified better if this happens

jak-pan (approved):

PR #1023, opened 2025-02-11
fix: use bounded vec in extrinsics

dmoka:
In API extrinsics, we should use BoundedVec instead of Vec as it might be a potential attack vector.

- [x] verify if it works in JS


integration-tests/src/dca.rs
```diff
@@ -518,7 +518,7 @@ mod omnipool {
 				DAI,
 				amount_out,
 				Balance::MAX,
-				trade
+				BoundedVec::truncate_from(trade)
```
Roznovjak, line 521: Can `.try_into().unwrap()` be used here? I would prefer it over something truncating.

pallets/route-executor/src/lib.rs
```diff
@@ -350,21 +345,21 @@ pub mod pallet {
 
 			let existing_route = Self::get_route(asset_pair);
 
-			match Self::validate_route(&existing_route) {
+			match Self::validate_route(existing_route.clone()) {
 				Ok((reference_amount_in, reference_amount_in_for_inverse)) => {
-					let new_route_validation = Self::validate_sell(new_route.clone(), reference_amount_in);
+					let new_route_validation = Self::validate_sell(new_route.clone().to_vec(), reference_amount_in);
```
Roznovjak, line 350: Parameter type in `validate_sell()` can be changed to `Route`

pallets/route-executor/src/lib.rs
```diff
@@ -741,7 +736,18 @@ impl<T: Config> Pallet<T> {
...
+			};
+
+			let sell_result = Self::sell(
+				origin,
+				asset_in,
+				asset_out,
+				amount_in,
+				u128::MIN.into(),
+				route_as_bounded_vec,
+			);
```
Roznovjak, line 750: This would not be necessary if `route` param was `Route`

pallets/route-executor/src/lib.rs
```diff
@@ -820,11 +826,8 @@ impl<T: Config> Pallet<T> {
 		Ok(amount_in_and_outs)
 	}
 
-	fn insert_route(asset_pair: AssetPair<T::AssetId>, route: Vec<Trade<T::AssetId>>) -> DispatchResultWithPostInfo {
-		let route_as_bounded_vec: BoundedVec<Trade<T::AssetId>, sp_runtime::traits::ConstU32<MAX_NUMBER_OF_TRADES>> =
-			route.try_into().map_err(|_| Error::<T>::MaxTradesExceeded)?;
```
Roznovjak, line 825: What's the main issue with the previous implementation. The storage is bounded so there should be not any attack vector.

scripts/init-testnet/chopsticks/exec.js
```diff
@@ -0,0 +1,67 @@
+// Simple HydraDX Extrinsic Executor with hardcoded values
```
Roznovjak, line 1: what are these files?

pallets/route-executor/src/lib.rs
dmoka: but someone could send an encoded extrinsic with a large number of routes there. If we bound it in the extrinsic param level, we can prevent that.

Roznovjak: Would it be a problem? To receive such extrinsic

dmoka: in theory yes, as the data size can be very large of such encoded data. @jak-pan mentioned that there is some barrier mechanism for this within substrate which probably prevents it, we still decided to fix it

PR #1025, opened 2025-02-13
feat: bifrost oracle implementation

dmoka:
- [x] rebench on reference
- [x] fix todos
- [x] add sanity check for not being able to set insane value. Check PR comment for more info


pallets/ema-oracle/src/lib.rs
```diff
@@ -295,15 +304,39 @@ pub mod pallet {
...
+					T::BlockNumberProvider::current_block_number(),
+				);
+				if ordered_pair(asset_a, asset_b) == (asset_a, asset_b) {
+					e
+				} else {
+					e.inverted()
+				}
+			};
+
+			Self::on_entry(BITFROST_SOURCE, (asset_a, asset_b), entry).map_err(|_| Error::<T>::TooManyUniqueEntries)?;
```
enthusiastmartin, line 338: here should be ordered pair, not (asset_a,asset_b)

dmoka: nice catch

dmoka:
One more sanity check:

- We should validate that the next set price within some boundaries. 10% or so, configurable.

- So we make sure so people can not send some insane value.

pallets/ema-oracle/src/lib.rs
```diff
@@ -142,6 +145,13 @@ pub mod pallet {
 		/// Whitelist determining what oracles are tracked by the pallet.
 		type OracleWhitelist: Contains<(Source, AssetId, AssetId)>;
 
+		/// Asset id converter based on xcm locations
+		type CurrencyIdConvert: sp_runtime::traits::Convert<Location, Option<AssetId>>;
+
+		/// Maximum allowed percentage difference for bifrost oracle price update
+		#[pallet::constant]
+		type MaxAllowedPriceDifference: Get<(u32, u32)>;
```
enthusiastmartin, line 153: why not Permill? wouldnt it be more clear?

pallets/ema-oracle/src/lib.rs
```diff
@@ -88,12 +92,11 @@ pub use types::*;
 #[allow(clippy::all)]
 pub mod weights;
 pub use weights::WeightInfo;
-
-mod benchmarking;
```
enthusiastmartin, line 92: why need to remove this from here?

pallets/ema-oracle/src/lib.rs
```diff
@@ -299,11 +316,42 @@ pub mod pallet {
...
+			asset_b: Box<polkadot_xcm::VersionedLocation>,
+			price: (Balance, Balance),
 		) -> DispatchResult {
 			T::BifrostOrigin::ensure_origin(origin)?;
+
+			let asset_a_loc = Location::try_from(*asset_a).map_err(|()| Error::<T>::BadVersion)?;
+			let asset_b_loc = Location::try_from(*asset_b).map_err(|()| Error::<T>::BadVersion)?;
+
+			let asset_a = T::CurrencyIdConvert::convert(asset_a_loc).ok_or(Error::<T>::AssetNotFound)?;
+			let asset_b = T::CurrencyIdConvert::convert(asset_b_loc).ok_or(Error::<T>::AssetNotFound)?;
```
enthusiastmartin, line 329: hm.. why the need of two conversion ? 

i would epxect to have just passed the given param into the conversion function, and it should handle it

pallets/ema-oracle/src/lib.rs
```diff
@@ -142,6 +145,13 @@ pub mod pallet {
 		/// Whitelist determining what oracles are tracked by the pallet.
 		type OracleWhitelist: Contains<(Source, AssetId, AssetId)>;
 
+		/// Asset id converter based on xcm locations
+		type CurrencyIdConvert: sp_runtime::traits::Convert<Location, Option<AssetId>>;
```
enthusiastmartin, line 149: i believe it is Location conversion.  Name suggest otherwise.

pallets/ema-oracle/src/lib.rs
mrq1911: yes, percent or permill pls

pallets/ema-oracle/src/lib.rs
mrq1911: yes, but we are naming it like this in runtime already :)

mrq1911: i guess renaming would require bigger refactor, this is atleast consistent

pallets/ema-oracle/src/lib.rs
```diff
@@ -299,11 +316,42 @@ pub mod pallet {
...
+				);
+				if ordered_pair == (asset_a, asset_b) {
+					e
+				} else {
+					e.inverted()
+				}
+			};
+
+			if let Some(reference_entry) = Self::oracle((BIFROST_SOURCE, ordered_pair, OraclePeriod::TenMinutes)) {
+				ensure!(
```
enthusiastmartin, line 347: please log error in this case

pallets/ema-oracle/src/lib.rs
dmoka: had a lot of challenge with benchmarking, so the easiest was to replace it to runtime.

pallets/ema-oracle/src/lib.rs
dmoka: not sure what yuu mean, the trait is Convert where the implementation can only convert one location at a time.

pallets/ema-oracle/src/lib.rs
dmoka: due to consistency reason, I will keep it as it is, only changing the comment.

dmoka:
One more change: we should not only update oracle ,but also create it if doesn't exist. Because we dont want to do it manually on prod

mrq1911 (approved):

pallets/ema-oracle/src/lib.rs
```diff
@@ -71,11 +71,16 @@ use frame_support::pallet_prelude::*;
 use frame_support::sp_runtime::traits::{BlockNumberProvider, One, Zero};
 use frame_support::traits::Contains;
 use frame_system::pallet_prelude::BlockNumberFor;
+use hydra_dx_math::ema::EmaPrice;
 use hydradx_traits::{
 	AggregatedEntry, AggregatedOracle, AggregatedPriceOracle, Liquidity, OnCreatePoolHandler,
 	OnLiquidityChangedHandler, OnTradeHandler, Volume,
 };
+use polkadot_xcm::opaque::lts::Location;
```
enthusiastmartin, line 79: this should not be needed

enthusiastmartin (approved):
one last  comment.

PR #1026, opened 2025-02-13
feat:  stableswap v2 - drift peg support

enthusiastmartin:


pallets/stableswap/src/lib.rs
```diff
@@ -1025,11 +1046,56 @@ pub mod pallet {
...
+			#[cfg(any(feature = "try-runtime", test))]
 			Self::ensure_remove_liquidity_invariant(pool_id, &initial_reserves);
 
 			Ok(())
 		}
+
+		#[pallet::call_index(11)]
+		#[pallet::weight(<T as Config>::WeightInfo::create_pool())]
+		#[transactional]
+		pub fn create_pool_with_pegs(
```
vgantchev, line 1058: _drifting_pegs ?

pallets/stableswap/src/lib.rs
```diff
@@ -328,6 +341,14 @@ pub mod pallet {
 
 		/// Failed to retrieve asset decimals.
 		UnknownDecimals,
+
+		/// List of provided pegs is incorrect.
+		IncorrectInitialPegs,
+
+		MissingTargetPegOracle,
```
dmoka, line 348: Error message is mising

pallets/stableswap/src/lib.rs
```diff
@@ -1582,3 +1702,224 @@ impl<T: Config> StableswapAddLiquidity<T::AccountId, T::AssetId, Balance> for Pa
...
+		}
+
+		let first_asset = pool_assets[0];
+
+		let mut r = vec![];
+		for (asset_id, source) in pool_assets.iter().zip(peg_sources.iter()) {
+			let p = match source {
+				PegSource::Value(peg) => (*peg, block_no),
+				PegSource::Oracle((source, period)) => {
+					//TODO: what if error - do we fail completely ?
```
dmoka, line 1759: todo

pallets/stableswap/src/lib.rs
```diff
@@ -166,6 +171,8 @@ pub mod pallet {
 		#[pallet::constant]
 		type AmplificationRange: Get<RangeInclusive<NonZeroU16>>;
 
+		type TargetPegOracle: RawOracle<Self::AssetId, Balance, BlockNumberFor<Self>>;
```
dmoka, line 174: Doc is missing

pallets/stableswap/src/tests/invariants.rs
```diff
@@ -1299,16 +1324,283 @@ proptest! {
...
+				for (x_n_plus, x_n) in final_reserves.iter().zip(reserves_add.iter()) {
+					let x_n_plus = U256::from(x_n_plus.amount);
+					let x_n = U256::from(x_n.amount);
+					assert!(x_n_plus * s >= x_n * s_plus);
+				}
+			});
+	}
+}
+
+proptest! {
```
mrq1911, line 1345: i didnt find any invariant tests that involve pegs

pallets/stableswap/src/lib.rs
```diff
@@ -706,7 +742,7 @@ pub mod pallet {
 				who: who.clone(),
 				shares,
 				amounts: vec![AssetAmount { asset_id, amount }],
-				fee: 0u128, // dev note: figure out the actual fee amount in this case. For now, we dont need it.
+				fee: 0u128, // dev note: figure out the actual fee amount in this case. For now, we dont need it. is deprecated anyway in favor of unified events
```
mrq1911, line 745: you can remove this todo then

pallets/stableswap/src/lib.rs
```diff
@@ -1048,6 +1243,7 @@ impl<T: Config> Pallet<T> {
 		asset_in: T::AssetId,
 		asset_out: T::AssetId,
 		amount_in: Balance,
+		update_peg: bool, // This is needed to distinguish between just calculation and trade execution
```
mrq1911, line 1246: you have to recalculate peg in both cases, bonly difference is if you store it or no, therefore something like write_peg or just_calculate would be better name for this, and you woudlnt have to comment it

math/src/stableswap/math.rs
```diff
@@ -28,21 +29,23 @@ pub fn calculate_out_given_in<const D: u8, const Y: u8>(
...
 	amplification: Balance,
+	pegs: &[(Balance, Balance)],
 ) -> Option<Balance> {
-	if idx_in == idx_out {
-		return None;
-	}
-	if idx_in >= initial_reserves.len() || idx_out >= initial_reserves.len() {
+	// Safety guards
+	let assets_ct = initial_reserves.len();
+	if assets_ct <= 1 || idx_in == idx_out || idx_in >= assets_ct || idx_out >= assets_ct || pegs.len() != assets_ct {
```
mrq1911, line 36: ideally i would skip all peg related logic if there are no pegs

pallets/stableswap/src/lib.rs
```diff
@@ -1024,11 +1112,118 @@ pub mod pallet {
...
+		#[pallet::call_index(11)]
+		#[pallet::weight(<T as Config>::WeightInfo::create_pool_with_pegs())]
+		#[transactional]
+		pub fn create_pool_with_pegs(
+			origin: OriginFor<T>,
+			share_asset: T::AssetId,
+			assets: BoundedVec<T::AssetId, ConstU32<MAX_ASSETS_IN_POOL>>,
+			amplification: u16,
+			fee: Permill,
+			peg_source: BoundedPegSources,
```
mrq1911, line 1157: it should be noted that oracle source always loads price of first asset in poll vs target asset, mby this could be extended by pair specification as well. seems limiting otherwise

pallets/stableswap/src/lib.rs
```diff
@@ -166,6 +174,9 @@ pub mod pallet {
 		#[pallet::constant]
 		type AmplificationRange: Get<RangeInclusive<NonZeroU16>>;
 
+		/// Oracle providing prices for asset pegs (if configured for pool)
+		type TargetPegOracle: RawOracle<Self::AssetId, Balance, BlockNumberFor<Self>>;
```
mrq1911, line 178: ability to use only raw oracle seems like oddly specific, reason?

mrq1911 (changes requested):

math/src/stableswap/math.rs
```diff
@@ -130,26 +129,25 @@ pub fn calculate_shares<const D: u8>(
...
 	fee: Permill,
+	pegs: &[(Balance, Balance)],
 ) -> Option<(Balance, Vec<Balance>)> {
-	if initial_reserves.len() != updated_reserves.len() {
-		return None;
-	}
-	let n_coins = initial_reserves.len();
-	if n_coins <= 1 {
+	// Safety guards
+	let assets_ct = initial_reserves.len();
```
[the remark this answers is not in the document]
enthusiastmartin: renamed it back.

pallets/stableswap/proptest-regressions/tests/invariants.txt
```diff
@@ -0,0 +1 @@
+cc 5f78856750028eda0c4274967a6c8f5efbd12733a1c56e1bc21f5614207b3d50
```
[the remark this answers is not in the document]
mrq1911: this is to make sure protest actually retest this case that previously failed

math/src/stableswap/math.rs
```diff
@@ -546,7 +607,28 @@ pub(crate) fn calculate_d_internal<const D: u8>(xp: &[Balance], amplification: B
 }
 
 /// Calculate Y. Reserves must be already normalized.
-fn calculate_y_internal<const D: u8>(xp: &[Balance], d: Balance, amplification: Balance) -> Option<Balance> {
+fn calculate_y_internal<const D: u8>(
+	xp: &[Balance],
+	d: Balance,
+	amplification: Balance,
+	pegs: &[(Balance, Balance)],
+	peg_omit: (Balance, Balance),
```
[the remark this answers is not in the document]
enthusiastmartin: i believe this is ok here, peg omit is a peg value for asset that we calculate the result for.

math/src/stableswap/math.rs
enthusiastmartin: there is always peg, if pool does not have pegs, it is all set to 1. 

initially, i had Option<Pegs> everywhere, but it looked very confusing, and introduces many IF THEN paths. It is not only list of pegs, but eg for Y calculation you nee to select a peg value that it is ommitted, that would have to an option too etc. 

with this approach there is no need to distinguish between pegs and no pegs.  And it does not introduce big overhead.

pallets/stableswap/src/lib.rs
enthusiastmartin: renamed to persit_peg and deleted the comment

pallets/stableswap/src/lib.rs
enthusiastmartin: we don't want the value that delayed, that's not how the mechanism is designed

added comment too.

mrq1911 (approved):

Issue #1027, opened 2025-02-14
Bug: No xyk pool existence check


dmoka:
## Description

PROD failing tx: https://hydration.subscan.io/extrinsic/6951559-2

The issue is that the schedule asset_in with id 102 is not a transaction fee currency, so it tries to do conversion calculation based on the XYK pool between 102-DOT. As there is no such pool, it fails with with error in math calculation, namely with Arithmetic error.

The error is indeed not descriptive, and to fix it we need a pool existence check, and if doesn't exist, we can return a descriptive error to the user. I will create an issue for this, and possibly fix it asap.

## Possible fix:
we could add guard statement like XYK:exists() and return descriptive dispatch error:
https://github.com/galacticcouncil/hydration-node/blob/46ba29ff2699324b932aaa83de906ba9fb054812/runtime/hydradx/src/assets.rs#L1822-L1832

Arsalaan-Alam:
@dmoka please take a look at #1028

PR #1028, opened 2025-02-16
fix: Add XYK pool existence check before conversion (#1027)

Arsalaan-Alam:
## **Description**
Added a check to verify if the **XYK pool exists** before performing **conversion calculations** in `calculate_in_given_out`. This prevents **arithmetic errors** and improves error handling with descriptive messages.

## **Related Issue**
Fixes: #1027 

## **Motivation and Context**
This change ensures transactions do not fail due to missing **XYK liquidity pools**. Instead of encountering an **arithmetic error**, users will receive a clear **DispatchError** message indicating that the pool does not exist.

##**How Has This Been Tested?**
- Built the runtime successfully using `cargo build`
- Ran unit tests with `cargo test --all`
- Verified transactions involving non-existent XYK pools return a proper error instead of failing with `ArithmeticError::Overflow`.

## **Checklist:**
- [x] I have added tests to cover my changes, including regression tests for missing XYK pools.

### DIFF
--- a/runtime/hydradx/src/assets.rs
+++ b/runtime/hydradx/src/assets.rs
@@ -1824,13 +1824,17 @@ impl SwappablePaymentAssetTrader<AccountId, AssetId, Balance> for XykPaymentAsse
 		asset_out: AssetId,
 		asset_out_amount: Balance,
 	) -> Result<Balance, DispatchError> {
-		let asset_pair_account = XYK::get_pair_id(AssetPair::new(insuff_asset_id, asset_out));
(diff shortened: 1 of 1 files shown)


dmoka:
Nice, please also add an integration test to verify the change and document this behaviour.

And also don't forget to bump runtime version.

Arsalaan-Alam:
on it

dmoka:
Closed due to inactivity, moved to https://github.com/galacticcouncil/hydration-node/pull/1066

PR #1034, opened 2025-02-28
feat: atokens swapping router support

mrq1911:
- adds new amm "aave" to router
- erc20 can be reserved (required by dca)
- on chain route is validated only by calculation now
- erc20mapping renamed to more accurate erc20encoding as it is not really 2 way mapping
- implemented true 2 way erc20 to asset mapping
- added support for aave to ema oracle


runtime/hydradx/src/evm/precompiles/erc20_mapping.rs
```diff
@@ -22,15 +22,20 @@
...
 use hex_literal::hex;
-use hydradx_traits::{evm::Erc20Mapping, RegisterAssetHook};
+use hydradx_traits::{evm::Erc20Encoding, RegisterAssetHook};
 use primitive_types::{H160, H256};
 use primitives::AssetId;
 
 pub struct HydraErc20Mapping;
 
-/// Erc20Mapping logic for HydraDX
+// impl<T: pallet_asset_registry::Config + BoundErc20> Erc20Mapping<AssetId> for HydraErc20Mapping {
```
dmoka, line 31: we can remove this if not needed

pallets/route-executor/src/lib.rs
```diff
@@ -724,29 +720,14 @@ impl<T: Config> Pallet<T> {
 	}
 
 	fn validate_sell(route: Vec<Trade<T::AssetId>>, amount_in: T::Balance) -> Result<T::Balance, DispatchError> {
+		// Validate that the route is properly structured
 		let asset_in = route.first().ok_or(Error::<T>::InvalidRoute)?.asset_in;
 		let asset_out = route.last().ok_or(Error::<T>::InvalidRoute)?.asset_out;
 
-		with_transaction::<T::Balance, DispatchError, _>(|| {
```
dmoka, line 730: What if the route is invalid, because of low liquidity etc? We wanna allow to set route for such?

integration-tests/src/router.rs
```diff
@@ -4370,197 +4370,6 @@ mod set_route {
 			});
 		}
 
-		#[test]
-		fn set_route_should_work_when_stored_route_is_broken_due_to_frozen_asset() {
```
dmoka, line 4374: With the new change, if the assets are frozen, we still allow setting route for that. I wonder if this comes with some security implication, but probably not really as they are frozen anyway so no trade is allowed.

runtime/adapters/src/lib.rs
```diff
@@ -611,6 +611,7 @@ where
 						Err(_) => return None,
 					}
 				}
+				PoolType::Aave => EmaPrice::from(1),
```
dmoka, line 614: why is the oracle price for any AAVE always `1`?

runtime/hydradx/src/assets.rs
```diff
@@ -1056,6 +1063,7 @@ impl AmmTradeWeights<Trade<AssetId>> for RouterWeightInfo {
 				}
 				PoolType::XYK => weights::pallet_xyk::HydraWeight::<Runtime>::router_execution_buy(c, e)
 					.saturating_add(<Runtime as pallet_xyk::Config>::AMMHandler::on_trade_weight()),
+				PoolType::Aave => Weight::zero(),
```
dmoka, line 1062: any specific reason that the weight is zero? Same for `calculate_spot_price_with_fee_weight`

runtime/hydradx/src/evm/aave_trade_executor.rs
```diff
@@ -0,0 +1,400 @@
...
+		_asset_out: AssetId,
+		amount_in: Balance,
+	) -> Result<Balance, ExecutorError<Self::Error>> {
+		if pool_type != PoolType::Aave {
+			return Err(ExecutorError::NotSupported);
+		}
+
+		// For both supply and withdraw, amount out is always 1:1
+		// to save weight we just assume the operation will be available
+		Ok(amount_in)
```
dmoka, line 294: here we need to calculate amount_out, why do we just return amount_in?

integration-tests/src/aave_router.rs
```diff
@@ -0,0 +1,449 @@
...
+
+		execution(pool);
+	});
+}
+
+const HDX: AssetId = 0;
+const DOT: AssetId = 5;
+const DAI: AssetId = 2;
+const ADOT: AssetId = 1_000_037;
+const ONE: u128 = 1 * 10_u128.pow(10);
```
dmoka, line 130: clippy would complain for `1 *`. But i guess we have clippy disabled anyway :D

pallets/route-executor/src/lib.rs
mrq1911: low liquidity situations are replicated by calculation

runtime/hydradx/src/evm/precompiles/erc20_mapping.rs
mrq1911: i forgot to finish implementation of it :)

integration-tests/src/router.rs
mrq1911: worst case different route has to be set

runtime/adapters/src/lib.rs
mrq1911: because aave always swaps for this price

mrq1911: given you are swapping underlying asset for atoken and vice versa

runtime/hydradx/src/assets.rs
mrq1911: cos i just return constant

mrq1911: 1

runtime/hydradx/src/evm/aave_trade_executor.rs
mrq1911: price is 1 always

integration-tests/src/aave_router.rs
mrq1911: 1 is 1

mrq1911: fuck clippy

martinfridrich (commented):
https://github.com/galacticcouncil/hydration-node/pull/1034/files#r1976057575 - finish this LGTM

PR #1035, opened 2025-02-28
feat: treasury rewards collators

mrq1911:


martinfridrich (approved):
IMO `RewardsAccount` would be better name than `RewardsBag` but this is not blocking

PR #1042, opened 2025-03-12
feat: stableswap - add liquidity limit

enthusiastmartin:
This PR adds new extrinsic that adds liquidity `add_assets_liquidity`. 

The difference between add_liquidity and add_assets_liquidity is that this new extrisnics allows to specify min shares that LP expects to receive.

### DIFF
--- a/pallets/stableswap/src/lib.rs
+++ b/pallets/stableswap/src/lib.rs
@@ -1181,6 +1181,24 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// Add liquidity to selected pool.
+		///
+		/// First call of `add_assets_liquidity` must provide "initial liquidity" of all assets.
+		///
+		/// If there is liquidity already in the pool, LP can provide liquidity of any number of pool assets.
+		///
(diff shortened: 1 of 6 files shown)


pallets/stableswap/src/lib.rs
```diff
@@ -1177,6 +1180,23 @@ pub mod pallet {
 			});
 			Ok(())
 		}
+
+		#[pallet::call_index(12)]
```
dmoka, line 1184: Please add a doc comment for the extrinsic

pallets/stableswap/src/lib.rs
```diff
@@ -494,6 +494,9 @@ pub mod pallet {
 
 		/// Add liquidity to selected pool.
 		///
+		/// Use `add_assets_liquidity` instead.
+		/// This extrinsics will be removed in the future.
```
dmoka, line 498: maybe we could use #[deprecated(note = "...")] annotation here?

dmoka (approved):

PR #1043, opened 2025-03-12
feat: stableswap audit minor findings 

enthusiastmartin:
This PR address minor findings in stableswap pallet.

1. Doc/comment update
2. Vec -> boundedvec
3. remove liquidity min amounts check
4. destroy pool removes peg info
5. added invariant check to withdraw_asset_amount to be consistent with others

### DIFF
--- a/integration-tests/src/aave_router.rs
+++ b/integration-tests/src/aave_router.rs
@@ -95,7 +95,7 @@ fn with_stablepool(execution: impl FnOnce(AssetId)) {
 		assert_ok!(Stableswap::create_pool(
 			hydradx_runtime::RuntimeOrigin::root(),
 			pool,
-			[DOT, ADOT].to_vec(),
+			BoundedVec::truncate_from([DOT, ADOT].to_vec()),
 			amplification,
 			fee,
 		));
--- a/pallets/stableswap/src/lib.rs
+++ b/pallets/stableswap/src/lib.rs
@@ -380,7 +380,7 @@ pub mod pallet {
 		pub fn create_pool(
 			origin: OriginFor<T>,
 			share_asset: T::AssetId,
-			assets: Vec<T::AssetId>,
+			assets: BoundedVec<T::AssetId, ConstU32<MAX_ASSETS_IN_POOL>>,
 			amplification: u16,
 			fee: Permill,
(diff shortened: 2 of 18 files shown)


integration-tests/src/aave_router.rs
```diff
@@ -95,7 +95,7 @@ fn with_stablepool(execution: impl FnOnce(AssetId)) {
 		assert_ok!(Stableswap::create_pool(
 			hydradx_runtime::RuntimeOrigin::root(),
 			pool,
-			[DOT, ADOT].to_vec(),
+			BoundedVec::truncate_from([DOT, ADOT].to_vec()),
```
dmoka, line 98: Instead of using truncate_from(), I would use `try_into().unwrap()` for most tests.

When the list has only 1-2 items, it's fine, but when it has more, it would be better for the tests to fail when the bound limit changes, rather than silently truncating it.

enthusiastmartin: no, i dont agree here. these are existing tests, and you can clearly see that it only uses two assets. no need to use try into. 
in fact, you never need to use it.  you were not be able to create pool with more assets before ,   the create pool would return error. And in integration tests you dont need to test it if you can create pool with more assets, you usually only setting up correct pools.

dmoka (approved):

pallets/stableswap/src/lib.rs
```diff
@@ -1042,8 +1046,11 @@ pub mod pallet {
 			// We convert vec of min amounts to a map.
 			// We first ensure the length , and if any asset is not found later on, we can return an error.
 			ensure!(min_amounts_out.len() == pool.assets.len(), Error::<T>::IncorrectAssets);
-			let mut min_amounts_out: BTreeMap<T::AssetId, Balance> =
-				min_amounts_out.into_iter().map(|v| (v.asset_id, v.amount)).collect();
+			let mut min_amounts_out_map = BTreeMap::new();
+			for v in min_amounts_out.into_iter() {
+				let r = min_amounts_out_map.insert(v.asset_id, v.amount);
+				ensure!(r.is_none(), Error::<T>::IncorrectAssets);
```
dmoka, line 1052: can r even be none after inserting some item to it?

PR #1049, opened 2025-03-15
feat: load forked state from rpc

dmoka:
- [x] test/verify created chainspec


mrq1911 (changes requested):
i will fix docker stuff

mrq1911 (approved):

PR #1057, opened 2025-03-31
feat: liquidation offchain worker

Roznovjak:


pallets/liquidation/src/lib.rs
```diff
@@ -274,14 +232,117 @@ pub mod pallet {
...
+		<T as Config>::Currency::transfer(
+			debt_asset,
+			&pallet_acc,
+			&T::ProfitReceiver::get(),
+			profit,
+			Preservation::Expendable,
+		)?;
+
+		Self::deposit_event(Event::Liquidated {
+			liquidator: maybe_signed_by.unwrap_or(pallet_acc),
```
mrq1911, line 335: just remove this field from event, then you never need signature

pallets/liquidation/src/lib.rs
```diff
@@ -274,14 +232,117 @@ pub mod pallet {
...
+			evm_address: user,
```
mrq1911, line 336: rename field evm_address -> user

mrq1911 (changes requested):

pallets/liquidation/src/lib.rs
```diff
@@ -189,79 +219,7 @@ pub mod pallet {
 			route: Route<AssetId>,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
```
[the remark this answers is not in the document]
Roznovjak: The question is whether we need signed version or not. Signed version of the extrinsic was there before, and we added unsigned version used by the worker. So I extracted the body of the extrinsics into a function and reuse it in both extrinsics.

runtime/hydradx/src/lib.rs
```diff
@@ -402,7 +402,30 @@ impl fp_self_contained::SelfContainedCall for RuntimeCall {
...
 		match self {
-			RuntimeCall::Ethereum(call) => call.validate_self_contained(info, dispatch_info, len),
+			RuntimeCall::Ethereum(call) => {
+				let mut tx_validity = call.validate_self_contained(info, dispatch_info, len);
+				if let pallet_ethereum::Call::transact { transaction } = call {
+					if let pallet_ethereum::Transaction::Legacy(legacy) = transaction {
+						// check if the transaction is DIA oracle update
+						if let pallet_ethereum::TransactionAction::Call(call_address) = legacy.action {
+							if call_address
+								== H160::from_slice(hex!("48ae7803cd09c48434e3fc5629f15fb76f0b5ce5").as_slice())
```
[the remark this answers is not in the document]
Roznovjak: I'm afraid that this is not possible, node gets the priority from runtime validate functions. But I will have a look, maybe there is some other way.

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,590 @@
...
+		header: B::Header,
+		best_block_hash: Arc<std::sync::Mutex<B::Hash>>,
+		http_client: HttpClient,
+		transaction_pool: Arc<P>,
+		thread_pool: Arc<Mutex<ThreadPool>>,
+		config: LiquidationWorkerConfig,
+	) {
+		// We can ignore the result, because it's not important for us.
+		// All we want is to have some upper bound for execution time of this task.
+		let _ = tokio::time::timeout(std::time::Duration::from_secs(8), async {
```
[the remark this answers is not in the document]
Roznovjak: ok, reduced to 4 secs

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,590 @@
...
+						// iterate over all borrowers
+						for borrower in sorted_borrowers_data_c.iter() {
+							let Ok(user_data) = UserData::new(&money_market_data, borrower.0, current_evm_timestamp, config.runtime_api_caller.unwrap_or(RUNTIME_API_CALLER))
+								else { return };
+
+							if let Ok(Some(liquidation_option)) = money_market_data
+								.get_best_liquidation_option(&user_data, config.target_hf.into(), (base_asset_address, price.into())) {
+
+								let (Some(collateral_asset_id), Some(debt_asset_id)) =
+									(HydraErc20Mapping::decode_evm_address(liquidation_option.collateral_asset),
```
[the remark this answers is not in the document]
Roznovjak: IIRC this is not going to work, because accessing the storage outside of "externalities" is not allowed.

pallets/liquidation/src/types.rs
```diff
@@ -0,0 +1,1394 @@
...
+use sp_core::H160;
+use sp_std::vec::Vec;
+
+pub type Balance = u128;
+pub type AssetId = u32;
+pub type CallResult = (ExitReason, Vec<u8>);
+
+pub const UNSIGNED_TXS_PRIORITY: u64 = 1_000_000;
+
+pub mod money_market {
```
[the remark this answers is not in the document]
Roznovjak: moved to a dedicated crate

liquidation-worker-support/src/lib.rs
```diff
@@ -0,0 +1,1364 @@
...
+	current_liquidity_rate: u128,
+	variable_borrow_index: u128,
+	current_variable_borrow_rate: u128,
+	// current_stable_borrow_rate: u128,
+	last_update_timestamp: u64,
+	// id: u16,
+	a_token_address: H160,
+	stable_debt_token_address: H160,
+	variable_debt_token_address: H160,
+	// interest_rate_strategy_address: H160,
```
dmoka, line 487: pls remove all commented reserve data properties if they are not needed.

liquidation-worker-support/src/lib.rs
```diff
@@ -0,0 +1,1364 @@
...
+	// accrued_to_treasury: u128,
+	// unbacked: u128,
+	// isolation_mode_total_debt: u128,
+}
+
+impl ReserveData {
+	pub fn new(data: &[ethabi::Token]) -> Option<Self> {
+		Some(Self {
+			configuration: data[0].clone().into_uint()?,
```
dmoka, line 496: maybe we could use .get(index) for these to be sure we never panic

liquidation-worker-support/src/lib.rs
```diff
@@ -0,0 +1,1364 @@
...
+			Ok(ReserveData::new(&decoded).ok_or(ethabi::Error::InvalidData)?)
+		} else {
+			Err(LiquidationError::EvmError(call_info.exit_reason))
+		}
+	}
+
+	/// Calls Runtime API.
+	fn fetch_asset_symbol(asset_address: &EvmAddress, caller: EvmAddress) -> Result<Vec<u8>, LiquidationError> {
+		let data = Into::<u32>::into(Function::Symbol).to_be_bytes().to_vec();
+		let gas_limit = U256::from(500_000);
```
dmoka, line 922: is this equal to TRADE_GAS_LIMIT? https://github.com/galacticcouncil/hydration-node/blob/89e1cfc425d8f5d8274b3075580ba3e5f47ee991/runtime/hydradx/src/evm/aave_trade_executor.rs#L115

If so I would use the constant somehow to make sure we don't forget to update this if the TRADE_GAS_LIMIT increases.

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,582 @@
...
+
+	/// Address of the Pool Address Provider contract.
+	#[clap(long)]
+	pub pap_contract: Option<EvmAddress>,
+
+	/// EVM address of the account that calls Runtime API. Account needs to have WETH balance.
+	#[clap(long)]
+	pub runtime_api_caller: Option<EvmAddress>,
+
+	/* TODO: would be nice if we can also configure oracle_update_caller and oracle_update_call_address,
```
dmoka, line 57: todo

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,582 @@
...
+	//
+	// /// EVM address of the DIA oracle update call address.
+	// #[clap(long)]
+	// pub oracle_update_call_address: Option<EvmAddress>,
+	/// Target health factor
+	#[clap(long, default_value_t = TARGET_HF)]
+	pub target_hf: u128,
+}
+
+// TODO: maybe use struct with data:
```
dmoka, line 72: todo

node/src/liquidation_worker.rs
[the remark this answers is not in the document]
dmoka: maybe we could create and use a runtime API for that?

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,582 @@
...
+								);
+							}
+						}
+					}
+				});
+			}
+		}).await;
+	}
+
+	// TODO: return Result type
```
dmoka, line 311: todo

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,582 @@
...
+	/// Fetch the preprocessed data used to evaluate possible candidates for liquidation.
+	async fn fetch_borrowers_data(http_client: HttpClient) -> Option<BorrowerData<AccountId>> {
+		let url = ("https://omniwatch.play.hydration.cloud/api/borrowers/by-health")
+			.parse()
+			.ok()?;
+		let res = http_client.get(url).await.ok()?;
+		assert_eq!(res.status(), StatusCode::OK);
```
dmoka, line 318: are we fine if this panics since we are in the node context?!

runtime/hydradx/src/lib.rs
```diff
@@ -529,7 +552,18 @@ impl_runtime_apis! {
...
 			block_hash: <Block as BlockT>::Hash,
 		) -> TransactionValidity {
-			Executive::validate_transaction(source, tx, block_hash)
+			let mut tx_validity = Executive::validate_transaction(source, tx.clone(), block_hash);
+			let transaction = tx.clone().0;
+			if let RuntimeCall::Liquidation(pallet_liquidation::Call::liquidate { .. }) = transaction.function {
+				// bump the priority only if the transaction is unsigned
+				if let Some(false) = tx.is_signed() {
+				 if let Ok(ref mut v) = tx_validity {
+					// TODO: priority multiplier
```
dmoka, line 561: todo

node/src/liquidation_worker.rs
Roznovjak: fixed

liquidation-worker-support/src/lib.rs
```diff
@@ -0,0 +1,1364 @@
...
+			gas_limit,
+			None,
+			None,
+			None,
+			true,
+			None,
+		)?;
+
+		if call_info.exit_reason == Succeed(Returned) {
+			Ok(U256::checked_from(&call_info.value[0..32])
```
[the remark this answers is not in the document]
Roznovjak: This smart contract returns uint256, so it's always 32 bytes.

liquidation-worker-support/src/lib.rs
```diff
@@ -0,0 +1,1364 @@
...
+				.checked_div(unit_price.into())
+				.ok_or::<LiquidationError>(ArithmeticError::Overflow.into())?
+				.try_into()
+				.map_err(|_| ArithmeticError::Overflow)?;
+		}
+
+		// adjust decimals
+		actual_debt_to_liquidate = if debt_decimals > oracle_price_decimals {
+			actual_debt_to_liquidate
+				.checked_mul(U256::from(10).pow((debt_decimals - oracle_price_decimals).into()))
```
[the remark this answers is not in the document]
Roznovjak: It's safe. Check the `if` statement:
```if debt_decimals > oracle_price_decimals```

File: liquidation-worker-support/src/lib.rs [hunk 18]
```diff
@@ -0,0 +1,1364 @@
...
+{
+	pap_contract: EvmAddress, // PoolAddressesProvider
+	pool_contract: EvmAddress,
+	oracle_contract: EvmAddress,
+	reserves: Vec<Reserve>, // the order of reserves is given by fetch_reserves_list()
+	pub caller: EvmAddress,
+	_phantom: PhantomData<(Block, Runtime)>,
+}
+impl<Block: BlockT, Runtime: EthereumRuntimeRPCApiV5<Block>> MoneyMarketData<Block, Runtime> {
+	/// Calls Runtime API.
```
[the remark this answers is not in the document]
Roznovjak: It's to inform that a method performs "outside" calls. These calls can also take longer to execute, but I'm not sure about this.

liquidation-worker-support/src/lib.rs
Roznovjak: We call multiple methods from different contracts, that's why there is no constant. I don't expect these limits (or gas required for execution) to change in the future. This would require deployment of new contracts.

liquidation-worker-support/src/lib.rs
```diff
@@ -0,0 +1,1364 @@
...
+use fp_evm::{ExitReason::Succeed, ExitSucceed::Returned};
+use fp_rpc::runtime_decl_for_ethereum_runtime_rpc_api::EthereumRuntimeRPCApiV5;
+use frame_support::sp_runtime::traits::{Block as BlockT, CheckedConversion};
+use hydradx_traits::evm::EvmAddress;
+use sp_arithmetic::ArithmeticError;
+use sp_core::{RuntimeDebug, H256, U256};
+use sp_std::{boxed::Box, ops::BitAnd};
+
+#[derive(RuntimeDebug)]
+pub enum LiquidationError {
```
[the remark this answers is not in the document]
Roznovjak: Then we would put 95% of the file into types.rs, because the main part of the crate are just types with methods.

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,590 @@
...
+            // We don't need to set a deadline, because it's wrapped in a task with a deadline.
+            let Some(borrowers_data) = Self::fetch_borrowers_data(http_client.clone()).await else {
+				tracing::error!(target: LOG_TARGET, "fetch_borrowers_data failed");
+				return
+			};
+            let sorted_borrowers_data = Self::process_borrowers_data(borrowers_data, config.max_liquidations);
+
+            // New transaction in the transaction pool
+            let mut notification_st = transaction_pool.clone().import_notification_stream();
+            while let Some(notification) = notification_st.next().await {
```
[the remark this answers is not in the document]
Roznovjak: Do you suggest to try liquidations also when there is no DIA update TX?

liquidation-worker-support/src/lib.rs
```diff
@@ -0,0 +1,1364 @@
...
+
+		let reserve = self
+			.reserves()
+			.get(reserve_index)
+			.ok_or(LiquidationError::ReserveNotFound)?;
+
+		let old_price = reserve.price();
+		self.update_reserve_price(new_price.0, new_price.1);
+
+		// TODO: continue if the price of callateral decreased/debt increased (the cases when HF decreases)
```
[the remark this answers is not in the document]
Roznovjak: This is more like an idea for improvement, maybe in the future version.

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,988 @@
...
+	fn current_timestamp(&self, hash: Block::Hash) -> Option<u64> {
+		let block = self.0.current_block(hash).ok()??;
+		// milliseconds to seconds
+		block.header.timestamp.checked_div(1_000)
+	}
+	fn call(
+		&self,
+		hash: Block::Hash,
+		caller: EvmAddress,
+		mm_pool: EvmAddress,
```
mrq1911: contract_address

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,988 @@
...
+		tx_waitlist: Arc<std::sync::Mutex<Vec<([u8; 8], <<B as BlockT>::Header as Header>::Number)>>>,
+		transaction_pool: Arc<P>,
+		thread_pool: Arc<Mutex<ThreadPool>>,
+		config: LiquidationWorkerConfig,
+	) {
+		let now = std::time::Instant::now();
+
+		// We can ignore the result, because it's not important for us.
+		// All we want is to have some upper bound for execution time of this task.
+		let _ = tokio::time::timeout(std::time::Duration::from_secs(4), async {
```
mrq1911, line 258: we should make sure there is no time where tx can be received and its not processed

mrq1911: we should always watch transactions

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,988 @@
...
+	#[allow(clippy::type_complexity)]
+	/// Executes when a new transaction is added to the transaction pool.
+	/// Listens to borrow and DIA oracle update transactions.
+	fn on_new_transaction(
+		notification: <P as TransactionPool>::Hash,
+		client: Arc<C>,
+		spawner: SpawnTaskHandle,
+		header: B::Header,
+		current_block_hash: B::Hash,
+		borrowers: Arc<std::sync::Mutex<Vec<(H160, U256)>>>,
```
mrq1911, line 339: this effectivelly causes all HF checks being done in 1 thread, more efficient would be to have lock per borrower so more jobs can update borrowers health factors concurrently

node/src/liquidation_worker.rs
```diff
@@ -0,0 +1,988 @@
...
+		let runtime_api = client.runtime_api();
+
+		let Some(current_evm_timestamp) = ApiProvider::<&C::Api>(runtime_api.deref()).current_timestamp(header.hash())
+		else {
+			tracing::debug!(target: LOG_TARGET, "fetch_current_evm_block_timestamp failed");
+			return;
+		};
+
+		// List of liquidated users in this block.
+		// We don't try to liquidate a user more than once in a block.
```
mrq1911, line 424: should be probably on block import level then as there can be multiple oracle updates received in the same block

mrq1911: the oracle updates should be aggregated instead than considered separately

mrq1911 (approved):

PR #1064, opened 2025-04-08
fix: chainlink adapter returns prices with 8 decimals

Roznovjak:


martinfridrich (approved):

runtime/hydradx/src/evm/precompiles/chainlink_adapter.rs
```diff
@@ -197,18 +193,13 @@ pub fn is_oracle_address(address: H160) -> bool {
...
+/// The fractional part contains 8 decimals.
+/// E.g. 7.1234 is stored as 712_340_000.
+fn normalize_price_to_u256(price: Price) -> Result<U256, PrecompileFailure> {
+	let decimals: usize = 8;
 
-	U256::exp10(decimals.into())
+	U256::exp10(decimals)
 		.checked_mul(price.n.into())
 		.ok_or(PrecompileFailure::Error {
 			exit_status: pallet_evm::ExitError::Other("Price conversion failed.".into()),
```
mrq1911, line 205: this cannot ever happen, you have to scale number and descale them safely

integration-tests/src/evm.rs
```diff
@@ -1840,6 +1851,75 @@ mod chainlink_precompile {
...
+					sender: Default::default(),
+					origin: Default::default(),
+				},
+				input,
+				100_000,
+			);
+			let dia_price = U256::from(output.as_slice());
+
+			//TODO: shouldn't be exactly same, but comparable within 5%
+			assert_eq!(dia_price, precompile_price);
```
mrq1911, line 1910: these numbers should be comparable within 5%, or atleast same order of magnitude :)

mrq1911 (changes requested):

runtime/hydradx/src/evm/precompiles/chainlink_adapter.rs
mrq1911: see https://github.com/galacticcouncil/hydration-node/blob/174ac5d13f0eca23b120b61ed57b854079d2f939/runtime/adapters/src/lib.rs#L571 for inspiration

PR #1066, opened 2025-04-11
fix: xyk pool existence check

dmoka:
fixes https://github.com/galacticcouncil/hydration-node/issues/1027


enthusiastmartin:
hm. i just looked at this. 

imo, the calculate functions should not be even called when the pool does not exist. it should not get that far imo

in this trait/struct, you already have "is_trade_supported" which calls the xyk::exists function, so why not used that.

dmoka:
> hm. i just looked at this.
> 
> imo, the calculate functions should not be even called when the pool does not exist. it should not get that far imo
> 
> in this trait/struct, you already have "is_trade_supported" which calls the xyk::exists function, so why not used that.

might be an option, the issue with this that everytime `calculate_in_given_out` is used somewhere, the developer should remember using the `is_treade_supported` before. So it would add this implicit knowledge, I would avoid.

Maybe calculate_in_given_out is not the best name for the purpose...

jak-pan (approved):

runtime/hydradx/src/assets.rs
```diff
@@ -1845,7 +1845,12 @@ impl SwappablePaymentAssetTrader<AccountId, AssetId, Balance> for XykPaymentAsse
 		asset_out: AssetId,
 		asset_out_amount: Balance,
 	) -> Result<Balance, DispatchError> {
-		let asset_pair_account = XYK::get_pair_id(AssetPair::new(insuff_asset_id, asset_out));
+		let asset_pair = AssetPair::new(insuff_asset_id, asset_out);
+		if !XYK::exists(asset_pair) {
+			return Err(DispatchError::Other("XYK pool does not exist for fee payment asset"));
```
mrq1911, line 1850: the text from other is dropped on the runtime, so please change some more descriptive error, ideally `TokenPoolNotFound` from XYK pallet

mrq1911 (changes requested):

mrq1911 (approved):

mrq1911:
fix lock and LGTM

PR #1067, opened 2025-04-11
feat: add evm call wrapper extrinsic to dispatcher pallet

### LINKED ISSUE #974: Create method that fails for reverted evm calls
Transactions that revert return ok exit code making it impossible to use batches. We should create a wrapper function that will fail in substrate.

F3Joule:
## Description
This PR introduces a new extrinsic in `pallet_dispatcher` to wrap a single EVM call. This is done in order to determine whether the EVM call exits successfully or either has a faulty exit status. This extrinsic is mostly going to be used in batch calls to revert state correctly (in case of `batchAll`); that's because EVM calls don’t fail when reverted, but return success with `ExitStatus::Revert`

## Related Issue
Fixes #974 

## Checklist:


pallets/dispatcher/src/lib.rs
```diff
@@ -169,6 +200,50 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			target: EvmAddress,
+			input: Vec<u8>,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+
+			let caller_evm_address = T::EvmAccounts::evm_address(&who);
+			let context = CallContext::new_call(target, caller_evm_address);
+
+			let (exit_reason, output) = T::Evm::call(context, input.clone(), U256::zero(), T::GasLimit::get());
```
mrq1911, line 228: please wrap `pallet_evm` call instead https://apidocs.bsx.fi/Hydration/pallets/EVM/extrinsic/call

with all of its parameters, instead of replicating parameters in the function, you should use call itself as the parameter

mrq1911 (changes requested):

pallets/dispatcher/src/lib.rs
```diff
@@ -236,6 +251,47 @@ pub mod pallet {
...
+		#[pallet::weight(call.get_dispatch_info().weight)]
+		pub fn dispatch_evm_call(
+			origin: OriginFor<T>,
+			call: Box<<T as Config>::RuntimeCall>,
+		) -> DispatchResultWithPostInfo {
+			ensure!(T::EvmCallIdentifier::is_evm_call(&call), Error::<T>::NotEvmCall);
+
+			let (result, actual_weight) = Self::do_dispatch(origin, *call);
+
+			match result {
```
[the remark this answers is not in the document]
F3Joule: yeah, at least the code will look more neat

pallets/dispatcher/src/lib.rs
```diff
@@ -236,6 +251,47 @@ pub mod pallet {
...
+		/// - `origin`: Signed origin.
+		/// - `call`: presumably `pallet_evm::Call::call` as boxed `RuntimeCall`.
+		///
+		/// Emits `EvmCallFailed` event when failed.
+		#[pallet::call_index(4)]
+		#[pallet::weight(call.get_dispatch_info().weight)]
+		pub fn dispatch_evm_call(
+			origin: OriginFor<T>,
+			call: Box<<T as Config>::RuntimeCall>,
+		) -> DispatchResultWithPostInfo {
```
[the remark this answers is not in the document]
mrq1911: yeah, just wrap it if you need it

pallets/dispatcher/src/lib.rs
```diff
@@ -271,4 +327,12 @@ impl<T: Config> Pallet<T> {
 			ExtraGas::<T>::set(new_value);
 		}
 	}
+
+	pub fn set_last_evm_call_failed(status: bool) {
+		if !status {
+			LastEvmCallFailed::<T>::kill();
+		} else {
+			LastEvmCallFailed::<T>::put(status);
+		}
```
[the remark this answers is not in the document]
enthusiastmartin: it is confusing to me... see my other comment to perhaps make it more readable ?

runtime/hydradx/src/evm/runner.rs
```diff
@@ -158,7 +160,16 @@ where
...
 			config,
-		)
+		)?;
+
+		let CallInfo { exit_reason, .. } = &result;
+		let call_succeeded = matches!(
+			exit_reason,
+			ExitReason::Succeed(ExitSucceed::Returned) | ExitReason::Succeed(ExitSucceed::Stopped)
+		);
+		pallet_dispatcher::Pallet::<T>::set_last_evm_call_failed(!call_succeeded);
```
mrq1911, line 170: why not just store `last_runner_result` and let whoever is reading that to deal with it?

pallets/dispatcher/src/lib.rs
```diff
@@ -101,6 +105,20 @@ pub mod pallet {
 	#[pallet::getter(fn extra_gas)]
 	pub type ExtraGas<T: Config> = StorageValue<_, u64, ValueQuery>;
 
+	#[pallet::storage]
+	#[pallet::whitelist_storage]
+	#[pallet::getter(fn last_evm_call_failed)]
+	pub type LastEvmCallFailed<T: Config> = StorageValue<_, bool, ValueQuery>;
```
enthusiastmartin: I find it little bit confusing..especially when you set this value in the runner. with all the negatives and invered values (` :set_last_evm_call_failed(!call_succeeded)` )

Also, the check that it matches the stopped/returned values in the runnner is not really needed there, as it is the purpose of this new dispatcher call.

I'd consider / suggest that you would set the exit reason instead of true/false.

and call LastEvmCallExitReason, perhaps you can add the value as well, so you can log it within this call as well.

And you simply check the returned or stopped in here, rather that in the runner ( as it is not its job really ). 

i would find more readable - you have the exit reason, you check what you need. - instead of calling set_failed with !call_success values which you need to do some mind effort to understand.

pallets/dispatcher/src/lib.rs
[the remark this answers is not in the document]
mrq1911: then you should make sure the value is killed on finalize, we have direct calls that fail all the time

we dont really need this storage to be persistent between blocks

pallets/dispatcher/src/lib.rs
mrq1911: i agree just, passing the last evm result seems cleaner and more useful for other usecases (like checking for outofgas)

pallets/dispatcher/src/lib.rs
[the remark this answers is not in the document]
F3Joule: The main reason why I didn't use `ExitReason` in this storage is because it doesn't derive `MaxEncodedLen`, which `StorageInfo` expects it to implement. And I don't really want to use `pallet::unbounded` for a storage, which will potentially have unbounded string (e.g. `ExitError::Other(Cow<'static, str>)`)

enthusiastmartin:
build is failiing, some tests, so i guess it still needs some work

pallets/dispatcher/src/lib.rs
```diff
@@ -236,6 +254,45 @@ pub mod pallet {
...
+		/// This extrinsic will fail if the EVM call returns any other ExitReason than `ExitSucceed(Returned)` or `ExitSucceed(Stopped)`.
+		/// Look the [hydradx_runtime::evm::runner::WrapRunner] implementation for details.
+		///
+		/// Parameters:
+		/// - `origin`: Signed origin.
+		/// - `call`: presumably `pallet_evm::Call::call` as boxed `RuntimeCall`.
+		///
+		/// Emits `EvmCallFailed` event when failed.
+		#[pallet::call_index(4)]
+		#[pallet::weight(call.get_dispatch_info().weight)]
```
enthusiastmartin: i believe this should include also the overhead of this extrinsic, not only the dispatch call.

mrq1911 (changes requested):
ensure that evm wallet can dispatch this as well unlike regular evm::call

pallets/dispatcher/src/lib.rs
```diff
@@ -271,4 +328,12 @@ impl<T: Config> Pallet<T> {
 			ExtraGas::<T>::set(new_value);
 		}
 	}
+
+	pub fn set_last_evm_call_failed(status: bool) {
```
enthusiastmartin: This is no longer used anywhere!

enthusiastmartin (changes requested):

enthusiastmartin (approved):

runtime/hydradx/src/evm/mod.rs
```diff
@@ -139,14 +141,35 @@ parameter_types! {
...
+where
+	M: AddressMapping<AccountId32>,
+	OuterOrigin: Into<Result<RawOrigin<AccountId32>, OuterOrigin>> + From<RawOrigin<AccountId32>>,
+{
+	type Success = AccountId32;
+
+	fn try_address_origin(address: &sp_core::H160, origin: OuterOrigin) -> Result<AccountId32, OuterOrigin> {
+		origin.into().and_then(|o| match o {
+			RawOrigin::Signed(who)
+				if who == M::into_account_id(*address) || AsRef::<[u8; 32]>::as_ref(&who)[0..20] == address[0..20] =>
```
mrq1911: potencially only `truncated_account_id(address)` would be needed here, bound accounts should be already covered

mrq1911 (approved):

runtime/hydradx/src/evm/mod.rs
mrq1911: also would be nice to have some test for this case

PR #1078, opened 2025-04-24
fix: DCA expecting exact amount of aTokens received

dmoka:
in aave sell amount calculation,we dont account for rounding currently.

This PR fixes that.

Some failing DCA: https://hydration.subscan.io/block/7396904?tab=event&event=7396904-31

### DIFF
--- a/runtime/hydradx/src/evm/aave_trade_executor.rs
+++ b/runtime/hydradx/src/evm/aave_trade_executor.rs
@@ -387,13 +387,18 @@ where
 
 	fn calculate_in_given_out(
 		pool_type: PoolType<AssetId>,
-		asset_in: AssetId,
-		asset_out: AssetId,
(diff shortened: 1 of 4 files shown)


runtime/hydradx/src/evm/aave_trade_executor.rs
```diff
@@ -387,8 +391,9 @@ where
 		asset_out: AssetId,
 		amount_out: Balance,
 	) -> Result<Balance, ExecutorError<Self::Error>> {
+		let buffer = AAVE_ROUNDING_BUFFER.saturating_mul(2);
 		Self::calculate_out_given_in(pool_type, asset_in, asset_out, amount_out)
```
mrq1911, line 395: dont reuse the calc out given in amymore pls

mrq1911 (changes requested):

PR #1087, opened 2025-05-01
feat: dispatch with extra gas

enthusiastmartin:
This pr adds a new dispatch call which can be used if user needs to purchase some extra gas for more expnesive calls to evm. 

The way it works:
- use calls dispatch_with_extra_gas and pays for the given extra gas he wants to use
- dispatcher pallet stores this information into temporary storage
- when it comes to evm execution - exeutor reads from this storage and adds this extra gas to current GAS_LIMIT
- after the execution - we calculate how much of the extr gas was used and subtract it from the sotred value - the reason is that the remainig extra gas can be used for subsequant evm calls ( eg. if in batch )
- at the end - stored extra gas is cleared from the storage
- extra gas is not refunded.


pallets/dispatcher/src/lib.rs
```diff
@@ -91,6 +100,10 @@ pub mod pallet {
 	#[pallet::getter(fn aave_manager_account)]
 	pub type AaveManagerAccount<T: Config> = StorageValue<_, T::AccountId, ValueQuery, T::DefaultAaveManagerAccount>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn account_extra_gas)]
+	pub type AccountExtraGas<T: Config> = StorageMap<_, Blake2_128Concat, EvmAddress, u64, OptionQuery>;
```
mrq1911, line 105: no need for map, can be just one value shared

runtime/hydradx/src/evm/executor.rs
```diff
@@ -52,15 +53,28 @@ where
...
-			executor.transact_call(context.sender, context.contract, value, data, gas, vec![])
+		let extra_gas = pallet_dispatcher::Pallet::<T>::get_account_extra_gas(&context.sender);
+		let gas_limit = gas.saturating_add(extra_gas);
+		log::trace!(target: "evm::executor", "Call with extra gas {:?}", extra_gas);
+
+		Self::execute(context.origin, gas_limit, |executor| {
+			let result = executor.transact_call(context.sender, context.contract, value, data, gas_limit, vec![]);
+			log::trace!(target: "evm::executor", "Call executed - used gas {:?}", executor.used_gas());
+
+			if extra_gas > 0 {
```
mrq1911, line 71: view() should have this as well

runtime/hydradx/src/weights/pallet_dispatcher.rs
```diff
@@ -90,4 +90,9 @@ impl<T: frame_system::Config> pallet_dispatcher::WeightInfo for HydraWeight<T> {
         Weight::from_parts(3_864_000, 0)
             .saturating_add(T::DbWeight::get().writes(1_u64))
     }
+
+    fn dispatch_with_extra_gas(n: u32) -> Weight {
+        Weight::from_parts(3_864_000, 0)
+            .saturating_add(T::DbWeight::get().writes(1_u64))
```
mrq1911, line 96: is there really a write?

mrq1911 (changes requested):

runtime/hydradx/src/weights/pallet_dispatcher.rs
enthusiastmartin: no, i dont thin so - i havent yet benchmarked, so this is just a random copy pasta from another extrinaic. will update

pallets/dispatcher/src/lib.rs
enthusiastmartin: hm. that;s true

enthusiastmartin: and then i dont have "issues" with address mappings, as i dont need any adresss, and Ican get rid of some additional requirements due to the AsRef

much better

enthusiastmartin: done

runtime/hydradx/src/weights/pallet_dispatcher.rs
enthusiastmartin: updated weights - there seems to be a 1 write. but looking at the dispatch functions - they all do one 1 write.

pallets/dispatcher/src/benchmarking.rs
```diff
@@ -53,5 +53,14 @@ benchmarks! {
 	note_aave_manager {
 	}: _(RawOrigin::Root, Pallet::<T>::aave_manager_account())
 
+	dispatch_with_extra_gas{
+		let n in 1 .. 10_000;
+		let remark = sp_std::vec![1u8; n as usize];
+
+		let call: <T as pallet::Config>::RuntimeCall = frame_system::Call::remark { remark }.into();
```
dmoka, line 60: what is the reason that we benchmark it with just a remark call?

I would comment that reason in the code.

enthusiastmartin: we dont want to benchmark any specific call, because it would be included the actual weight.

remark does nothing. 

 the call weight is calculated separately for each call that you want to dispatch. and added to weight

pallets/dispatcher/src/lib.rs
```diff
@@ -169,6 +177,37 @@ pub mod pallet {
...
+			call: Box<<T as Config>::RuntimeCall>,
+			extra_gas: u64,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+
+			// Store the gas limit for this account
+			ExtraGas::<T>::set(extra_gas);
+
+			// Dispatch the call
+			let (result, _) = Self::do_dispatch(who.clone(), *call);
```
mrq1911, line 205: please test that this actually returns correct weight back and that you pay for extra gas and that unused extra gas is refunded as weight. seems to me that when you do it this way you dont actually pay for extra gas as inner call ahs no notion of it

mrq1911, line 205: dont recreate origin, reuse the original one (pass origin, not who)

mrq1911 (changes requested):

pallets/dispatcher/src/lib.rs
enthusiastmartin: yes. i was wondering the same why it is not like that.

the do_dispatch - i only used already existing one. 

i might rework this

mrq1911: makes sense for governance stuff as you actually are changing origin to some signed address

pallets/dispatcher/src/lib.rs
```diff
@@ -169,6 +178,62 @@ pub mod pallet {
...
+			T::WeightInfo::dispatch_with_extra_gas(call_len)
+				.saturating_add(call_weight)
+				.saturating_add(gas_weight)
+		})]
+		pub fn dispatch_with_extra_gas(
+			origin: OriginFor<T>,
+			call: Box<<T as Config>::RuntimeCall>,
+			extra_gas: u64,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
```
mrq1911, line 200: dont ensure signed, just pass whatever origin there is

pallets/dispatcher/src/lib.rs
```diff
@@ -169,6 +178,62 @@ pub mod pallet {
...
+
+			let remaining_extra_gas = ExtraGas::<T>::take();
+			let gas_used = extra_gas.saturating_sub(remaining_extra_gas);
+
+			// if gas used is zero, we can return the result directly
+			if gas_used == 0 {
+				return result;
+			}
+			// we can return result directly also when the actual weight is None ( for some reason)
+			let Some(actual_weight) = actual_weight else {
```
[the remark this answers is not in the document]
mrq1911: i would like to avoid scenario where you can fill block by just setting insane extra gas and then not paying it

pallets/dispatcher/src/lib.rs
```diff
@@ -169,6 +178,62 @@ pub mod pallet {
...
+				return result;
+			};
+			//Otherwise, we need to account for the extra gas used.
+			let extra_weight = T::GasWeightMapping::gas_to_weight(gas_used, true);
+			let actual_weight = Some(actual_weight.saturating_add(extra_weight));
+
+			match result {
+				Ok(call_post_info) => Ok(PostDispatchInfo {
+					actual_weight,
+					pays_fee: call_post_info.pays_fee,
```
mrq1911, line 226: i think you should always pay extra gas, inner tx can waive the fee, but it shouldnt be able to do it for wrapper as well

mrq1911 (changes requested):

pallets/dispatcher/src/lib.rs
enthusiastmartin: changed to Pays:Yes

pallets/dispatcher/src/lib.rs
enthusiastmartin: reworked

Issue #858, opened 2024-07-05
staking - clear position votes storage if no votes


enthusiastmartin:
Currently we keep storage with empty list of votes for a position. 

We can remove this entry if there are no votes for the position.

Yung-Beef:
I'm working on this.

Yung-Beef:
Here is the [PR](https://github.com/galacticcouncil/hydration-node/pull/1088).

PR #1088, opened 2025-05-02
fix(staking): clear votes storage for a position if that position has no votes

Yung-Beef:
## Description
When a vote is removed with `on_remove_vote()`, if the function removes the last vote for that position id, that position id's entry will be removed from the `Votes` storage.

This change was added for legacy democracy as well, but can be removed if not needed.

I tried to make it cleaner by using `mutate_exists()`, but as far as I can tell that method is not possible due to `Votes` using ValueQuery.

Also some misc typos.

## Related Issue
[staking - clear position votes storage if no votes #858](https://github.com/galacticcouncil/hydration-node/issues/858)

## Motivation and Context
This change will result in a smaller state (for the staking pallet), slightly reducing storage requirements and improving performance of the staking pallet.

## How Has This Been Tested?
A new test was added (`on_remove_vote()` is not used in any tests within the staking pallet) with two votes, and checked that the Votes storage was updated correctly as they were removed one by one, and that when the last vote was removed, the storage did not contain the key for that position id. A test was not written for legacy democracy.

All 55 staking tests pass.

## Checklist:


pallets/staking/src/integrations/conviction_voting.rs
```diff
@@ -122,6 +122,10 @@ where
 					Votes::<T>::mutate(position_id, |voting| {
 						voting.votes.remove(vote_idx);
 					});
+					// Clear the entry from Votes if positionid has no more votes stored
+					if Votes::<T>::get(position_id).votes.last_mut() == None {
```
enthusiastmartin, line 126: you could use .is_empty?

Yung-Beef:
Thanks for the suggestion @enthusiastmartin, I'd never written one of those before so hadn't thought about it.

I created the one you suggested, but couldn't think of any other scenarios that might matter (probably due to lack of knowledge of the the greater system as a whole). The only other thing in the pallet that touches `Votes` is `balance_locked_on_unsuccessful_vote()` which doesn't seem to be called from anywhere in the entire repo.

Yung-Beef:
Thanks for the help, should be good now.

martinfridrich (changes requested):
What is an issue with `mutate_exists`? I didn't try it but why something like this doesn't work?

```rust 
Votes::<T>::mutate_exists(position_id, |voting|{
  if let Some(v) = voting {
    v.votes.remove(vote_idx);
    if v.votes.is_empty() {
      *voting = None;
    }
  }
});
```

integration-tests/src/staking.rs
```diff
@@ -2107,6 +2107,78 @@ fn remove_vote_should_extend_lock_when_votes_are_already_processed() {
...
+		assert_ok!(ConvictionVoting::remove_vote(
+			hydradx_runtime::RuntimeOrigin::signed(ALICE.into()),
+			Some(ROOT_TRACK),
+			r
+		));
+
+		// This call shouldn't return anything, but the storage uses ValueQuery so the default is returned,
+		// but we can check it's empty. Staking pallet tests show the storage is indeed cleared.
+		let stake_voting = pallet_staking::Pallet::<hydradx_runtime::Runtime>::get_position_votes(stake_position_id);
+		assert!(stake_voting.votes.is_empty());
```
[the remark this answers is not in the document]
Yung-Beef: The storage is already pub, but I can't figure out how to reference it from in the test, I just keep getting errors about ambiguous types.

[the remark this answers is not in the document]
martinfridrich: This should work `pallet_staking::Votes::<hydradx_runtime::Runtime>::contains_key(stake_position_id)` but you have to make storage `pub` not `pub(super)`

pallets/staking/src/integrations/conviction_voting.rs
```diff
@@ -119,8 +119,14 @@ where
 							position.action_points = position.action_points.saturating_add(points);
 						}
 					}
-					Votes::<T>::mutate(position_id, |voting| {
-						voting.votes.remove(vote_idx);
+					Votes::<T>::mutate_exists(position_id, |voting| {
+						if let Some(v) = voting {
+							v.votes.remove(vote_idx);
+							// Clear the entry from Votes if positionid has no more votes stored
```
martinfridrich, line 125: This is not blocker for me but this comment just says what code does, these kind of comments are not helpful IMO

pallets/staking/src/integrations/conviction_voting.rs
```diff
@@ -119,8 +119,14 @@ where
 							position.action_points = position.action_points.saturating_add(points);
 						}
 					}
-					Votes::<T>::mutate(position_id, |voting| {
-						voting.votes.remove(vote_idx);
+					Votes::<T>::mutate_exists(position_id, |voting| {
```
martinfridrich, line 122: Please do the same also for old democracy.

Yung-Beef:
Fixed @martinfridrich

enthusiastmartin:
> I got that fixed, but for some reason pulling from the upstream branch added all of these other files getting changed, not sure why
> 
> But on my branch on github it doesn't say it's any commits behind yours, so hopefully that's fine
> 
> Otherwise lmk if there's some way to transfer ownership of the branch or repo to you and you can figure it out cause idk

there seems to a lot of unrelated changes to files which should not be changed. 

this is perhaps result of merge conflicts and how it has been resovled,

i would suggest to clone the latest master, and do the code changes again on the latest/correct master version. And create new PR, unless yo uwant to resolve all these unrealted changes manually one by one.

enthusiastmartin:
closing due to inactivity.

as my last comment stated - there are many changes to unrelated files, which makes it hard to review. 

The change is pretty simple, so it should not requires that many files to be changed. 

suggesting to rework thsi as another PR.

PR #1094, opened 2025-05-08
fix: invert back the stableswap spot price

### LINKED ISSUE #1009: Bug: fix spot price calculation of stableswap
Stableswap spot price calculation returns the result in reserved, so we should inverse it to be consistent with other APIs.

https://github.com/galacticcouncil/hydration-node/blob/ea4e622fe5adb0dda12dd4000ac6b55849ff7649/math/src/stableswap/math.rs#L771

dmoka:
Fixes -> https://github.com/galacticcouncil/hydration-node/issues/1009


pallets/stableswap/src/trade_execution.rs
```diff
@@ -322,6 +322,8 @@ where
 					Some(trade_fee),
 					&asset_pegs,
 				)
+				.ok_or_else(|| ExecutorError::Error(ArithmeticError::Overflow.into()))?
+				.reciprocal()
```
enthusiastmartin, line 326: wouldnt be be better to just swap the asset as input to the calculate spot price function ?

i believe it would be more clear - this reciprocal call is bit lost here imo

dmoka: done

enthusiastmartin (changes requested):
the change/fix looks ok.

however,  i do have question why the need to change various values in the tests ? 

eg. fees, issuance etc.

i would expect the tests not to change.

and it needs to resolve the conflicts.

dmoka:
> the change/fix looks ok.
> 
> however, i do have question why the need to change various values in the tests ?
> 
> eg. fees, issuance etc.
> 
> i would expect the tests not to change.
> 
> and it needs to resolve the conflicts.

agree, reverted all those changes

pallets/hsm/src/lib.rs
```diff
@@ -1485,8 +1485,6 @@ where
 				// just to be safe
 				return None;
 			}
-			let after_spot = FixedU128::one().div(after_spot);
```
enthusiastmartin, line 1529: i am not still sure about this one.

this changes the implementation compared to python. 

let me look into it,

math/src/stableswap/math.rs
```diff
@@ -902,6 +902,7 @@ pub fn calculate_spot_price(
 				fee,
 				pegs,
 			)
+			.map(|p| p.reciprocal())?
```
enthusiastmartin: we should fix the underlaying function rather than inverting the price here.

pallets/stableswap/src/trade_execution.rs
enthusiastmartin, line 326: i think this is an issue?!

dmoka:
Given the effort we already put, the little risk and also not much value to gain, we close this issue.

PR #1097, opened 2025-05-13
feat: unsigned liquidation call

Roznovjak:


pallets/liquidation/src/lib.rs
```diff
@@ -129,13 +129,45 @@ pub mod pallet {
...
+					// receiving unsigned transaction from network - disallow
+					return InvalidTransaction::Call.into();
+				}
+				TransactionSource::Local => {}   // produced by off-chain worker
+				TransactionSource::InBlock => {} // some other node included it in a block
+			};
+
+			let valid_tx = |provide| {
+				ValidTransaction::with_tag_prefix("liquidate_unsigned_call")
+					.priority(1_000_000)
```
jak-pan, line 151: why not max?

mrq1911: oracle update has to fit before, this should be outside of range any user can do already

mrq1911 (approved):

PR #1103, opened 2025-05-16
fix: HSM - prevent hollar as collateral

enthusiastmartin:


pallets/hsm/src/tests/add_collateral_tests.rs
```diff
@@ -562,3 +562,54 @@ fn add_collateral_should_fail_when_max_is_reached() {
...
+		.with_collateral(
+			DAI,
+			100,
+			Permill::from_percent(1),
+			FixedU128::one(),
+			Permill::from_percent(1),
+		)
+		.build()
+		.execute_with(|| {
+			// Try to add DAI as collateral again
```
dmoka, line 600: Is this comment correct? it seems we add HOLLAR, instead of DAI

dmoka (approved):
lgtm

PR #1106, opened 2025-05-16
fix: HSM - add flash mint facilitator

enthusiastmartin:


runtime/hydradx/src/weights/pallet_hsm.rs
```diff
@@ -222,4 +222,8 @@ impl<T: frame_system::Config> pallet_hsm::WeightInfo for HydraWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(1_u64))
 			.saturating_add(T::DbWeight::get().writes(1_u64))
 	}
+
+	fn set_flash_mint_facilitator() -> Weight {
+		Weight::zero()
```
dmoka, line 227: benchmarking is still left

dmoka, line 227: this is still not updated, I guess we wait for the benchmark build, but leaving this comment for a todo

enthusiastmartin: no longer relevant

pallets/hsm/src/lib.rs
```diff
@@ -785,7 +788,9 @@ pub mod pallet {
 			if hollar_amount_to_trade > 0 {
 				let hsm_account = Self::account_id();
 
-				Self::mint_hollar(&hsm_account, hollar_amount_to_trade)?;
+				// If flash minter not set, we use HSM facilitator.
+				let flash_minter = T::FlashMintFacilitator::get();
+				Self::mint_hollar(&hsm_account, hollar_amount_to_trade, flash_minter)?;
```
mrq1911, line 793: this is not how flashloan works, you have to wrap you call into it, flashloan itself makes sure that the loan is burned in the end

see impl https://github.com/aave/gho-core/blob/397336cc1cc6d27a28a82564443a77ba049d3a48/src/contracts/facilitators/flashMinter/GhoFlashMinter.sol#L68

mrq1911 (changes requested):

enthusiastmartin:
closing in favor of #1114

PR #1111, opened 2025-05-22
fix: missing EVM logs

dmoka:
EVM logs were missing since we just constructed the EVM executor ourselves, which doesn't include emitting the EVM events.

By using the EvmRunner, to be exact, our WrappedRunner, we can trigger the EVM event logs.

Only EVM::call has been changed, EVM::view remains intact as there we don't care much about the events.

### DIFF
--- a/runtime/hydradx/src/evm/executor.rs
+++ b/runtime/hydradx/src/evm/executor.rs
@@ -107,7 +107,7 @@ where
 			None,               // max_priority_fee_per_gas
 			None,               // nonce
 			vec![],
-			false, // is_transactional - we dont need to check for  EIP-3607, and it also makes the total_fee_per_gas to default (0)
+			false, // is_transactional - we dont need to check for  EIP-3607, and it also makes the payed fee zeo
 			false, // validate
 			None,  // weight_limit
 			None,  // proof_size_base_cost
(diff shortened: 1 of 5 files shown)


runtime/hydradx/src/evm/executor.rs
```diff
@@ -1,25 +1,53 @@
...
 pub type BalanceOf<T> =
 	<<T as pallet_evm::Config>::Currency as frame_support::traits::Currency<pallet_evm::AccountIdOf<T>>>::Balance;
 pub type NonceIdOf<T> = <<T as Config>::AccountProvider as AccountProvider>::Nonce;
 
+pub struct Executor<R>(sp_std::marker::PhantomData<R>);
+
+type EVMRunner = WrapRunner<
+	Runtime,
+	pallet_evm::runner::stack::Runner<Runtime>, // Evm runner that we wrap
+	hydradx_adapters::price::FeeAssetBalanceInCurrency<
```
mrq1911, line 43: do we really need it if we dont use it? cannot we just have some default impl there or do we even have to wrap it?

mrq1911: or why dont we use runner we already wrapped for runtime?

mrq1911 (approved):

PR #1120, opened 2025-06-05
feat: HSM router support

enthusiastmartin:
This Pr adds router support for HSM, so it is possible to trade hollar via router

It also refactor HSM pallet to extract math functions into math package - in prepartion for wasm support

### DIFF
--- a/integration-tests/src/hsm.rs
+++ b/integration-tests/src/hsm.rs
@@ -1,6 +1,6 @@
 use crate::polkadot_test_net::hydra_live_ext;
 use crate::polkadot_test_net::hydradx_run_to_next_block;
-use crate::polkadot_test_net::{TestNet, ALICE, BOB};
+use crate::polkadot_test_net::{TestNet, ALICE, BOB, HDX};
 use fp_evm::ExitSucceed::Returned;
 use fp_evm::{ExitReason::Succeed, ExitSucceed::Stopped};
 use frame_support::assert_ok;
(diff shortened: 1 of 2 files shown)


pallets/hsm/src/trade_execution.rs
```diff
@@ -0,0 +1,207 @@
...
+			_ => Err(ExecutorError::NotSupported),
+		}
+	}
+
+	fn calculate_spot_price_with_fee(
+		pool_type: PoolType<T::AssetId>,
+		asset_a: T::AssetId,
+		asset_b: T::AssetId,
+	) -> Result<FixedU128, ExecutorError<Self::Error>> {
+		match pool_type {
```
dmoka, line 191: this function is pretty much just delegation, but i would write some unit tests to ensure it really works as we expect.

pallets/hsm/src/trade_execution.rs
```diff
@@ -0,0 +1,207 @@
...
+					return Err(ExecutorError::Error(Error::<T>::InvalidAssetPair.into()));
+				}
+				let amount_in = if asset_in == T::HollarId::get() {
+					// Calculate Hollar amount in given collateral asset out
+					|| -> Result<Balance, DispatchError> {
+						let collateral_info = Collaterals::<T>::get(asset_out).ok_or(Error::<T>::AssetNotApproved)?;
+						let pool_state = Self::get_stablepool_state(collateral_info.pool_id)?;
+						let hollar_amount = Self::simulate_out_given_in(
+							collateral_info.pool_id,
+							asset_out,
```
dmoka, line 94: Should not we need to swap these params?

As the 2nd param of `simulate_out_given_in` is named as `asset_in`, the 3rd is `assset_out`. But here I read it the other way around.

integration-tests/src/hsm.rs
```diff
@@ -11,7 +11,7 @@ use hydradx_runtime::{
 		precompiles::{handle::EvmDataWriter, Bytes},
```
dmoka, line 11: will hsm route type be used with other route types in one route? I mean, can we have routes like
assetA -> assetB (Omnipool)
AssetB -> assetC (HSM)

or similar? If so I would write an integration test, to document this usecase.

pallets/hsm/src/trade_execution.rs
enthusiastmartin: no, it is correct = it is the way it works. it might look confusing but it is fine.

pallets/hsm/src/trade_execution.rs
enthusiastmartin: tested using integration tests.

integration-tests/src/hsm.rs
enthusiastmartin: added

pallets/hsm/src/trade_execution.rs
dmoka: since generally it doesn't make sense to pass `asset_out` in param named `asset_in`, I would leave a comment WHY you did that, helping future maintainers.

enthusiastmartin: we simulate the opposite operation, so in sell we simulate buying stuff so we can figure out the execution price.  it is the way hsm works - this is just doiing same stuff as we do in the sell/buy

PR #1125, opened 2025-06-11
feat: add runtime fuzzer

enthusiastmartin:
This PR moves our runtime-fuzzer into this hydration monorepo.

The reason is that it is easier to maintain the dependencies, fuzzer also requires runtime code, so it makes sense to have it here.

The fuzzer is excluded from the build, as it requires nightly rust. it has its own toolchain file in the subdirectory.

This PR also pins frontier version to specific version


enthusiastmartin:
we keep fuzzer in standalone repo

PR #1129, opened 2025-06-12
feat: declare precompiles using builder

dmoka:
- use builder to declare precompiles as it is more declarative and has better checks
- set nesting to 0 for dispatch to prevent reentrancy attack

### DIFF
--- a/integration-tests/src/evm.rs
+++ b/integration-tests/src/evm.rs
@@ -8,9 +8,7 @@ use frame_support::traits::fungible::Mutate;
 use frame_support::{assert_ok, dispatch::GetDispatchInfo, sp_runtime::codec::Encode, traits::Contains};
 use frame_system::RawOrigin;
 use hex_literal::hex;
-use sp_core::bounded_vec::BoundedVec;
-
-use hydradx_runtime::evm::precompiles::DISPATCH_ADDR;
+use hydradx_runtime::evm::precompiles::DispatchAddress;
 use hydradx_runtime::evm::EvmAddress;
 use hydradx_runtime::evm::ExtendedAddressMapping;
 use hydradx_runtime::evm::Function;
@@ -29,6 +27,8 @@ use orml_traits::MultiCurrency;
 use pallet_evm::*;
 use pretty_assertions::assert_eq;
 use primitives::{AssetId, Balance};
+use sp_core::bounded_vec::BoundedVec;
+use sp_core::Get;
 use sp_core::{blake2_256, H160, H256, U256};
 use sp_runtime::TransactionOutcome;
 use sp_runtime::{traits::SignedExtension, DispatchError, FixedU128, Permill};
@@ -41,6 +41,7 @@ mod account_conversion {
 	use super::*;
 	use fp_evm::ExitSucceed;
 	use frame_support::{assert_noop, assert_ok};
+	use hydradx_runtime::evm::precompiles::DispatchAddress;
 	use pretty_assertions::assert_eq;
 
 	#[test]
@@ -119,7 +120,7 @@ mod account_conversion {
 			assert_ok!(EVM::call(
 				evm_signed_origin(evm_address),
 				evm_address,
-				DISPATCH_ADDR,
+				DispatchAddress::get(),
 				data,
 				U256::from(0),
 				1000000,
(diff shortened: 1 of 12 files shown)


runtime/hydradx/src/evm/precompiles/mod.rs
```diff
@@ -76,40 +77,40 @@ impl From<Address> for H160 {
...
+		pub struct $name;
+		impl Get<H160> for $name {
+			fn get() -> H160 {
+				$addr
+			}
+		}
+	};
 }
 
+def_address_getter!(ECRecoverAddress, H160(hex!("0000000000000000000000000000000000000001")));
```
Roznovjak: Personally, I don't like macros that define structs. Generated structs are invisible for IDEs. If we need to use something that implements `Get` trait, we could leave constants for `pub const SHA256: H160 = H160(hex!("0000000000000000000000000000000000000002"));` and use this struct for places where `Get` needs to be implemented ```pub struct AddressU64<const N: u64>;
impl<const N: H160> Get<H160> for AddressU64<N> {
	#[inline(always)]
	fn get() -> H160 {
		N
	}
}```
What do you think about it?

Roznovjak:
```
pub struct AddressU64<const N: u64>; 
impl<const N: H160> Get<H160> for AddressU64<N> 
{ 
   #[inline(always)] 
   fn get() -> H160 
   { 
      N 
   } 
}
```

integration-tests/src/evm.rs
```diff
@@ -2783,3 +2796,156 @@ fn create_xyk_pool_with_amounts(asset_a: u32, amount_a: u128, asset_b: u32, amou
...
+		data.extend_from_slice(&64u128.to_be_bytes()[4..]);
+		// length of bytes data
+		data.extend_from_slice(&[0u8; 32]);
+
+		let context = CallContext {
+			contract: contract,
+			sender: crate::contracts::deployer(),
+			origin: crate::contracts::deployer(),
+		};
+		let (res, data) = Executor::<Runtime>::call(context, data, U256::zero(), 100_000);
```
talhadaar, line 2885: res here is entirely unused so maybe replace with a underscore(_)?

integration-tests/src/evm_permit.rs
```diff
@@ -30,6 +30,7 @@ use pallet_transaction_multi_payment::EVMPermit;
 use pretty_assertions::assert_eq;
 use primitives::constants::currency::UNITS;
 use primitives::{AssetId, Balance};
+use sp_core::Get;
 use sp_core::{H256, U256};
```
talhadaar, line 34: This will need a round of cargo fmt and clippy fixes, reduce redundancies and make stuff more readable

runtime/hydradx/src/evm/precompiles/chainlink_adapter.rs
```diff
@@ -213,10 +218,12 @@ where
 	}
 }
 
-pub fn is_oracle_address(address: H160) -> bool {
-	let oracle_address_prefix = &(H160::from(hex!("0000010000000000000000000000000000000000"))[0..3]);
+pub fn oracle_address_prefix() -> Vec<u8> {
+	H160::from(hex!("0000010000000000000000000000000000000000"))[0..3].to_vec()
```
talhadaar, line 222: Maybe think about evaluating this manually and keeping this as a constant somewhere in this file rather than deriving each time? In my opinion, that's always easier to maintain and much cleaner, however depending on the usage this should be meh

runtime/hydradx/src/evm/precompiles/erc20_mapping.rs
```diff
@@ -93,10 +94,12 @@ impl Erc20Encoding<AssetId> for HydraErc20Mapping {
 	}
 }
 
-pub fn is_asset_address(address: H160) -> bool {
-	let asset_address_prefix = &(H160::from(hex!("0000000000000000000000000000000100000000"))[0..16]);
+pub fn asset_address_prefix() -> Vec<u8> {
+	H160::from(hex!("0000000000000000000000000000000100000000"))[0..16].to_vec()
```
talhadaar, line 98: Same comment here as above

dmoka:
Closed as we discussed that the refactoring touches core layer of ethereum transaction, involving risks, with little gain.

Issue #1022, opened 2025-02-08
Make WETH<>HDX price in EVM based on oracle


jak-pan:
https://github.com/galacticcouncil/hydration-node/blob/62288711f7f83753374d604d465e120bde287c04/pallets/dynamic-evm-fee/src/lib.rs#L70

dmoka:
For the reference price, I will use the 10mins oracle and compare the current (short oracle) price with it. If the price is increasing we will increase the eth fee accordingly

PR #1134, opened 2025-06-19
fix: use oracle reference price for dynamic evm fee

dmoka:
Fixes https://github.com/galacticcouncil/hydration-node/issues/1022

### DIFF
--- a/pallets/dynamic-evm-fee/src/lib.rs
+++ b/pallets/dynamic-evm-fee/src/lib.rs
@@ -68,8 +68,6 @@ use sp_core::U256;
 use sp_runtime::FixedPointNumber;
 use sp_runtime::FixedU128;
 
-pub const ETH_HDX_REFERENCE_PRICE: FixedU128 = FixedU128::from_inner(8945857934143137845); //Current onchain ETH price on at block #4,534,103
-
 #[frame_support::pallet]
 pub mod pallet {
 	use crate::*;
@@ -100,6 +98,9 @@ pub mod pallet {
 
 		/// Native price oracle
 		type NativePriceOracle: NativePriceOracle<Self::AssetId, EmaPrice>;
+
+		/// Referene Native price oracle
+		type ReferenceNativePriceOracle: NativePriceOracle<Self::AssetId, EmaPrice>;
 
 		/// WETH Asset Id
 		#[pallet::constant]
@@ -147,10 +148,20 @@ pub mod pallet {
 					return;
 				};
 
-				let Some(price_diff) =
-					FixedU128::checked_from_rational(eth_hdx_price.into_inner(), ETH_HDX_REFERENCE_PRICE.into_inner())
+				let Some(eth_hdx_reference_price) = T::ReferenceNativePriceOracle::price(T::WethAssetId::get()) else {
+					log::warn!(target: "runtime::dynamic-evm-fee", "Could not get ETH-HDX reference price from oracle");
+					return;
+				};
+				let Some(eth_hdx_reference_price) =
+					FixedU128::checked_from_rational(eth_hdx_reference_price.n, eth_hdx_reference_price.d)
 				else {
-					log::warn!(target: "runtime::dynamic-evm-fee", "Could not get rational of eth-hdx price, current price: {}, reference price: {}", eth_hdx_price, ETH_HDX_REFERENCE_PRICE);
+					log::warn!(target: "runtime::dynamic-evm-fee", "Could not get rational of eth-hdx reference price, n: {}, d: {}", eth_hdx_reference_price.n, eth_hdx_reference_price.d);
+					return;
+				};
+				let Some(price_diff) =
+					FixedU128::checked_from_rational(eth_hdx_price.into_inner(), eth_hdx_reference_price.into_inner())
+				else {
+					log::warn!(target: "runtime::dynamic-evm-fee", "Could not get rational of eth-hdx price, current price: {}, reference price: {}", eth_hdx_price, eth_hdx_reference_price);
 					return;
 				};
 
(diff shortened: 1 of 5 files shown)


pallets/dynamic-evm-fee/src/lib.rs
```diff
@@ -101,6 +99,9 @@ pub mod pallet {
 		/// Native price oracle
 		type NativePriceOracle: NativePriceOracle<Self::AssetId, EmaPrice>;
 
+		/// Referene Native price oracle
+		type ReferenceNativePriceOracle: NativePriceOracle<Self::AssetId, EmaPrice>;
```
enthusiastmartin: not a fan. it is essentially the same. 

let me think if we can find a better approach

pallets/dynamic-evm-fee/src/lib.rs
```diff
@@ -147,10 +148,20 @@ pub mod pallet {
...
+				};
+				let Some(eth_hdx_reference_price) =
+					FixedU128::checked_from_rational(eth_hdx_reference_price.n, eth_hdx_reference_price.d)
+				else {
+					log::warn!(target: "runtime::dynamic-evm-fee", "Could not get rational of eth-hdx reference price, n: {}, d: {}", eth_hdx_reference_price.n, eth_hdx_reference_price.d);
+					return;
+				};
 				let Some(price_diff) =
-					FixedU128::checked_from_rational(eth_hdx_price.into_inner(), ETH_HDX_REFERENCE_PRICE.into_inner())
+					FixedU128::checked_from_rational(eth_hdx_price.into_inner(), eth_hdx_reference_price.into_inner())
```
enthusiastmartin: so here is the issue. 

you get two prices, as ratios. you convert them into fixedu128, which internally does some ratio math magic.

this is done 2x

then you do it again for the prifce_diff

and then again when you do saturating_mul_int. 

so 4x instead of 2 -  i would believe it can be simplified into two multiplications.

it might even be more precise. 

however - i would first try to resolve the tests results to ensure all tests pass before refactoring.

dmoka: not sure how you meant it, can you please propose how you would imagine it with just 2x?

runtime/hydradx/src/assets.rs
```diff
@@ -853,6 +854,15 @@ pub type FeePriceOracle = AssetFeeOraclePriceProvider<
 	ShortOraclePeriod,
 >;
 
+pub type FeePriceOracleForTenMinutes = AssetFeeOraclePriceProvider<
```
enthusiastmartin, line 909: feels out of place here

enthusiastmartin (commented):
just a minor objection

pallets/dynamic-evm-fee/src/lib.rs
enthusiastmartin:
```rust
	let eth_per_hdx = Ratio::from((eth_per_hdx.n, eth_per_hdx.d));
	let eth_per_hdx_reference = Ratio::from((eth_per_hdx_reference.n, eth_per_hdx_reference.d));

	let price_diff = eth_per_hdx.saturating_div(&eth_per_hdx_reference);

	let Some(new_base_fee_per_gas) = multiply_by_rational_with_rounding(
		new_base_fee_per_gas,
		price_diff.n,
		price_diff.d,
		Rounding::Down,
	) else {
		return;
	};
```

pallets/dynamic-evm-fee/src/lib.rs
```diff
@@ -138,19 +159,27 @@ pub mod pallet {
...
 					return;
 				};
 
+				let Some(eth_per_hdx_reference) =
+					FixedU128::checked_from_rational(eth_per_hdx_reference.n, eth_per_hdx_reference.d)
+				else {
+					log::warn!(target: "runtime::dynamic-evm-fee", "Could not get rational of eth-hdx reference price, n: {}, d: {}", eth_per_hdx_reference.n, eth_per_hdx_reference.d);
+					return;
+				};
 				let Some(price_diff) =
```
enthusiastmartin: nitpick:

it is not really a price difference when you divide one price by another price.

it is a price ratio, or relative price.

runtime/hydradx/src/assets.rs
dmoka: I just placed next to the the other oracle `FeePriceOracle`, open to some suggestions.

runtime/hydradx/src/evm/mod.rs
```diff
@@ -70,8 +76,8 @@ pub const GAS_PER_SECOND: u64 = 40_000_000;
 // Approximate ratio of the amount of Weight per Gas.
 const WEIGHT_PER_GAS: u64 = WEIGHT_REF_TIME_PER_SECOND / GAS_PER_SECOND;
 
-// Fixed gas price of 0.016 gwei per gas
-pub const DEFAULT_BASE_FEE_PER_GAS: u128 = 15_000_000;
+// Fixed gas price of 0.006 gwei per gas
```
jak-pan, line 83: what did we derive this from?

runtime/hydradx/src/evm/mod.rs
```diff
@@ -59,6 +63,8 @@ mod runner;
 pub use erc20_currency::Erc20Currency;
 pub use erc20_currency::Function;
 pub use executor::Executor;
+use hydra_dx_math::ema::EmaPrice;
+use hydradx_traits::NativePriceOracle;
 pub use primitives::AccountId as AccountIdType;
 
 // Current approximation of the gas per second consumption considering
```
jak-pan, line 74: This approximation is outdated. We have 2sec execution. Do we want to keep it?

runtime/hydradx/src/evm/mod.rs
dmoka: based on the integration tests like -> compare_fee_in_eth_between_evm_and_native_omnipool_calls
 
 also verified it with live fork using zombienet

runtime/hydradx/src/evm/mod.rs
dmoka: not sure how it was appriximated back then, maybe @cl0w5 ?

dmoka:
Closing this for now since we’re waiting on the new frontier runtime upgrade, which can take some time, but then we can readjust the fee. We can reopen when we pick it up again.

PR #1138, opened 2025-06-25
feat:  add support for routed price to liquidity mining's price adjustment adapter

martinfridrich:
This PR adds routed oracle support to liquidity mining's price adjustment adapter


enthusiastmartin (approved):
i am not fan of the adapter but for now - i guess it does the job.

mrq1911 (approved):

PR #1142, opened 2025-07-01
feat: circuit breaker for deposit

dmoka:
TODO: 
- [x] wait for last rebench commit of ref machine
- [x] grab benchmarks from bench machine
- [x] remove amount from release
- [x] create issue for untrapping assets as in XcmEchangeAsset we trap instead or reserve tokens when depost limit exceeds
- [x] rebench exchange asset too - worst case is buy as we calculate amount in too
- [x] wait for rebech of two LM pallets then replace

- [ ] add action to set xcm-rate-limits for famous, birdged, sharetokens, etc assets. Limit should be also calibrated based on bug bounty values, making it always more profitable for a hacker to report the issue than to exploit it. MAYBE IT IS BETTER TO DETERMINE the daily mint, and multiplying it. Do a histogram, but not only how much we mint and how much we burn, so net mint should be calculated. We query daily net mint, and pick end of the historgram, without outliers
- [ ] we need bot for unlocking assets


pallets/currencies/src/lib.rs
```diff
@@ -960,6 +960,12 @@ where
 	}
 }
 
+impl<T: Config> GetByKey<CurrencyIdOf<T>, BalanceOf<T>> for Pallet<T> {
```
enthusiastmartin: Create a struct AssetTotalIssuance and implement Get by key for that structure, instead of implementing it for Pallet.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -423,6 +467,97 @@ pub mod pallet {
...
+		/// Parameters:
+		/// - `origin`: The dispatch origin for this call. Can be signed or root.
+		/// - `who`: The account that is saving the deposit.
+		/// - `asset_id`: The identifier of the asset.
+		/// - `amount`: The amount of the asset to save as a deposit
+		///
+		/// Emits `DepositSaved` event when successful.
+		#[pallet::call_index(5)]
+		#[pallet::weight(<T as Config>::WeightInfo::save_deposit())]
+		pub fn save_deposit(
```
enthusiastmartin: rename to release_deposit

enthusiastmartin (commented):
few minor comments for now

cl0w5:
Are the most important invariants defined under try-runtime (fuzzing)?

dmoka:
> Why are read/writes sames for all `add_liquidity()`? Doesn't circuit breaker add storage ops?

Nice catch, it doesn't add because the deposit limit is not set for those assets in benchmarks, so the circuit breaker stops early. Let me adjust those benchmark tests

pallets/circuit-breaker/src/tests/lockdown_asset.rs
```diff
@@ -0,0 +1,103 @@
+use crate::tests::mock::{CircuitBreaker, ExtBuilder, RuntimeOrigin, System, Test, Tokens, ALICE};
+use crate::types::LockdownStatus;
+use crate::{AssetLockdownState, Error, Event};
+use frame_support::{assert_noop, assert_ok};
+use orml_traits::MultiCurrency;
+
+pub const ASSET_ID: u32 = 10000;
+
+#[test]
+fn lockdown_asset_should_fork_for_new_asset() {
```
martinfridrich:
```suggestion
fn lockdown_asset_should_work_for_new_asset() {
```
please fix it for rest of tests

integration-tests/src/deposit_limiter.rs
```diff
@@ -0,0 +1,741 @@
...
+use polkadot_xcm::opaque::lts::WeightLimit;
+use polkadot_xcm::opaque::v3::{
+	Junction,
+	Junctions::{X1, X2},
+	MultiLocation, NetworkId,
+};
+use primitives::constants::currency::UNITS;
+
+#[test]
+fn hydra_should_block_asset_from_other_hain_when_over_limit() {
```
martinfridrich:
```suggestion
fn hydra_should_block_asset_from_other_chain_when_over_limit() {
```

runtime/hydradx/src/weights/pallet_omnipool.rs
```diff
@@ -249,21 +253,19 @@ impl<T: frame_system::Config> pallet_omnipool::WeightInfo for HydraWeight<T> {
...
+			.saturating_add(T::DbWeight::get().writes(26_u64))
 	}
 	/// Storage: `Omnipool::Assets` (r:2 w:2)
 	/// Proof: `Omnipool::Assets` (`max_values`: None, `max_size`: Some(85), added: 2560, mode: `MaxEncodedLen`)
 	/// Storage: `AssetRegistry::Assets` (r:3 w:0)
 	/// Proof: `AssetRegistry::Assets` (`max_values`: None, `max_size`: Some(125), added: 2600, mode: `MaxEncodedLen`)
 	/// Storage: `Tokens::Accounts` (r:7 w:7)
 	/// Proof: `Tokens::Accounts` (`max_values`: None, `max_size`: Some(108), added: 2583, mode: `MaxEncodedLen`)
-	/// Storage: `DynamicFees::AssetFeeConfiguration` (r:2 w:0)
-	/// Proof: `DynamicFees::AssetFeeConfiguration` (`max_values`: None, `max_size`: Some(93), added: 2568, mode: `MaxEncodedLen`)
```
martinfridrich: Why are these missing in your weights? also in other places

enthusiastmartin: it has to be rebenchmarked to include this. as it is new features that was just merged.

pallets/circuit-breaker/src/lib.rs
```diff
@@ -423,6 +463,90 @@ pub mod pallet {
...
+		///
+		/// Emits `DepositReleased` event when successful.
+		#[pallet::call_index(5)]
+		#[pallet::weight(<T as Config>::WeightInfo::release_deposit())]
+		pub fn release_deposit(
+			origin: OriginFor<T>,
+			who: T::AccountId,
+			asset_id: T::AssetId,
+		) -> DispatchResultWithPostInfo {
+			ensure_signed_or_root(origin)?;
```
jak-pan: could be also authority?

pallets/circuit-breaker/src/lib.rs
```diff
@@ -320,6 +354,12 @@ pub mod pallet {
 		MaxLiquidityLimitPerBlockReached,
 		/// Asset is not allowed to have a limit
 		NotAllowed,
+		/// Asset in lockdown
```
jak-pan: Let's make the docs better, you should understand what happened by reading it.

dmoka: let's do that, adjusted!

pallets/circuit-breaker/src/lib.rs
dmoka: make sense, i see that as an usecase, it can be called by anyone then

martinfridrich (approved):

mrq1911 (approved):

PR #1152, opened 2025-07-10
feat: track pools' shares issuance in ema-oracle 

martinfridrich:
This PR adds `shares_issuance: Option<Balance>` to `OracleEntry`. This field will be used to track pool's shares issuance.
`shares_issuance` is initialized to `None` during migration that migrates all existing oracles. It's value will be updated lazyly by pools using `OnActivityHandler`.


pallets/ema-oracle/src/migrations/v1.rs
```diff
@@ -0,0 +1,89 @@
...
+			reads += 1;
+
+			migrated_entries.push((
+				key,
+				(
+					OracleEntry::new(
+						val_v0.0.price,
+						val_v0.0.volume,
+						val_v0.0.liquidity,
+						None,
```
jak-pan: shouldn't we insert correct values here?

PR #1154, opened 2025-07-10
feat: stableswap - support to update pool peg data

enthusiastmartin:
This PR adds two new exitrnisic

1. update_asset_peg_source - change peg source for given asset in given pool
2. update_max_peg_update - update max peg update field for given pool.


pallets/stableswap/src/lib.rs
```diff
@@ -1250,6 +1264,105 @@ pub mod pallet {
...
+		///
+		#[pallet::call_index(13)]
+		#[pallet::weight(<T as Config>::WeightInfo::update_asset_peg_source())]
+		#[transactional]
+		pub fn update_asset_peg_source(
+			origin: OriginFor<T>,
+			pool_id: T::AssetId,
+			asset_id: T::AssetId,
+			peg_source: PegSource<T::AssetId>,
+			preserve_price: bool,
```
mrq1911: we will always want to preserve peg

pallets/stableswap/src/lib.rs
```diff
@@ -1250,6 +1264,105 @@ pub mod pallet {
...
+		/// Emits `PoolMaxPegUpdateUpdated` event when successful.
+		///
+		/// # Errors
+		/// - `PoolNotFound`: If the specified pool does not exist.
+		/// - `NoPegSource`: If the pool does not have pegs configured.
+		///
+		#[pallet::call_index(14)]
+		#[pallet::weight(<T as Config>::WeightInfo::update_pool_max_peg_update())]
+		#[transactional]
+		pub fn update_pool_max_peg_update(
```
mrq1911, line 1336: we maybe only need one extrinsic for both peg source and max peg update

pallets/stableswap/src/lib.rs
```diff
@@ -1250,6 +1264,105 @@ pub mod pallet {
...
+		#[pallet::weight(<T as Config>::WeightInfo::update_asset_peg_source())]
+		#[transactional]
+		pub fn update_asset_peg_source(
+			origin: OriginFor<T>,
+			pool_id: T::AssetId,
+			asset_id: T::AssetId,
+			peg_source: PegSource<T::AssetId>,
+			preserve_price: bool,
+		) -> DispatchResult {
+			T::UpdateTradabilityOrigin::ensure_origin(origin)?;
```
mrq1911: should be some more priviledged origin than TC

pallets/stableswap/src/lib.rs
enthusiastmartin: max peg update is not per peg source. it is same for all pegs in the pool.

pallets/stableswap/src/lib.rs
enthusiastmartin: can be same origin as create pool - AuthorityOrigin.

pallets/stableswap/src/lib.rs
enthusiastmartin: will remove

PR #1156, opened 2025-07-14
fix: unlock leftover reserves

dmoka:
### TODOS:
- [x] ❌ This fails, so the current solution wont work - verify on chopsticks if the leftover from DCA 18118 on block 8082390 can be unreserved with this new extrinsic

### Description
Workaround to unlock reserved tokens in DCA.

The long term solution would be to fix AToken sell all correctly.

Hisotry
```
It seems this leftover is coming from his schedule before his last one, so with id "18118", which is completed at block 8082390

Before that, the dca executions are fine, and dont leave funds reserved.

Link: https://hydration.subscan.io/block/8082390?tab=event&event=8082390-40

It is indeed because of the rounding error.

For BUY DCA, we have a special case for the last trade, namely if the next trade amount is smaller than the remaining amount, then we complete the DCA and unreserve all remaining amount.

As for the unreserving part, we do it silently, as we should never really fail when completing a DCA.

It fails at this line: https://github.com/galacticcouncil/hydration-node/blob/b47d1a3d290c2139b16fa9248a30d3c67bc2e8ce/pallets/currencies/src/lib.rs#L587

Because it tries to transfer 25701, but the balance of the user is only 25700. Hence the leftover reserved balance. 

```


Roznovjak (commented):
I would remove all weight changes that are not related to the PR changes.

dmoka:
the proposed solution wont work due to the same rounding issue, so we gotta solve th AToken transfer properly as initially planned

pallets/dca/src/lib.rs
```diff
@@ -637,6 +646,47 @@ pub mod pallet {
...
+		/// Emits `ReserveUnlocked` event when successful.
+		///
+		#[pallet::call_index(2)]
+		#[pallet::weight(<T as Config>::WeightInfo::unlock_reserves())]
+		#[transactional]
+		pub fn unlock_reserves(origin: OriginFor<T>, who: T::AccountId, asset_id: T::AssetId) -> DispatchResult {
+			let _ = ensure_signed_or_root(origin)?;
+
+			ensure!(
+				!ScheduleOwnership::<T>::iter_prefix(who.clone()).next().is_some(),
```
Roznovjak:
```suggestion
				ScheduleOwnership::<T>::iter_prefix(who.clone()).next().is_none(),
```

dmoka:
Closed as the proposed solution leads to the same rounding issue coming from AAVE contract, we need a collective and secure way for atoken transfers so it is fixed.

dmoka:
Reopened since the proposed solution might work on diffferent block

integration-tests/src/dca.rs
```diff
@@ -4139,6 +4139,68 @@ fn terminate_should_work_for_freshly_created_dca() {
 	});
 }
 
+#[test]
+fn unclock_should_not_work_when_user_has_active_schedule() {
```
mrq1911: unclock typo

PR #1159, opened 2025-07-15
feat: support to set custom omnipool dynamic/fixed fee for single asset

enthusiastmartin:
This PR adds a support to set a custom fee (asset and protocol) per single asset.

Currently, it is global fee parameters that are applied for each asset. It is global config set in hydration runtime and any change requires runtime upgrade.

With this support - it will be possible to set a custom fee parametes for selected asset.

If nothing is set for an asset - it falls back to original behaviour and default params are used. 

Moreover, it is also now possible to set a FIXED fee instead of dynamic.

### DIFF
--- a/integration-tests/src/evm_permit.rs
+++ b/integration-tests/src/evm_permit.rs
@@ -1755,7 +1755,7 @@ fn dispatch_permit_should_charge_tx_fee_when_call_fails() {
 		let hdx_balance = user_acc.balance(HDX);
 		let tx_fee = initial_user_hdx_balance - hdx_balance;
 
-		assert_eq!(tx_fee, 4491170276643);
+		assert_eq!(tx_fee, 4210472144093);
 	})
 }
 
@@ -1962,7 +1962,7 @@ fn dispatch_permit_should_not_pause_tx_when_call_execution_fails() {
 		let hdx_balance = user_acc.balance(HDX);
 		let tx_fee = initial_user_hdx_balance - hdx_balance;
 
-		assert_eq!(tx_fee, 4491170276643);
+		assert_eq!(tx_fee, 4210472144093);
 
 		let call = RuntimeCall::MultiTransactionPayment(pallet_transaction_multi_payment::Call::dispatch_permit {
 			from: user_evm_address,


pallets/dynamic-fees/src/lib.rs
```diff
@@ -116,16 +130,72 @@ pub mod pallet {
...
+		/// Information on runtime weights.
+		type WeightInfo: WeightInfo;
 	}
 
 	#[pallet::event]
-	pub enum Event<T: Config> {}
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Asset fee configuration has been set
+		AssetFeeConfigSet { asset_id: T::AssetId },
```
martinfridrich: Shouldn't fee's config be in event?

pallets/dynamic-fees/src/tests/asset_config.rs
```diff
@@ -0,0 +1,320 @@
...
+		assert_ok!(DynamicFees::set_asset_fee(RuntimeOrigin::root(), asset_id, config));
+
+		// Verify storage
+		let stored_config = AssetFeeConfiguration::<Test>::get(asset_id);
+		assert_eq!(stored_config, Some(config));
+
+		// Verify event was emitted
+		// System::assert_last_event(
+		// 	crate::Event::AssetFeeConfigSet { asset_id }.into()
+		// );
```
martinfridrich: Why is this commented(also in other tests)? Remove or uncomment it

runtime/hydradx/src/weights/pallet_dynamic_fees.rs
```diff
@@ -0,0 +1,83 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+//! Autogenerated weights for `pallet_dynamic_fees`
+//!
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 43.0.0
+//! DATE: 2025-07-15, STEPS: `16`, REPEAT: `1`, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
martinfridrich: Rebenchmark with prod. params.

enthusiastmartin:
> I don't like comments that just says what code does(these are useless IMO) and that you are using name `timestamp` but it's block number. These doesn't affect functionality so I didn't comment on these. I requested changes because of benchmarks

i agree with useless code comments however i dont see any in this PR. So can you point ?

Timestamp was there before, so not goiing to change that now. 

if you mean those in tests and benchmarks - yes. will remove.

martinfridrich (approved):

PR #1165, opened 2025-08-05
feat: improve hsm arbitrage

enthusiastmartin:
This PR improves arbitrage in HSM - adding handling of scenario when there is less hollar in the stablepool.

it also improve offchain worker which sends execute_arb only if there is an opportunity.


pallets/hsm/src/lib.rs
```diff
@@ -1248,49 +1285,192 @@ where
...
+		collateral_asset_id: T::AssetId,
+		imbalance: Balance,
+		peg: PegType,
+		info: &CollateralInfo<T::AssetId>,
+		state: &PoolSnapshot<T::AssetId>,
+	) -> Option<Balance> {
+		let mut sell_amount_max = imbalance;
+		let mut sell_amount_min = 0u128;
+		let mut sell_amount = sell_amount_max / 2;
+		for _ in 0..50 {
```
dmoka, line 1378: any specific reason for looping for 50? Maybe we could introduce a config for this?

mrq1911 (approved):

PR #1168, opened 2025-08-14
feat: stableswap - allow lower values for max peg update

enthusiastmartin:


pallets/stableswap/src/types.rs
```diff
@@ -4,7 +4,7 @@
 use serde::{Deserialize, Serialize};
 
 use crate::{Config, Pallet, PoolPegs, MAX_ASSETS_IN_POOL};
-use sp_runtime::Permill;
+use sp_runtime::{Perbill, Permill};
```
Roznovjak, line 7: Remove `Permill` if not needed anymore

enthusiastmartin: it is used

PR #1173, opened 2025-08-20
feat: lower the evm base fee in testnets

Roznovjak:
This PR reduces the EVM `BaseFeePerGas` computed by `pallet-dynamic-evm-fee` if the testnet flag in `pallet-parameters` is `true` . Final fee is reduced 10 times.

### DIFF
--- a/pallets/dynamic-evm-fee/src/lib.rs
+++ b/pallets/dynamic-evm-fee/src/lib.rs
@@ -105,7 +105,7 @@ pub mod pallet {
 		#[pallet::constant]
 		type WethAssetId: Get<Self::AssetId>;
 
-		/// Base fee multiplier
+		/// Base fee multiplier to scale fee for test environments
 		type BaseFeePerGasMultiplier: Get<FixedU128>;
 
 		/// Weight information for extrinsics in this pallet.


pallets/dynamic-evm-fee/src/lib.rs
```diff
@@ -105,6 +105,9 @@ pub mod pallet {
 		#[pallet::constant]
 		type WethAssetId: Get<Self::AssetId>;
 
+		/// Testnet flag
+		type TestnetFlag: Get<bool>;
```
enthusiastmartin: i dont like using "testnet"  word here.

i would suggest to do:

`type FeeMultiplier: Get<FixedU128>;   // or other type if you prefer`

then in the code below - you would just do base_fee * multiplier

and actually, in the runtime config - you would do 

```rust
if testnet {
   return 0.1
}else{
    return 1;
}
```

pallets/dynamic-evm-fee/src/lib.rs
```diff
@@ -105,6 +105,9 @@ pub mod pallet {
 		#[pallet::constant]
 		type WethAssetId: Get<Self::AssetId>;
 
+		/// Base fee multiplier
```
dmoka: I would include the WHY here, so something like
```suggestion
		/// Base fee multiplier to scale fee for test environments
```

dmoka (approved):
Added

PR #1175, opened 2025-08-22
feat: add testnet config

dmoka:
Set tx version and evmChainId to different values for test environments


mrq1911:
no need for feature flag, we'll just adjust tx cversion manually if needed...

PR #1176, opened 2025-08-25
feat: hsm - limit flash loan amount by capcacity

enthusiastmartin:


pallets/hsm/src/lib.rs
```diff
@@ -1699,6 +1704,55 @@ where
 			primitive_types::U128::from(buy_price.1).full_mul(primitive_types::U128::from(max_price.0));
 		buy_price_check <= max_price_check
 	}
+
+	fn get_max_flash_loan_amount() -> Balance {
```
dmoka, line 1708: would not be better to return Option<Balance> and return None instead of zeros?

And also for get_hsm_bucket_free_capacity

The find_ideal_trade_size also returns Option, so we could just propagate the Nones...

enthusiastmartin: it does not make any difference. in any case, we return no opportunity.  but it might look better, will consider it.

mrq1911 (approved):

PR #1178, opened 2025-08-25
fix: possible overflow in price provider for given route

enthusiastmartin:
The pr fixes potential overflow in the price provider implementation that calculates a price for given route.

if the route contains more than 4-5 trades - it may overflow.

- [x] Add unit tests

### DIFF
--- a/runtime/adapters/src/lib.rs
+++ b/runtime/adapters/src/lib.rs
@@ -629,18 +629,12 @@ where
 		if prices.is_empty() {
 			return None;
 		}
-
-		let nominator = prices
-			.iter()
-			.try_fold(U512::from(1u128), |acc, price| acc.checked_mul(U512::from(price.n)))?;
-
-		let denominator = prices
-			.iter()
-			.try_fold(U512::from(1u128), |acc, price| acc.checked_mul(U512::from(price.d)))?;
-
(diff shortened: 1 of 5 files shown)


runtime/adapters/src/lib.rs
```diff
@@ -629,18 +629,12 @@ where
...
-			.try_fold(U512::from(1u128), |acc, price| acc.checked_mul(U512::from(price.n)))?;
-
-		let denominator = prices
-			.iter()
-			.try_fold(U512::from(1u128), |acc, price| acc.checked_mul(U512::from(price.d)))?;
-
-		let rat_as_u128 = round_u512_to_rational((nominator, denominator), Rounding::Nearest);
-
-		Some(EmaPrice::new(rat_as_u128.0, rat_as_u128.1))
+		let price = prices.iter().try_fold((1u128, 1u128), |acc, price| {
```
mrq1911: only integrate it every 4 steps as uint128 * 4 will always fit in uint512

mrq1911 (changes requested):

runtime/adapters/src/lib.rs
enthusiastmartin: done

mrq1911 (approved):

Issue #998, opened 2025-01-23
Omnipool liquidator off chain worker


mrq1911:
makes sure that on chain omnipool liquidation is triggered by block producer (collator) directly after oracle update transaction

mrq1911:
bonus: only liquidate partionally (less than 50%) to get position to some safe place

jak-pan:
This could be done by specifying Priority out of bounds for normal TX so it's always first.

mrq1911:
#937

mrq1911:
#1181

enthusiastmartin:
all merged

PR #1181, opened 2025-09-08
feat: improved offchain liquidation

Roznovjak:
- [ ] set hard limit for number of liquidation per block, smth like 50% of blocktime

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5388,7 +5388,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx"
-version = "14.5.0"
+version = "14.6.0"
 dependencies = [
  "async-trait",
  "clap",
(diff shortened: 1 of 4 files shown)


liquidation-worker-support/src/lib.rs
```diff
@@ -135,12 +135,42 @@ pub struct BorrowerDataDetails<AccountId> {
...
 	pub last_global_update: u32,
 	pub last_update: u32,
 	pub borrowers: Vec<(H160, BorrowerDataDetails<AccountId>)>,
 }
 
+#[derive(Clone, Debug)]
+pub struct Borrower {
+	pub user_address: EvmAddress,
+	pub health_factor: U256,
+	pub assets: Vec<EvmAddress>,
```
enthusiastmartin: looking at the code - i think some HashSet would be better here ?

it seems that you require the elements to be unique.

when adding a new entry to the list - you do "contains" - if hashset used, there is only O(1) instead of current O(n)

Roznovjak: Thanks for the comment.
Adding new borrower to the list doesn't happen very often. The most common operation we do is iteration, and iter over `Vec` is faster. Also `HashSet` is not sorted.

mrq1911:
fixes #998

PR #1182, opened 2025-09-14
feat(evm):  add build evm tx pallet

Pessina:
## Description

- Include new pallet with helper function to build EIP-1559 EVM transaction

## Motivation and Context

This PR is part of integration with Sig Network. 

## Checklist:


Pessina:
As discussed with @enthusiastmartin on the call. 

Currently this pallet it's very simple, so we could think it should be a util function. There are 2 reasons to keep it as a pallet: 

- Make it a precompiled in the future, so EVM contract also have access to transaction building
   - EVM transaction building in Solidity it's simple, but other chains, like: BTC, Cosmos, Solana, etc. are more complex, so it's easier to implement them in Rust and upgrade the pallets
- Re-use across other pallets

pallets/build-evm-tx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::pallet_prelude::*;
+use frame_system::{ensure_signed, pallet_prelude::OriginFor};
+use sp_std::vec::Vec;
+
+use alloy_consensus::{TxEip1559, TxType};
```
mrq1911: we already have dependencies to encode ethereum transactions in runtime, we use `ethereum 0.15.0` crate. please try to use that if possible of introducing new dependencies.

Pessina: Fixed :)

enthusiastmartin:
closing in favor of #1189 . it is already part of that PR too.

PR #1187, opened 2025-09-17
feat: dusting support for atoken

dmoka:
todos:
- [ ] rebenchmark
- [ ] regenerate benches
- [ ] try out to dust aUSDT on prod zombienet for account 149AQApEw4Xe1DxWesqywiuGQHazbETMgWgT89WjEYFAc29y

### DIFF
--- a/pallets/duster/src/lib.rs
+++ b/pallets/duster/src/lib.rs
@@ -66,11 +66,6 @@ pub mod pallet {
 	pub type AccountBlacklist<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, (), OptionQuery>;
 
 	#[pallet::storage]
-	#[pallet::getter(fn reward_account)]
-	/// Account to take reward from.
-	pub type RewardAccount<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
-
-	#[pallet::storage]
 	#[pallet::getter(fn dust_dest_account)]
 	/// Account to send dust to.
 	pub type DustAccount<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
@@ -116,10 +111,6 @@ pub mod pallet {
 		/// The minimum amount required to keep an account.
 		type MinCurrencyDeposits: GetByKey<Self::CurrencyId, Self::Balance>;
 
-		/// Reward amount
-		#[pallet::constant]
-		type Reward: Get<Self::Balance>;
-
 		/// Native Asset Id
 		#[pallet::constant]
 		type NativeCurrencyId: Get<Self::CurrencyId>;
@@ -130,7 +121,7 @@ pub mod pallet {
 		/// Duster for accounts with AToken dusts
 		type ATokenDuster: hydradx_traits::evm::ATokenDuster<Self::AccountId, Self::CurrencyId>;
 
-		/// Default account for `reward_account` and `dust_account` in genesis config.
+		/// Default account for `dust_account` in genesis config.
 		#[pallet::constant]
 		type TreasuryAccountId: Get<Self::AccountId>;
 
@@ -142,7 +133,6 @@ pub mod pallet {
 	#[derive(frame_support::DefaultNoBound)]
 	pub struct GenesisConfig<T: Config> {
 		pub account_blacklist: Vec<T::AccountId>,
-		pub reward_account: Option<T::AccountId>,
 		pub dust_account: Option<T::AccountId>,
 	}
 
@@ -153,7 +143,6 @@ pub mod pallet {
 				AccountBlacklist::<T>::insert(account_id, ());
 			});
 
-			RewardAccount::<T>::put(self.reward_account.clone().unwrap_or_else(T::TreasuryAccountId::get));
 			DustAccount::<T>::put(self.dust_account.clone().unwrap_or_else(T::TreasuryAccountId::get));
 		}
 	}
@@ -198,7 +187,11 @@ pub mod pallet {
 		/// IF account balance is < min. existential deposit of given currency, and account is allowed to
 		/// be dusted, the remaining balance is transferred to selected account (usually treasury).
 		///
-		/// Caller is rewarded with chosen reward in native currency.
+		/// In case of AToken, the dusting is performed via ATokenDuster traait, which does a wihtdraw all on behalf of the dust receiver
+		///
+		/// The transaction fee is returned back in case of sccessful dusting.
+		///
+		/// Emits `Dusted` event when successful.
 		#[pallet::call_index(0)]
 		#[pallet::weight(<T as Config>::WeightInfo::dust_account())] //TODO: rebenchmark it
 		pub fn dust_account(
@@ -227,10 +220,6 @@ pub mod pallet {
 			} else {
 				Self::transfer_dust(&account, &dust_dest_account, currency_id, dust)?;
 			}
-
-			//TODO: remove reward
-			// Ignore the result, it fails - no problem.
-			let _ = Self::reward_duster(&who, currency_id, dust);
 
 			Self::deposit_event(Event::Dusted {
 				who: account,
@@ -289,17 +278,6 @@ impl<T: Config> Pallet<T> {
 		let total = T::MultiCurrency::total_balance(currency_id, account);
 
 		(total < ed, total)
-	}
-
-	/// Send reward to account which did the dusting.
-	fn reward_duster(_duster: &T::AccountId, _currency_id: T::CurrencyId, _dust: T::Balance) -> DispatchResult {
-		// Error should never occur here
-		let reserve_account = Self::reward_account().ok_or(Error::<T>::ReserveAccountNotSet)?;
-		let reward = T::Reward::get();
-
-		T::MultiCurrency::transfer(T::NativeCurrencyId::get(), &reserve_account, _duster, reward)?;
-
-		Ok(())
 	}
 
 	/// Transfer dust amount to selected DustAccount ( usually treasury)
--- a/pallets/duster/src/lib.rs
+++ b/pallets/duster/src/lib.rs
@@ -29,8 +29,8 @@ pub mod migration;
 pub mod weights;
 
 pub use crate::weights::WeightInfo;
-
 use frame_support::{dispatch::DispatchResult, ensure, traits::Contains, traits::Get};
+use hydradx_traits::evm::ATokenDuster;
 
 use orml_traits::{
 	arithmetic::{Signed, SimpleArithmetic},
@@ -127,6 +127,9 @@ pub mod pallet {
 		/// The origin which can manage whiltelist.
 		type BlacklistUpdateOrigin: EnsureOrigin<Self::RuntimeOrigin>;
 
+		/// Duster for accounts with AToken dusts
+		type ATokenDuster: hydradx_traits::evm::ATokenDuster<Self::AccountId, Self::CurrencyId>;
+
 		/// Default account for `reward_account` and `dust_account` in genesis config.
 		#[pallet::constant]
 		type TreasuryAccountId: Get<Self::AccountId>;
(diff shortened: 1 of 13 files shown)


runtime/hydradx/src/migrations/mod.rs
```diff
@@ -63,4 +63,6 @@ pub type Migrations = (
 	frame_support::migrations::RemovePallet<CouncilPalletName, <Runtime as frame_system::Config>::DbWeight>,
 	frame_support::migrations::RemovePallet<PhragmenElectionPalletName, <Runtime as frame_system::Config>::DbWeight>,
 	frame_support::migrations::RemovePallet<TipsPalletName, <Runtime as frame_system::Config>::DbWeight>,
+	// Duster pallet migration to rename AccountBlacklist to AccountWhitelist
```
Roznovjak: please add a comment with the runtime version. This way it's clear when the migration can be removed.

pallets/duster/src/migration.rs
```diff
@@ -53,7 +53,7 @@ pub mod v1 {
 		);
 		reads += 1;
 		account_blacklist.iter().for_each(|account_id| {
-			crate::AccountBlacklist::<T>::insert(account_id, ());
+			crate::AccountWhitelist::<T>::insert(account_id, ());
```
Roznovjak, line 56: rename all occurrencies of `account_blacklist`

pallets/duster/src/migration.rs
```diff
@@ -79,3 +79,80 @@ pub mod v1 {
...
+			reads += blacklisted_accounts.len() as u64;
+
+			log::info!(
+				target: "runtime::duster",
+				"Migrating {} accounts from AccountBlacklist to AccountWhitelist",
+				blacklisted_accounts.len()
+			);
+
+			// Insert all entries into the new AccountWhitelist storage
+			for (account_id, _) in &blacklisted_accounts {
```
Roznovjak: We can use `for (account_id, _) in AccountBlacklist::<T, P>::iter() `.
It's not necessary to `collect()` first and then iterate again. Also `reads == writes`.

pallets/duster/src/migration.rs
```diff
@@ -79,3 +79,80 @@ pub mod v1 {
...
+				crate::AccountWhitelist::<T>::insert(account_id, ());
+				writes += 1;
+			}
+
+			// Remove all entries from the old storage
+			let removed_keys = AccountBlacklist::<T, P>::drain().count();
```
Roznovjak, line 138: IIRC this leaves the empty storage there. We want to also "kill" it.

dmoka: AFAIK the storage map itself cannot be killed, only the entries, and the drain does that so I think there is no need for additioinal kill. Also, SotrageMap has no kill method, which might explain it

pallets/duster/src/lib.rs
```diff
@@ -82,52 +76,23 @@ pub mod pallet {
...
-		type Reward: Get<Self::Balance>;
-
-		/// Native Asset Id
-		#[pallet::constant]
-		type NativeCurrencyId: Get<Self::CurrencyId>;
+		/// Existential deposit required to keep an account.
+		type ExistentialDeposit: GetByKey<Self::AssetId, Balance>;
 
 		/// The origin which can manage whiltelist.
 		type BlacklistUpdateOrigin: EnsureOrigin<Self::RuntimeOrigin>;
```
enthusiastmartin: still blacklist used.

pallets/duster/src/lib.rs
```diff
@@ -262,33 +238,22 @@ pub mod pallet {
...
 	/// Check is account's balance is below minimum deposit.
-	fn is_dustable(account: &T::AccountId, currency_id: T::CurrencyId) -> (bool, T::Balance) {
-		let ed = T::MinCurrencyDeposits::get(&currency_id);
+	fn is_dustable(account: &T::AccountId, currency_id: T::AssetId) -> (bool, Balance) {
+		let ed = T::ExistentialDeposit::get(&currency_id);
 
 		let total = T::MultiCurrency::total_balance(currency_id, account);
 
 		(total < ed, total)
 	}
```
enthusiastmartin: there is no point having a special function for this. it is used only once.

enthusiastmartin (changes requested):

pallets/duster/src/lib.rs
```diff
@@ -297,34 +254,33 @@ use orml_traits::currency::OnDust;
...
 impl<T: Config> Contains<T::AccountId> for DusterWhitelist<T> {
 	fn contains(t: &T::AccountId) -> bool {
-		AccountBlacklist::<T>::contains_key(t)
+		AccountWhitelist::<T>::contains_key(t)
 	}
 }
 
 use frame_support::sp_runtime::DispatchError;
+use frame_support::traits::tokens::Preservation;
 use hydradx_traits::pools::DustRemovalAccountWhitelist;
```
enthusiastmartin: move this to the top of the file

pallets/duster/src/mock.rs
```diff
@@ -125,18 +127,37 @@ parameter_type_with_key! {
...
+	type AssetId = AssetId;
+	type MultiCurrency = FungibleCurrencies<Test>;
+	type ExistentialDeposit = MinDeposits;
+	type WhitelistUpdateOrigin = EnsureRoot<AccountId>;
+	type Erc20Support = ATokenDusterMock;
 	type TreasuryAccountId = TreasuryAccount;
 	type WeightInfo = ();
 }
 
+pub struct ATokenDusterMock;
```
enthusiastmartin, line 149: why this is not tested as part of the pallet ? 

i believe you can test it easilly, nicely - you dont need full evm support. just make sure it is all called correcrtly.

runtime/hydradx/src/evm/aave_trade_executor.rs
```diff
@@ -174,8 +176,44 @@ where
 		}
 	}
 
-	pub fn is_atoken(_address: EvmAddress) -> bool {
-		false
+	pub fn is_atoken(address: EvmAddress) -> bool {
```
enthusiastmartin, line 138: why we bring this back again here. ? 

it still have the same issue as before, no ?

pallets/duster/src/lib.rs
```diff
@@ -297,34 +254,33 @@ use orml_traits::currency::OnDust;
...
 	fn add_account(account: &T::AccountId) -> Result<(), Self::Error> {
-		AccountBlacklist::<T>::insert(account, ());
+		AccountWhitelist::<T>::insert(account, ());
 		Ok(())
 	}
 
 	fn remove_account(account: &T::AccountId) -> Result<(), Self::Error> {
-		AccountBlacklist::<T>::mutate(account, |maybe_account| -> Result<(), DispatchError> {
-			ensure!(!maybe_account.is_none(), Error::<T>::AccountNotBlacklisted);
+		AccountWhitelist::<T>::mutate(account, |maybe_account| -> Result<(), DispatchError> {
```
enthusiastmartin, line 289: we should be just use remove method on the storage. 

simpler.

integration-tests/proptest-regressions/dust.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc a42eda86ce372dd6468f5624c7cfb708c3a366d82b14fbbb0d724bc18e93c7af # shrinks to 1
```
enthusiastmartin: what is this about ?

pallets/duster/src/lib.rs
dmoka: i think by using remove, we would change the existing behaviour, because this mutate checks is the account is already whitelisted, if not it returns error AccountNotWhitelisted, otherwise removes it from whitelist.

So i think it is better to keep it, for safety, so it is not called by developer when the account is not whitelisted yet.

integration-tests/proptest-regressions/dust.txt
dmoka: good catch, the test was wrong, fixed, and removed the file!

runtime/hydradx/src/evm/aave_trade_executor.rs
dmoka: Here is what I did to make sure, pls double check too:

- added tests for about the fact that trasnfering atoken leaves dust: https://github.com/galacticcouncil/hydration-node/blob/28562bba8cb267de962ba7d87a3896a44eaf2fd2/integration-tests/src/aave_router.rs#L804

- adding tests for the stableswap add liquidity shares exploit: https://github.com/galacticcouncil/hydration-node/blob/28562bba8cb267de962ba7d87a3896a44eaf2fd2/integration-tests/src/aave_router.rs#L950

- reverted the erc20_currency transfer, so we always do just a transfer, instead of checking for atoken with `is_atoken` etc: https://github.com/galacticcouncil/hydration-node/pull/1187/files#diff-e38ad9b0843012d7155f707981ac71227426a598aa697e5bc69e9f6b49830eee

- completely removed the aaveTradeExecutor transfer function that had the bug with saturating math: https://github.com/galacticcouncil/hydration-node/pull/1187/files#diff-93adca368da3a534f6d904da263d9c8d07c3969eec10697eb60b2abb63172af1

pallets/duster/src/lib.rs
```diff
@@ -195,45 +157,62 @@ pub mod pallet {
...
+			ensure!(!dust.is_zero(), Error::<T>::ZeroBalance);
 
 			// Error should never occur here
-			let dust_dest_account = Self::dust_dest_account().ok_or(Error::<T>::DustAccountNotSet)?;
-
-			Self::transfer_dust(&account, &dust_dest_account, currency_id, dust)?;
+			let dust_dest_account = T::TreasuryAccountId::get();
+
+			if T::Erc20Support::is_atoken(currency_id) {
+				//Temporarily adding the account to blacklist to prevent ED error when AToken is withdrawn from contract
```
enthusiastmartin: blacklist - > whitelist

pallets/duster/src/lib.rs
```diff
@@ -195,45 +157,62 @@ pub mod pallet {
...
+			currency_id: T::AssetId,
+		) -> DispatchResultWithPostInfo {
+			ensure_signed(origin)?;
 
-			let (dustable, dust) = Self::is_dustable(&account, currency_id);
+			ensure!(Self::whitelisted(&account).is_none(), Error::<T>::AccountWhitelisted);
 
-			ensure!(dust != T::Balance::from(0u32), Error::<T>::ZeroBalance);
+			let ed = T::ExistentialDeposit::get(&currency_id);
+			let dust = T::MultiCurrency::total_balance(currency_id, &account.clone().into());
```
enthusiastmartin: no need to clonse
```suggestion
			let dust = T::MultiCurrency::total_balance(currency_id, &account);
```

pallets/duster/src/lib.rs
```diff
@@ -195,45 +157,62 @@ pub mod pallet {
...
+			ensure!(dust < ed, Error::<T>::BalanceSufficient);
 
-			ensure!(dustable, Error::<T>::BalanceSufficient);
+			ensure!(!dust.is_zero(), Error::<T>::ZeroBalance);
 
 			// Error should never occur here
```
enthusiastmartin: useless comment

pallets/duster/src/lib.rs
```diff
@@ -193,47 +155,67 @@ pub mod pallet {
...
+		pub fn dust_account(
+			origin: OriginFor<T>,
+			account: T::AccountId,
+			currency_id: T::AssetId,
+		) -> DispatchResultWithPostInfo {
+			ensure_signed(origin)?;
 
-			let (dustable, dust) = Self::is_dustable(&account, currency_id);
+			ensure!(Self::whitelisted(&account).is_none(), Error::<T>::AccountWhitelisted);
+			ensure!(account != T::TreasuryAccountId::get(), Error::<T>::AccountWhitelisted); //In case if treasury account get dewhitelisted for some reason
```
enthusiastmartin: you can put this into one ensure

pallets/duster/src/lib.rs
```diff
@@ -193,47 +155,67 @@ pub mod pallet {
...
+			Ok(Pays::No.into())
 		}
 
 		/// Add account to list of non-dustable account. Account whihc are excluded from udsting.
-		/// If such account should be dusted - `AccountBlacklisted` error is returned.
+		/// If such account should be dusted - `AccountWhitelisted` error is returned.
 		/// Only root can perform this action.
 		#[pallet::call_index(1)]
 		#[pallet::weight(<T as Config>::WeightInfo::add_nondustable_account())]
 		pub fn add_nondustable_account(origin: OriginFor<T>, account: T::AccountId) -> DispatchResult {
```
enthusiastmartin: can we also rrename this to whitelst_account

and remove -> remove_from_whitelist

nondustable is weird word

pallets/duster/src/lib.rs
```diff
@@ -193,61 +155,88 @@ pub mod pallet {
...
+				Self::add_account(&account)?;
+				T::Erc20Support::on_dust(&account, &dust_dest_account, currency_id)?;
+				Self::remove_account(&account)?;
+			} else {
+				Self::transfer_dust(&account, &dust_dest_account, currency_id, dust)?;
+			}
+
+			//Sanity check that account is fully dusted
+			let leftover =
+				T::MultiCurrency::reducible_balance(currency_id, &account, Preservation::Expendable, Fortitude::Polite);
```
enthusiastmartin: shouldnt this be total_balance ?

dmoka: good catch, fixed!

PR #1189, opened 2025-09-22
Signet pallet

esaminu:
feat: Add Signet pallet for MPC signature requests

## Description
Adds the Signet pallet to enable signature requests from MPC networks. Includes deposit mechanism, admin controls, and support for both simple signatures and transaction signing.

## Motivation and Context
Required for cross-chain interoperability through decentralized signing.

## How Has This Been Tested?
Test coverage for:
- Initialization and configuration
- Signature requests (sign and sign_respond)
- Response handling (single/batch)
- Admin functions (deposit updates, withdrawals)

## Checklist:


pallets/signet/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+		/// Request a signature for a payload
+		#[pallet::call_index(3)]
+		#[pallet::weight(<T as Config>::WeightInfo::sign())]
+		pub fn sign(
+			origin: OriginFor<T>,
+			payload: [u8; 32],
+			key_version: u32,
+			path: Vec<u8>,
+			algo: Vec<u8>,
+			dest: Vec<u8>,
```
enthusiastmartin: any vec should be bounded by some reasonable limit.

pallets/signet/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+			// Ensure initialized
+			ensure!(Admin::<T>::get().is_some(), Error::<T>::NotInitialized);
+
+			// Get deposit amount
+			let deposit = SignatureDeposit::<T>::get();
+
+			// Transfer deposit from requester to pallet account
+			let pallet_account = Self::account_id();
+			T::Currency::transfer(&requester, &pallet_account, deposit, ExistenceRequirement::KeepAlive)?;
+
```
enthusiastmartin, line 316: as discussed - we probably want to support deposit in any currency, preferable a currency that user has setup as fee payment asset. 

let me check the best way to add this.

pallets/build-evm-tx/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+		/// - `data`: Transaction data/calldata
+		/// - `nonce`: Transaction nonce
+		/// - `gas_limit`: Maximum gas units for transaction
+		/// - `max_fee_per_gas`: Maximum total fee per gas (base + priority)
+		/// - `max_priority_fee_per_gas`: Maximum priority fee (tip) per gas
+		/// - `chain_id`: Target EVM chain ID
+		///
+		/// # Returns
+		/// RLP-encoded transaction data with EIP-2718 type prefix (0x02 for EIP-1559)
+		pub fn build_evm_tx(
```
enthusiastmartin, line 74: it feels like this does not need to be a separate pallet. 

i believe it can be part of the signet pallet. so we have all in one place.

pallets/signet/src/lib.rs
```diff
@@ -0,0 +1,464 @@
...
+		/// Initialize the pallet with admin, deposit, and chain ID
+		#[pallet::call_index(0)]
+		#[pallet::weight(<T as Config>::WeightInfo::initialize())]
+		pub fn initialize(
+			origin: OriginFor<T>,
+			admin: T::AccountId,
+			signature_deposit: BalanceOf<T>,
+			chain_id: BoundedVec<u8, T::MaxChainIdLength>,
+		) -> DispatchResult {
+			let _initializer = ensure_signed(origin)?;
```
enthusiastmartin, line 236: i would suggest to allow to initialize the pallet only by chosen origin such as technical commitee. Leaving it this up to any account does not feel right.

pallets/signet/src/lib.rs
```diff
@@ -0,0 +1,464 @@
...
+				chain_id: chain_id.to_vec(),
+			});
+
+			Ok(())
+		}
+
+		/// Update the signature deposit amount (admin only)
+		#[pallet::call_index(1)]
+		#[pallet::weight(<T as Config>::WeightInfo::update_deposit())]
+		pub fn update_deposit(origin: OriginFor<T>, new_deposit: BalanceOf<T>) -> DispatchResult {
```
enthusiastmartin, line 258: as dicusssed in other channels - we need probably to set some max limit that this deposit can be set to.

enthusiastmartin (changes requested):

enthusiastmartin:
closing in favor of #1235 . it is part of that PR

PR #1191, opened 2025-10-01
fix: fee payment asset selection for EVM accounts

F3Joule:
## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5537,7 +5537,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "351.0.0"
+version = "352.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
@@ -11066,7 +11066,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-transaction-multi-payment"
-version = "10.1.7"
+version = "10.1.8"
 dependencies = [
  "frame-support",
  "frame-system",
@@ -14634,7 +14634,7 @@ checksum = "48fd7bd8a6377e15ad9d42a8ec25371b94ddc67abe7c8b9127bec79bebaaae18"
 
 [[package]]
 name = "runtime-integration-tests"
-version = "1.52.5"
+version = "1.53.0"
 dependencies = [
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-parachain-system",
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "runtime-integration-tests"
-version = "1.52.5"
+version = "1.53.0"
 description = "Integration tests"
 authors = ["GalacticCouncil"]
 edition = "2021"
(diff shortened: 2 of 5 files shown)


pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -556,11 +556,24 @@ pub mod pallet {
...
+	/// - Otherwise, default by account type:
+	///     * EVM account     → `T::EvmAssetId::get()`
+	///     * non-EVM account → `T::NativeAssetId::get()`
 	pub fn account_currency(who: &T::AccountId) -> AssetIdOf<T>
 	where
 		BalanceOf<T>: FixedPointOperand,
 	{
-		Pallet::<T>::get_currency(who).unwrap_or_else(T::NativeAssetId::get)
+		Pallet::<T>::get_currency(who.clone()).unwrap_or_else(|| {
+			if T::InspectEvmAccounts::is_evm_account(who.clone()) {
```
enthusiastmartin, line 571: clone here should not be needed?!

mrq1911:
fixes #1200

runtime/hydradx/src/evm/runner.rs
enthusiastmartin, line 1: not sure what thense changes are good for.

i bet they are not needed and should be reverted

integration-tests/src/evm_permit.rs
```diff
@@ -265,6 +273,7 @@ fn fee_should_be_paid_in_hdx_when_permit_is_dispatched_and_address_is_not_bounde
 		);
 
 		// Prepare user evm account - bind and fund
+		set_evm_account_currency(HDX);
```
enthusiastmartin: not sure if i understandthis change.

why need to set it to HDX ?

F3Joule: `dispatch_permit` retrieves the `account_currency` of the "from" account to determine the currency override. Previously, it always defaulted to HDX for all accounts. Now, it falls back to WETH for EVM accounts and HDX for native Substrate accounts.

In other words, if we want HDX to be used as the payment asset for EVM accounts, we must now explicitly set it.

Out of your question I can assume that this "fee fix" changes expected flow/logic a bit, that is also a reason why I've marked this test as questionable, because it looks like:
https://github.com/galacticcouncil/hydration-node/blob/121af97c77ccecbc9f360cb6a832ed07ff31bed7/integration-tests/src/evm_permit.rs#L169-L170

enthusiastmartin: now i am confused. i thought that the problem was that for evm accounts - it was always WETH. and that's what was the issue. 
now, if nothing is set and it is evm account - it should be still WETH, right ? there is no change there.

F3Joule: > now, if nothing is set and it is evm account - it should be still WETH, right ? there is no change there.

That's what is questionable in this test.
If no currency set and it is evm account - it should be still WETH, then why _dispatch permit fee should be paid in hdx when no currency is set_ (referring to the test name)?

mrq1911: i think we changed that behaviour, so test should be changed

enthusiastmartin (approved):

PR #1192, opened 2025-10-02
feat: make aave executor safe again

enthusiastmartin:
This PR only makes the aave executor safer, by ensurng correct vec lengths and return values.

### DIFF
--- a/runtime/hydradx/src/evm/aave_trade_executor.rs
+++ b/runtime/hydradx/src/evm/aave_trade_executor.rs
@@ -7,6 +7,7 @@ use crate::{Runtime, Vec};
 use codec::{Decode, Encode, MaxEncodedLen};
 use ethabi::{decode, ParamType};
 use evm::ExitReason::Succeed;
+use evm::ExitSucceed;
 use frame_support::dispatch::DispatchResult;
 use frame_support::ensure;
 use frame_support::pallet_prelude::TypeInfo;
@@ -185,7 +186,7 @@ where
 		let (res, reserves_data) = Executor::<T>::view(context, data, VIEW_GAS_LIMIT);
 
 		ensure!(
-			matches!(res, Succeed(_)),
+			matches!(res, Succeed(ExitSucceed::Returned)),
 			ExecutorError::Error("Failed to get reserves list".into())
 		);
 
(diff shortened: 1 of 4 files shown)


runtime/hydradx/src/evm/aave_trade_executor.rs
```diff
@@ -195,6 +196,9 @@ where
 		let decoded = decode(&param_types, reserves_data.as_ref())
 			.map_err(|_| ExecutorError::Error("Failed to decode reserves list".into()))?;
 
+		// ensure sufficient length
+		ensure!(decoded.len() >= 1, ExecutorError::Error("Empty reserve list".into()));
```
martinfridrich:
```suggestion
		ensure!(decoded.len() == param_types.len(), ExecutorError::Error("Empty reserve list".into()));
```

runtime/hydradx/src/evm/aave_trade_executor.rs
```diff
@@ -240,6 +244,12 @@ where
 		let decoded = decode(&param_types, reserve_data.as_ref())
 			.map_err(|_| ExecutorError::Error("Failed to decode reserve data".into()))?;
 
+		// Ensure sufficient length
+		ensure!(
+			decoded.len() >= param_types.len(),
```
martinfridrich:
```suggestion
			decoded.len() == param_types.len(),
```

martinfridrich (approved):
I left 2 suggestions. If you disagree or there is reason why not to do it that way feel free to merge it as is, looks good otherwise.

runtime/hydradx/src/evm/aave_trade_executor.rs
enthusiastmartin: i considered it. not sure if there can be more or not. so feel like having >= does not hurt.

runtime/hydradx/src/evm/aave_trade_executor.rs
enthusiastmartin: same as previsou comment

runtime/hydradx/src/evm/aave_trade_executor.rs
enthusiastmartin: in the end,  changed to ==. decode cannot return more than asked.

runtime/hydradx/src/evm/aave_trade_executor.rs
enthusiastmartin: changed to ==

PR #1193, opened 2025-10-02
feat: withdraw and deposit erc20, XCM Aliasers

cl0w5:
This PR:
* adds impl for withdraw and deposit for erc20, allowing them to be used as a fee payment currency
* Adds AssetHub as a Trusted Aliaser https://github.com/galacticcouncil/hydration-node/pull/1226

To be resolved in a separate PR with ecosystem tests:
* the changes in the conversion `Location <> AssetId` to always use GeneralIndex as AssetId on local assets

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5537,7 +5537,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "353.0.0"
+version = "354.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
@@ -14634,7 +14634,7 @@ checksum = "48fd7bd8a6377e15ad9d42a8ec25371b94ddc67abe7c8b9127bec79bebaaae18"
 
 [[package]]
 name = "runtime-integration-tests"
-version = "1.53.1"
+version = "1.53.2"
 dependencies = [
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-parachain-system",
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "runtime-integration-tests"
-version = "1.53.1"
+version = "1.53.2"
 description = "Integration tests"
 authors = ["GalacticCouncil"]
 edition = "2021"
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx-runtime"
-version = "353.0.0"
+version = "354.0.0"
 authors = ["GalacticCouncil"]
 edition = "2021"
 license = "Apache 2.0"
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -122,7 +122,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("hydradx"),
 	impl_name: create_runtime_str!("hydradx"),
 	authoring_version: 1,
-	spec_version: 353,
+	spec_version: 354,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,


runtime/hydradx/src/evm/erc20_currency.rs
```diff
@@ -271,34 +272,40 @@ where
...
-			)
-		};
-
-		// And handle the transfer according to the token type
-		if AaveTradeExecutor::<T>::is_atoken(contract) {
-			AaveTradeExecutor::<T>::transfer(contract, from, to, amount, erc20_transfer)
-		} else {
-			erc20_transfer()
-		}
+		<Self as ERC20>::transfer(
```
enthusiastmartin, line 276: there was a different code path if the asset is atoken ? 

why removed ?

mrq1911: this was actually cause of issue we hotfixed, this is just removing unused code

runtime/hydradx/src/xcm.rs
```diff
@@ -135,6 +135,12 @@ parameter_types! {
 
 	pub UniversalLocation: InteriorLocation = [GlobalConsensus(RelayNetwork::get()), Parachain(ParachainInfo::parachain_id().into())].into();
 	pub AssetHubLocation: Location = (Parent, Parachain(ASSET_HUB_PARA_ID)).into();
+	pub HollarAssetLocation: Location = Location {
```
mrq1911: unused, remove

mrq1911 (approved):

runtime/hydradx/src/xcm.rs
```diff
@@ -223,12 +223,31 @@ mod remove_when_updating_to_stable2412 {
 	}
 }
 
+pub struct RestrictedAssetHubAliases;
+impl Contains<Location> for RestrictedAssetHubAliases {
+	fn contains(target: &Location) -> bool {
+		match target.unpack() {
+			// Allow system parachains under the Polkadot relay
+			(1, [Parachain(id)]) if matches!(id, 1001 | 1002 | 1004 | 1005) => true,
```
franciscoaguirre: I would just check if the para_id is less than 2000. All parachains with ids less than that are considered system chains

runtime/hydradx/src/xcm.rs
```diff
@@ -223,12 +223,31 @@ mod remove_when_updating_to_stable2412 {
...
+
+			// Allow Kusama relay itself
+			(1, [GlobalConsensus(Kusama)]) => true,
```
franciscoaguirre: Needs to be `parents: 2`

runtime/hydradx/src/xcm.rs
```diff
@@ -223,12 +223,31 @@ mod remove_when_updating_to_stable2412 {
...
+
+			// Allow Ethereum consensus
+			(1, [GlobalConsensus(Ethereum { .. })]) => true,
```
franciscoaguirre: Same

runtime/hydradx/src/xcm.rs
franciscoaguirre: You should also allow any interior location to Kusama. Specially accounts, since we want origin preservation to allow an account on Kusama Hub to execute something on Polkadot Hub for example

runtime/hydradx/src/xcm.rs
```diff
@@ -187,6 +187,69 @@ pub type Reserves = (
 pub type DynamicWeigher<RuntimeCall> =
 	WeightInfoBounds<crate::weights::xcm::HydraXcmWeight<RuntimeCall>, RuntimeCall, MaxInstructions>;
 
+// Types that exist in `xcm_builder` from `stable2412` onwards.
+mod remove_when_updating_to_stable2412 {
```
mrq1911, line 191: i guess we will keep this even when updating now

cl0w5: no, this only concerns the helper structs

franciscoaguirre: I named it that way so you can make sure to remove it when updating to `stable2412` 😄. Once in that version, you can import those helper structs from `xcm-builder`

runtime/hydradx/src/xcm.rs
```diff
@@ -226,16 +226,21 @@ mod remove_when_updating_to_stable2412 {
...
 		match target.unpack() {
 			// Allow system parachains under the Polkadot relay
-			(1, [Parachain(id)]) if matches!(id, 1001 | 1002 | 1004 | 1005) => true,
+			(1, [Parachain(id)]) if id < &2000 => true,
 
 			// Allow Kusama relay itself
 			(2, [GlobalConsensus(Kusama)]) => true,
 
 			// Allow Ethereum consensus
 			(2, [GlobalConsensus(NetworkId::Ethereum { .. })]) => true,
```
franciscoaguirre: We could also allow accounts from ethereum

PR #1223, opened 2025-10-17
fix: EVM nonce increasing twice

### LINKED ISSUE #1199: Bug: nonce incremented multiple times in batch
- [ ] #1223

F3Joule:
## Description
The Pull Request introduces changes to improve the handling of EVM transactions and their nonce behavior, particularly for batch calls and permit-based dispatches, as well as adjustments to avoid unintended nonce increments. It also adds new test cases to validate these updates

## Related Issue
Fixes #1199 

## Checklist:

- [X] I have added tests to cover my changes, regression test if fixing an issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5450,7 +5450,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx"
-version = "14.8.0"
+version = "14.8.1"
 dependencies = [
  "async-trait",
  "clap",
@@ -5598,7 +5598,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "371.0.0"
+version = "372.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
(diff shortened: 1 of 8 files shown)


integration-tests/src/evm.rs
```diff
@@ -95,6 +98,8 @@ mod account_conversion {
 	}
 
 	#[test]
+	// FIXME: will not pass, because nonce is increase by Executive;
+	// 	needs to mock SignedExtra and do UncheckedExtrinsic::new_signed and apply_extrinsic
 	fn bind_address_should_fail_when_nonce_is_not_zero() {
```
F3Joule: We already have unit test for this case:
https://github.com/galacticcouncil/hydration-node/blob/29f1b9e954d176d3168b26a0cd275bd6ffe7116d/pallets/evm-accounts/src/tests.rs#L102-L113

But to make a regression test correctly, we need to mock `Executive` and its dependencies (`UncheckedExtrinsic`, `SignedExtra`, etc.), because nonce isn't increased with just dispatching calls, it's increased for extrinsic by `CheckNonce` extra in `Executive`

enthusiastmartin (approved):

PR #1224, opened 2025-10-17
feat: HSM arbitrage improvements

### LINKED ISSUE #1221: HSM arb improvements
HSM auto-arbitrage fails in some scenario, mainly: 

- very small arb amounts
- hsm holding limit reached

We should make sure that we execute arbitrage only if it succeeds. 

- [ ] fix etsts

enthusiastmartin:
Fixed #1221

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5537,7 +5537,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "352.0.0"
+version = "353.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx-runtime"
-version = "352.0.0"
+version = "353.0.0"
 authors = ["GalacticCouncil"]
 edition = "2021"
 license = "Apache 2.0"
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -122,7 +122,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("hydradx"),
 	impl_name: create_runtime_str!("hydradx"),
 	authoring_version: 1,
-	spec_version: 352,
+	spec_version: 353,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,


integration-tests/src/hsm.rs
```diff
@@ -1527,15 +1527,129 @@ fn arbitrage_should_work() {
...
+		let (d, amount) = opp.unwrap();
+		assert_ok!(pallet_hsm::Pallet::<hydradx_runtime::Runtime>::simulate_arbitrage(
+			d, 2, amount
+		));
+
 		assert_ok!(HSM::execute_arbitrage(hydradx_runtime::RuntimeOrigin::none(), 2, None));
 		let final_hsm_dai_balance = Tokens::free_balance(2, &hsm_address);
 		let traded_amount = hsm_dai_balance - final_hsm_dai_balance;
-		assert_eq!(traded_amount, 999_642_225_291_583_959);
+		//assert_eq!(traded_amount, 999_642_225_291_583_959);
```
dmoka: we can remove this line if not needed

pallets/hsm/src/lib.rs
```diff
@@ -108,7 +107,7 @@ pub const ARBITRAGE_DIRECTION_BUY: u8 = 1;
 /// Direction for sell operations (more Hollar in pool, creates sell opportunities)
 pub const ARBITRAGE_DIRECTION_SELL: u8 = 2;
 
-pub const MIN_ARBITRAGE_AMOUNT: u128 = 100_000_000_000;
+pub const MIN_ARBITRAGE_AMOUNT: u128 = 1_000_000_000_000_000_000; // 1 HOLLAR
```
dmoka: maybe we can create a pallet config for this.

pallets/hsm/src/lib.rs
```diff
@@ -343,6 +344,8 @@ pub mod pallet {
 		FlashMinterNotSet,
 		/// Provided arbitrage data is invalid
 		InvalidArbitrageData,
+		/// Given arbitrage parameters, arbitrage cannot be succesfully executed.
```
dmoka:
```suggestion
		/// Given arbitrage parameters, arbitrage cannot be successfully executed.
```

pallets/hsm/src/lib.rs
```diff
@@ -1056,32 +1073,34 @@ where
...
-		// 3. Burn Hollar by calling GHO contract
-		Self::burn_hollar(final_hollar_amount)?;
+			// 3. Burn Hollar by calling GHO contract
+			Self::burn_hollar(final_hollar_amount)?;
 
-		// 5. Update Hollar amount received in this block
-		HollarAmountReceived::<T>::mutate(collateral_asset, |amount| {
-			*amount = amount.saturating_add(final_hollar_amount);
-		});
+			// 5. Update Hollar amount received in this block
```
dmoka: this should be step 4, or one step is missing
```suggestion
			// 4. Update Hollar amount received in this block
```

pallets/hsm/src/lib.rs
```diff
@@ -979,12 +996,12 @@ where
 	/// 6. Updating the amount of Hollar received in this block
 	///
 	/// Returns the final Hollar and collateral amounts traded.
-	#[require_transactional]
```
dmoka, line 982: in this case, can we have some undesired state-related side effects because we removed require_transactional?

mrq1911: why not simulate usign transactional instead of introducing custom logic?

enthusiastmartin: reworked. it is better now

enthusiastmartin:
> is it possible to reduce size of that snapshot? 100m added for one test seems excessive

removed

pallets/hsm/src/lib.rs
```diff
@@ -1249,67 +1285,30 @@ where
...
 
-			if let Some((arb_direction, amount)) = Self::find_arbitrage_opportunity(selected_collateral) {
-				let flash_amount = if arb_direction == ARBITRAGE_DIRECTION_BUY {
-					Some(amount)
-				} else {
-					None
-				};
+		// Select collateral asset based on block number
+		let bn: usize = block_number.saturated_into();
+		let idx = bn % collaterals.len();
```
dmoka, line 1338: I would add a guard statement for this too, and returning None if collaterals len = 0. Otherwise it can panic

enthusiastmartin: there is.

pallets/hsm/src/lib.rs
dmoka: this point is possibly reached only when there is some profit, but if there is bug or so, and user loses money in arbitrage, so the final balance decreases, then the profit will be 0 due to saturating math.

Maybe it is safer to use checked math, so we surely rule out this case, and let if fail

pallets/hsm/src/lib.rs
```diff
@@ -343,6 +346,8 @@ pub mod pallet {
 		FlashMinterNotSet,
 		/// Provided arbitrage data is invalid
 		InvalidArbitrageData,
+		/// Given arbitrage parameters, arbitrage cannot be successfully executed.
+		ArbitrageCannotBeExecuted,
```
dmoka: i can't find usage for this error, please remove if this is the case.

pallets/hsm/src/lib.rs
```diff
@@ -343,6 +346,8 @@ pub mod pallet {
 		FlashMinterNotSet,
 		/// Provided arbitrage data is invalid
```
dmoka, line 350: I would be more specific here as it just restates the error name.

Something like:

```suggestion
		/// Provided arbitrage data cannot be decoded or is malformed
```

pallets/hsm/src/lib.rs
dmoka: we also have this value in evm/precompiles/mod in `FLASH_LOAN_RECEIVER`, wondering if we could remove this duplication somehow.

pallets/hsm/src/lib.rs
```diff
@@ -827,18 +865,27 @@ pub mod pallet {
...
+					ensure!(arb_amount > 0, Error::<T>::NoArbitrageOpportunity);
+					// if provided, we know what to do, but need to verify the size is ok
+					let pool_state = Self::get_stablepool_state(collateral_info.pool_id)?;
+					ensure!(
+						Self::check_trade_size(collateral_asset_id, &collateral_info, &pool_state, arb_amount),
+						Error::<T>::NoArbitrageOpportunity
+					);
+					Arbitrage::HollarOut(arb_amount).into()
+				}
+				Some(Arbitrage::HollarIn(_)) => {
```
dmoka, line 880: what is the reason that the arb amount is ignored here? If intentional, I would add a comment to clarify the why

pallets/hsm/src/lib.rs
```diff
@@ -370,7 +375,29 @@ pub mod pallet {
...
+				if let Ok(_guard) = lock.try_lock() {
+					log::debug!(
+						target: "hsm::offchain_worker",
+						"Processing arbitrage opportunities at block: {:?}", block_number
+					);
+
+					if let Some(call) = Self::process_arbitrage_opportunities(block_number) {
+						if let Err(e) = SubmitTransaction::<T, Call<T>>::submit_unsigned_transaction(call.into()) {
+							log::error!(
+								target: "hsm::offchain_worker",
```
dmoka, line 391: this seems like an unexpected error, if so we could consider adding a  debug_assert so the fuzzer can find such cases

pallets/hsm/src/lib.rs
```diff
@@ -1592,6 +1630,30 @@ where
...
+	/// Returns:
+	/// - `Ok(())` if arbitrage is executed successfully
+	/// - `Err` if the arbitrage fails
+	///
+	/// Errors:
+	/// - `InvalidArbitrageData` if the data cannot be decoded or is malformed
+	/// - `AssetNotApproved` if the collateral asset is not approved
+	/// - `InvalidPoolState` if the pool ID doesn't match the collateral's configured pool
+	/// - Other errors from trade execution
 	pub fn execute_arbitrage_with_flash_loan(account: EvmAddress, loan_amount: Balance, data: &[u8]) -> DispatchResult
```
dmoka, line 1660: it might have been already discussed, but have we considered limiting the max expected profit? For example a possible oracle failure, bug or so could lead to unexpectedly large profits which could drain us...

But maybe it is not a good idea to cap max profit, so just thinking out of the box...

pallets/hsm/src/lib.rs
enthusiastmartin: there is a note just under it.
this would require to verify the amount in some way. 

i'd like to keep the original behaviour for now, hence it is not used. but using this approach will be clear idea what arb is executed by looking at the extrinisc parameters - whereas before it wasnt very clear.

pallets/hsm/src/lib.rs
enthusiastmartin: fuzzer cannot run offchain workers unfortunately.

pallets/hsm/src/lib.rs
enthusiastmartin: no, not really possible. it could prevent running arbitrage which is not desireable.

pallets/hsm/src/lib.rs
enthusiastmartin: agree. reworked

PR #1230, opened 2025-10-27
chore: safer dyn fees calculation + additional tests

enthusiastmartin:
This PR only adds a check of dynamic fees configuration to ensure correct min and max fees in omnipool's dynamic fee calculation.
and fixes validation. 
It is only for safety reasons ,because if max fee would be greater than min fee - it could panic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14634,7 +14634,7 @@ checksum = "48fd7bd8a6377e15ad9d42a8ec25371b94ddc67abe7c8b9127bec79bebaaae18"
 
 [[package]]
 name = "runtime-integration-tests"
-version = "1.53.2"
+version = "1.53.3"
 dependencies = [
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-parachain-system",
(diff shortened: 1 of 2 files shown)


integration-tests/src/dynamic_fees.rs
```diff
@@ -575,3 +576,290 @@ fn test_fees_update_after_buying_with_lrna_in_multi_blocks() {
...
+		let hdx_fee = hydradx_runtime::DynamicFees::current_fees(HDX).unwrap();
+		assert_eq!(dot_fee.asset_fee, Permill::zero());
+		assert_eq!(hdx_fee.protocol_fee, Permill::zero());
+
+		dave_hdx_after - dave_hdx_before
+	});
+
+	TestNet::reset();
+
+	// Second scenario: non-zero fees (min = max)
```
F3Joule, line 752: Good since it works, but it would be clearer and easier to maintain if we split it into two separate tests: min=max=zero and min=max=fixed. At least to isolate potential failures so one doesn't hide the other until fixed

enthusiastmartin: the test needs both . because in the end , it compares and calcualte that correct fee was applied.

PR #1236, opened 2025-10-30
fix: hollar fee currency in evm

dmoka:
hollar as fee payment currency works for substrate wallet, but doesn't work for EVM wallet.

In TransferEvmFees we use FungibleCurrencies and that has the `burn_from` disabled, so no erc20 currency can be burned by using that.

As we recently added support for erc20 depost/withdraw, we can use withdraw for burning.

### DIFF
--- a/pallets/currencies/src/fungibles.rs
+++ b/pallets/currencies/src/fungibles.rs
@@ -10,7 +10,7 @@ use orml_traits::MultiCurrency;
 #[cfg(any(feature = "try-runtime", test))]
 use sp_runtime::traits::Zero;
 use sp_runtime::traits::{CheckedSub, Get};
-use sp_runtime::{DispatchError};
+use sp_runtime::DispatchError;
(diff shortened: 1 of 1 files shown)


pallets/currencies/src/fungibles.rs
```diff
@@ -264,7 +265,13 @@ where
 			<T::NativeCurrency as fungible::Mutate<T::AccountId>>::mint_into(who, amount.into()).into()
 		} else {
 			match T::BoundErc20::contract_address(asset) {
-				Some(_) => fail!(Error::<T>::NotSupported),
+				Some(contract) => {
+					let old_balance = Self::balance(asset, who);
+					T::Erc20Currency::deposit(contract, who, amount)?;
+					let new_balance = Self::balance(asset, who);
+					let actual = new_balance.saturating_sub(old_balance);
```
dmoka: i would propose to use checked math to prevent any unexpeted behaviours

PR #1237, opened 2025-10-31
fix: include dust whitelist for dust account extrinsic

dmoka:
Currently we only prevent dust of accounts stored in duster accont storage.

This is not correct, we should include all whitelisted entried, declared in our runtime implementation.

We also add the HOLDING_ADDRESS used for mint/burn of erc20 to the dust whitelist

- [x] generate and grab new benchmark

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9150,7 +9150,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-currencies"
-version = "3.3.1"
+version = "3.4.0"
 dependencies = [
  "frame-support",
  "frame-system",
--- a/integration-tests/src/aave_router.rs
+++ b/integration-tests/src/aave_router.rs
@@ -4,18 +4,21 @@ use crate::dca::create_schedule;
 use crate::dca::schedule_fake_with_sell_order;
 use crate::liquidation::supply;
 use crate::polkadot_test_net::*;
+use crate::utils::accounts::*;
 use frame_support::assert_ok;
 use frame_support::pallet_prelude::DispatchError::Other;
+use frame_support::pallet_prelude::ValidateUnsigned;
 use frame_support::storage::with_transaction;
 use frame_support::traits::OnInitialize;
 use frame_support::{assert_noop, BoundedVec};
 use hex_literal::hex;
 use hydradx_runtime::evm::aave_trade_executor::AaveTradeExecutor;
 use hydradx_runtime::evm::precompiles::erc20_mapping::HydraErc20Mapping;
+use hydradx_runtime::evm::precompiles::{CALLPERMIT, DISPATCH_ADDR};
 use hydradx_runtime::evm::Erc20Currency;
 use hydradx_runtime::{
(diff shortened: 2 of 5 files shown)


pallets/duster/src/lib.rs
```diff
@@ -91,7 +91,8 @@ pub mod pallet {
 		type Erc20Support: hydradx_traits::evm::Erc20Inspect<Self::AssetId>
 			+ hydradx_traits::evm::Erc20OnDust<Self::AccountId, Self::AssetId>;
 
-		/// Duster for accounts with AToken dusts
+		/// Whitelist for dust removal, used to check if an account should be excluded from dusting
+		type DustRemovalWhitelist: Contains<Self::AccountId>;
```
enthusiastmartin: the naming and description is bit confusing and feels like doing same thing somewhere else. 

i would suggest to call it differently, eg  ExtendedWhitelist

pallets/duster/src/lib.rs
```diff
@@ -173,7 +174,7 @@ pub mod pallet {
 			ensure_signed(origin)?;
 
 			ensure!(
-				Self::whitelisted(&account).is_none() && account != T::TreasuryAccountId::get(),
+				!T::DustRemovalWhitelist::contains(&account),
```
enthusiastmartin: also this is dangerous and defeats  the purpose

because if the integration would not include anything, you would also skip the pallet's whitelist storage.

also there are two ways to check is account is whitelist , which is not good as well.

there is a Contain impl for the duster pallet. and there is a special struct in runtime contains again list of accounts. 

there should be only one way to check if an account is whitelist account that cannot be dusted, I suggest to leave this job for the Duster palette. That means we will use the implementation of contain trait that is done in this Duster palette. We will get rid of the dust white list structure  and contains impl from the runtime and we will do the extended white list like I suggested in the previous comment. 

ExtendedWhitelist can be Get<Vec<AccountId>> 

so the implementation of the contain ( only in the duster pallet ) would be 

T::ExtendedWhitelist::get().contains(acc) || AccountWhitelist::<T>::contains_key(acc)

This will be the only way to check if the account is wirelisted and cannot be dusted.

runtime/hydradx/src/assets.rs
```diff
@@ -678,6 +678,7 @@ impl Contains<AccountId> for DustRemovalWhitelist {
 		get_all_module_accounts().contains(a)
 			|| HSM::is_flash_loan_account(a)
 			|| pallet_duster::DusterWhitelist::<Runtime>::contains(a)
+			|| a == &EVMAccounts::account_id(crate::evm::HOLDING_ADDRESS)
 	}
 }
```
enthusiastmartin: This will not be here, there will be a the list of extra accounts that are not stored in the Duster palette storage. So, there will be a implementation of `Get`  trait that will return a list of accounts (hard-coded accounts from the runtime). It won't include anything related to Duster storage; it will just provide an extended list of accounts that will be checked but not here but in the Duster palette.

enthusiastmartin (changes requested):

PR #1239, opened 2025-10-31
feat: CurrencyIdConvert

cl0w5:
This PR refactors CurrencyIdConvert used in XCM:
* local locations (parents: 0) also resolve to GeneralIndex <> AssetId
* non-local locations (parents > 0) resolve to the location stored in the asset registry

### DIFF
--- a/runtime/hydradx/src/xcm.rs
+++ b/runtime/hydradx/src/xcm.rs
@@ -486,18 +486,18 @@ use primitives::constants::chain::CORE_ASSET_ID;
 
 impl Convert<AssetId, Option<Location>> for CurrencyIdConvert {
 	fn convert(id: AssetId) -> Option<Location> {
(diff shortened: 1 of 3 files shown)


runtime/hydradx/src/xcm.rs
```diff
@@ -486,45 +486,39 @@ use primitives::constants::chain::CORE_ASSET_ID;
...
-				if let Some(location) = loc {
-					location.into()
-				} else {
-					None
-				}
+				// Return default location with GeneralIndex for local assets or
+				// assets without stored location
+				Some(Location {
+					parents: 0,
+					interior: X1(Arc::new([GeneralIndex(id.into())])),
```
enthusiastmartin, line 499: this can be dangerous. let's think about it.

PR #1245, opened 2025-11-05
fix: add an alias for benchmarks

### LINKED ISSUE #1244: AliasOrigin instruction has weight set to max
Since AliasOrigin was not enabled before, its weight was set to max: https://github.com/galacticcouncil/hydration-node/blob/76236c65046bfe54441cfa959b407ce10912cfac/runtime/hydradx/src/weights/xcm/mod.rs#L289

Now that the runtime has an Aliasers configuration and we intend to use it, we should benchmark it and give it a weight.

franciscoaguirre:
Fixes #1244

I return Ok from the function used in benchmarking AliasOrigin so we don't get Weight::MAX.
Also added a new barrier to allow receiving AliasOrigin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5535,7 +5535,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "357.0.0"
+version = "358.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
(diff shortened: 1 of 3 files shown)


dmoka:
to make the CI green, could you please do a `cargo fmt`, bump runtime main version, and also fix clippy warnings? As for checking clippy, you can use `make clippy-all`. Thank you.

Issue #1227, opened 2025-10-23
Improved Peg Calculation Method for Stableswap Pools


enthusiastmartin:
Suggestion to  modify the peg recalculation mechanism in pegged stableswap pools to achieve smoother peg transitions. Current observations indicate that peg updates are not very smooth despite utilizing the minimum permissible max_peg_update value. As seen in this pic.

<img width="2724" height="916" alt="Image" src="https://github.com/user-attachments/assets/42cb0d64-7f8c-4fce-956e-493a28ea1cf0" />

<img width="1089" height="457" alt="Image" src="https://github.com/user-attachments/assets/ff0d4f95-5fcd-4865-8b59-4fed0650d7b9" />

### Current peg update mechanism
Stableswap pools with assets pegged to specific oracle values must recalculate the current peg, adjusting it toward the latest oracle-provided target peg. This recalculation occurs lazily—triggered only when pool-modifying operations (e.g., trades or liquidity adjustments) are executed.

The problem is how this is done. 

The peg source configuration includes a max_peg_update parameter that constrains the maximum peg adjustment per block. The current formula is:

The formula is: current_peg + (peg_diff * max_peg_update * delta_blocks) 

delta_blocks is number block between current block and the block where the target peg oracle value was updated. 

The `delta_blocks` variable introduces unintended acceleration in peg updates. As the time since the last oracle update increases, the permitted adjustment magnitude grows proportionally, resulting in larger-than-desired peg movements, as illustrated below:

<img width="1934" height="1200" alt="Image" src="https://github.com/user-attachments/assets/224b5c0b-38b7-4da4-98ea-3f59f3808fed" />

Green line is target peg, that constant, updated as block 0
Yellow is current behavior
Pink/Purple is the proposed change.

### The new peg calculation

Revised formula to recalculate a peg:

current_peg + (peg_diff * max_peg_update * block_diff_since_current_update)

#### Key Modification
The critical change is the block differential calculation. Instead of measuring the time elapsed since the oracle's last update, the formula now measures the time elapsed since the current peg's last recalculation. This ensures consistent, predictable peg adjustments independent of oracle update frequency.

### Implementation details
The timestamp (or block number) of the last current peg update is not currently persisted in the pool state. Implementation requires:

- Add a field to store the block number of the last peg recalculation
- corresponding runtime migration to initialize this field for existing pools - the peg should be recalculated in the migration and set the current block
[...]

PR #1247, opened 2025-11-05
feat: stableswap improve peg calculation

martinfridrich:
This PR fixes: #1227

Notes:
* as discussed in the chat this PR is using modified formula from current implementation, not the one from linked issue. 
 used formula: `current_peg + (current_peg * max_peg_update * (current_block - current_peg_updated_at)`
* trade fees are tied to peg's recalculation and as a result of this PR they are recalculated only once per block

Wasm [PR](https://github.com/galacticcouncil/HydraDX-wasm/pull/44)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5535,7 +5535,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "365.0.0"
+version = "366.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx-runtime"
-version = "365.0.0"
+version = "366.0.0"
 authors = ["GalacticCouncil"]
 edition = "2021"
 license = "Apache 2.0"
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -122,7 +122,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("hydradx"),
 	impl_name: create_runtime_str!("hydradx"),
 	authoring_version: 1,
-	spec_version: 365,
+	spec_version: 366,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,


pallets/stableswap/src/types.rs
```diff
@@ -36,6 +45,7 @@ pub struct PoolInfo<AssetId, BlockNumber> {
 	pub initial_block: BlockNumber,
 	pub final_block: BlockNumber,
 	pub fee: Permill,
+	pub pegs_info: Option<PegUpateInfo<BlockNumber>>,
```
enthusiastmartin: we dont want to store this here. 

in fact , we dont need to preserve the fee across the blocks

i would put the updated_at info inside PoolPegs storage, basically alongside the peg current values. that's info is more relevant there. 

and the fee can be temporary storage for just the one block and cleared afterwards.

pallets/stableswap/src/types.rs
```diff
@@ -182,16 +192,17 @@ impl<AssetId> PoolPegInfo<AssetId> {
...
 
 #[derive(Encode, Decode, Clone, RuntimeDebug, TypeInfo, MaxEncodedLen)]
-pub struct PoolSnapshot<AssetId> {
+pub struct PoolSnapshot<AssetId, BlockNumber> {
 	pub assets: BoundedVec<AssetId, ConstU32<MAX_ASSETS_IN_POOL>>,
 	pub reserves: BoundedVec<AssetReserve, ConstU32<MAX_ASSETS_IN_POOL>>,
 	pub amplification: u128,
 	pub fee: Permill,
 	pub pegs: BoundedVec<PegType, ConstU32<MAX_ASSETS_IN_POOL>>,
+	pub pegs_info: Option<PegUpateInfo<BlockNumber>>,
```
enthusiastmartin: we dont also need this in the snapshot. 

you can simply add just one extra field- : block_fee  - instead.

math/src/stableswap/math.rs
```diff
@@ -1017,12 +1017,13 @@ pub fn calculate_spot_price_between_two_stable_assets(
 
 pub fn recalculate_pegs(
```
enthusiastmartin, line 1018: surprised there is no change in any unit tests. 

likely, we are missing some unit tests. can we add ?

pallets/stableswap/src/lib.rs
```diff
@@ -1325,6 +1334,7 @@ pub mod pallet {
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_finalize(_n: BlockNumberFor<T>) {
 			let _ = <PoolSnapshots<T>>::clear(u32::MAX, None);
+			let _ = <BlockFee<T>>::clear(u32::MAX, None);
```
enthusiastmartin, line 1337: do we need to update weights ?

martinfridrich: done

math/src/stableswap/math.rs
martinfridrich: done

PR #1248, opened 2025-11-06
fix: use AssetHub as reserve for xtokens

cl0w5:
This PR sets AssetHub as ReserveProvider for DOT in orml_xtokens


enthusiastmartin:
closing at probably not needed

PR #1251, opened 2025-11-07
feat: add runtime api for dust-whitelisted accounts

dmoka:
We need this for the dust-bot to prevent failing/payed transactions when automatically dusting/cleanup accounts with dusts.

todo:
- [x] add integration test
- [ ] test out by doing runtime upgrade on fork


integration-tests/src/dust.rs
```diff
@@ -517,6 +517,54 @@ mod atoken_dust {
 	}
 }
 
+#[test]
+fn is_whitelisted_runtime_api_should_not_work_for_dustable_accounts() {
```
Roznovjak: API should always work :). Please rename the test

integration-tests/src/dust.rs
```diff
@@ -517,6 +517,54 @@ mod atoken_dust {
...
+	use pallet_duster::runtime_decl_for_duster_api::DusterApi;
```
Roznovjak: This is duplicated in every test. Can be moved to the top of the file, where all imports are.

pallets/duster/src/lib.rs
```diff
@@ -295,3 +295,14 @@ impl<T: Config> DustRemovalAccountWhitelist<T::AccountId> for Pallet<T> {
 		})
 	}
 }
+
+sp_api::decl_runtime_apis! {
```
Roznovjak: We usually define APIs in a separate crate. Check evm-accounts or currencies pallets.

Roznovjak (approved):

PR #1258, opened 2025-11-10
fix(PEPL): left bit shift can result in negative value

Roznovjak:
Fix left bit shift resulting in negative value.

### DIFF
--- a/runtime/hydradx/src/governance/tracks.rs
+++ b/runtime/hydradx/src/governance/tracks.rs
@@ -40,10 +40,10 @@ use pallet_referenda::Curve;
 const APP_LINEAR: Curve = Curve::make_linear(7, 7, percent(50), percent(100));
 const APP_LINEAR_FLAT: Curve = Curve::make_linear(4, 7, percent(50), percent(100));
 const APP_RECIP: Curve = Curve::make_reciprocal(1, 7, percent(80), percent(50), percent(100));
-const SUP_LINEAR: Curve = Curve::make_linear(7, 7, percent(0), percent(50));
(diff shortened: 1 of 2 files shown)


liquidation-worker-support/src/lib.rs
```diff
@@ -237,14 +237,14 @@ impl UserConfiguration {
 	/// Returns `true` if the user uses the asset as collateral.
 	/// The asset index is the position of the asset in the `fetch_reserves_list()` array.
 	pub fn is_collateral(&self, asset_index: usize) -> bool {
-		let bit_mask = U256::from(2 << (2 * asset_index));
+		let bit_mask = U256::from(2) << (2 * asset_index);
```
dmoka, line 240: less likely, but if asset_index is bigger than `usize::MAX/2`, it will panic.

I would use checked math, and return false if multiplication fails

same for is_debt function

Roznovjak: Later in this function we use  256 bit map we get from the money market. This map can store data only for 128 assets (2 bits per asset). So the MM contract has to have some check that it's not possible to have more than 128 assets in MM. Otherwise the data structures used is the contract would not be able to store all the data. And 128 is way less than `usize::MAX/2`. WDYT?

dmoka: you are probably right, it is very unlikely that it can panic, but we never know, and I feel it is better be safe than sorry, even if the check is duplicated.

mrq1911: yeah, MM is limited to 128 assets

mrq1911 (approved):

PR #1262, opened 2025-11-13
fix: clear dynamic fees and oracle storage when removing token from omnipool

### LINKED ISSUE #1151: Clear related asset storage when asset is removed from omnipool
As part of remove-token call, we should also clear storage related to the asset being removed. those are:

- dynamic fess
- oracle entries - needs confirmation
- ??!

Suggested implementation: 

Since omnipool is not aware of anything outside itself ( does not know that dynamic fees  or oracles store something), it will just send an info out that an asset has been removed from omnipool. 

- take advantage of existing OmnipoolHooks trait
- add a new method - on_asset_removed( asset Id ). It returns nothing. We dont want to fail, so no need an error.
- just call the hook when asset is removed from omnipool. 
- runtime implementation will take  care of cleanup
- this will allow us extending the cleanup of other stuff in future ( if necessary ) without making additional changes to the omnipool pallet.

Tests:
- needs a unit test to ensure the hook is called
- needs integration tests to ensure cleanup works as exepcted

Benchmarking
- need to benchmarks the cleanup implementation
- and add weight to the omnipool's extrinsic explicitly (similar to other hooks)

v9ai:
## Description

This PR adds storage cleanup functionality when removing a token from the omnipool.

When an asset is removed from the omnipool via `remove_token` extrinsic, the following related storage entries are now cleared:

- **Dynamic fees** - Removes `AssetFee` and `AssetFeeConfiguration` entries
- **Oracle entries** - Removes oracle data for all supported periods, accumulator entries, and whitelist entries

The implementation uses two new trait abstractions:

- `AssetFeesRemover` - for clearing dynamic fee storage
- `OracleAssetRemover` - for clearing oracle-related storage

## Related Issue

Closes #1151

## Motivation and Context

Previously, when an asset was removed from the omnipool, related storage entries in other pallets (dynamic fees and oracle) were not being cleaned up. This could lead to stale data accumulating in storage.

## How Has This Been Tested?

- Existing test `remove_token_should_clear_related_storage` validates the removal functionality
- Runtime adapters implement the cleanup logic for both dynamic fees and oracle entries
- Integration tests confirm proper cleanup of all related storage

## Checklist


enthusiastmartin:
Thanks for your contribution. 

Idea is good, but it is unnecessary explicit.  

let me write some detailed "spec" in the issue. so if you want, you can rework it.

runtime/hydradx/src/assets.rs
```diff
@@ -627,6 +627,9 @@ parameter_types! {
 	pub MaxAllowedPriceDifferenceForBifrostOracleUpdate: Permill = Permill::from_percent(10);
 }
 
+/// Type alias for SupportedPeriods to use in OracleAssetRemover
+pub type SupportedOraclePeriods = SupportedPeriods;
```
enthusiastmartin: can be removed.

pallets/omnipool/src/lib.rs
```diff
@@ -238,7 +239,7 @@ pub mod pallet {
 	#[pallet::storage]
 	/// State of an asset in the omnipool
 	#[pallet::getter(fn assets)]
-	pub(super) type Assets<T: Config> = StorageMap<_, Blake2_128Concat, T::AssetId, AssetState<Balance>>;
+	pub type Assets<T: Config> = StorageMap<_, Blake2_128Concat, T::AssetId, AssetState<Balance>>;
```
enthusiastmartin: is this change necessary?

pallets/omnipool/src/lib.rs
```diff
@@ -1615,6 +1616,10 @@ pub mod pallet {
 			T::Currency::withdraw(T::HubAssetId::get(), &Self::protocol_account(), asset_state.hub_reserve)?;
 			T::Currency::transfer(asset_id, &Self::protocol_account(), &beneficiary, asset_state.reserve)?;
 			<Assets<T>>::remove(asset_id);
+
+			// Clear related storage entries via hooks
```
enthusiastmartin: no need to for a comment here. it is clear we call the hook . 

nad moreover, it is not certain this hooks only clears stuff. in theory, i can do whatever we want.

pallets/omnipool/src/traits.rs
```diff
@@ -73,6 +73,14 @@ where
 		fee_account: AccountId,
 		amount: Balance,
 	) -> Result<Option<(Balance, AccountId)>, Self::Error>;
+
+	/// Called when an asset is removed from the Omnipool.
+	/// Allows cleanup of any related storage (fees, oracle entries, etc.).
+	/// Returns the weight consumed by the cleanup operations.
+	fn on_asset_removed(asset_id: AssetId) -> Weight;
```
enthusiastmartin: i believe it does not need to return the weight here - it does not need to return anything.

i see why you did - but in this case, we dont need

pallets/omnipool/src/lib.rs
v9ai: reverted, but build fails, what would u suggest?

image

v9ai: fixed with other approach

runtime/hydradx/src/assets.rs
v9ai: removed, thx

pallets/omnipool/src/lib.rs
v9ai: removed, thank you

pallets/omnipool/src/traits.rs
v9ai: appreciate, refactored

runtime/adapters/src/lib.rs
```diff
@@ -513,8 +514,53 @@ where
...
+		pallet_dynamic_fees::AssetFee::<Runtime>::remove(asset_id);
+		pallet_dynamic_fees::AssetFeeConfiguration::<Runtime>::remove(asset_id);
+
+		// Clear oracle entries for this asset paired with LRNA (hub asset)
+		let hub_asset = Lrna::get();
+		let assets = if asset_id < hub_asset {
+			(asset_id, hub_asset)
+		} else {
+			(hub_asset, asset_id)
+		};
```
Roznovjak:
```suggestion
		let assets = pallet_ema_oracle::ordered_pair(asset_id, hub_asset);
```

runtime/adapters/src/lib.rs
```diff
@@ -513,8 +514,53 @@ where
 		MC::transfer(Lrna::get(), &fee_account, &ProtocolFeeRecipient::get(), amount)?;
 		Ok(Some((amount, ProtocolFeeRecipient::get())))
 	}
-}
 
+	fn on_asset_removed(asset_id: AssetId) {
```
Roznovjak, line 524: Maybe it would be better to create 2 new functions for this in the corresponding pallets and benchmark it there. @enthusiastmartin WDYT?

Roznovjak: These functions should also emit relevant events: `RemovedFromWhitelist` and `AssetFeeConfigRemoved`

pallets/omnipool/src/tests/remove_token.rs
```diff
@@ -205,3 +205,93 @@ fn remove_token_should_burn_hub_asset() {
...
+			));
+
+			assert_ok!(Omnipool::remove_token(RuntimeOrigin::root(), 1000, 1234),);
+
+			assert!(Assets::<Test>::get(1000).is_none());
+		});
+}
+
+#[test]
+fn remove_token_should_call_on_asset_removed_hook() {
```
Roznovjak, line 210: We try to use the AAA pattern (arrange, act, assert) in our test. At least in new tests. Example can be found here https://github.com/galacticcouncil/hydration-node/blob/dadf616ee0ae23bf58863a3fe39ac8a7a416b417/pallets/omnipool/src/tests/buy.rs#L22
Just add these 3 comments.

pallets/omnipool/src/tests/remove_token.rs
```diff
@@ -205,3 +205,93 @@ fn remove_token_should_burn_hub_asset() {
 			assert_eq!(lrna_issuance_after, lrna_issuance - state.hub_reserve);
 		});
 }
+
+#[test]
+fn remove_token_should_clear_related_storage() {
```
Roznovjak: This test can be removed. It's already tested here https://github.com/galacticcouncil/hydration-node/blob/dadf616ee0ae23bf58863a3fe39ac8a7a416b417/pallets/omnipool/src/tests/remove_token.rs#L93

node/src/cli.rs
```diff
@@ -113,6 +113,7 @@ pub enum Subcommand {
 	Revert(sc_cli::RevertCmd),
 
 	/// The custom benchmark subcommmand benchmarking runtime pallets.
+	#[cfg(feature = "runtime-benchmarks")]
```
Roznovjak, line 120: Please also bump the version of all the crates that has been changed + runtime version.

integration-tests/src/omnipool_remove_token.rs
```diff
@@ -0,0 +1,354 @@
...
+				pallet_ema_oracle::Oracles::<hydradx_runtime::Runtime>::get((*b"omnipool", assets, period));
+			assert!(
+				oracle_entry.is_none(),
+				"Oracle entry should be cleared for DOT after removal for period {:?}",
+				period
+			);
+		}
+
+		// Verify whitelist entry is removed
+		let whitelist = pallet_ema_oracle::WhitelistedAssets::<hydradx_runtime::Runtime>::get();
```
Roznovjak: check for Accumulator storage is missing

integration-tests/src/omnipool_remove_token.rs
```diff
@@ -0,0 +1,354 @@
...
+		assert!(
+			!whitelist.contains(&(*b"omnipool", assets)),
+			"DOT should be removed from oracle whitelist"
+		);
+	});
+}
+
+#[test]
+fn remove_token_should_clear_both_fees_and_oracle_entries() {
```
Roznovjak: I would merge `remove_token_should_clear_dynamic_fees_storage` and `remove_token_should_clear_dynamic_fees_storage` into this one test

enthusiastmartin:
@nicolad can you address the review comments and resolve some merge conflicts ?

v9ai:
> @nicolad can you address the review comments and resolve some merge conflicts ?

✅

Issue #1164, opened 2025-08-04
stableswap asset order dependency


enthusiastmartin:
Currently - when a new pool is created, it sorts the order of asset in pool by their asset id.

When a pool with pegs - the list of peg sources might not be in the same order as pool assets after the sort. 

That's is not an issue with existing pools but if not careful, might present a problem.

Suggestions to fix:
1. Remove the reordering of assets 
2. Order the list of peg sources if needed

The order of asset is likely only due to historic reasons, and it is no longer needed. Needs to be verified.

PR #1266, opened 2025-11-17
fix(stableswap): reorder peg sources to match sorted assets during pool creation

v9ai:
When creating a stableswap pool with pegs, assets are sorted by ID for consistent ordering, but the corresponding peg sources and current pegs were not reordered to match. This caused pegs to be associated with the wrong assets after sorting.

The fix creates an index mapping before sorting and uses it to reorder both `peg_info.source` and `peg_info.current` arrays to maintain correct asset-to-peg associations.

## Related Issue

Fixes: #1164

## Motivation and Context

When creating a pool with assets [3, 1, 2] and pegs [Peg3, Peg1, Peg2]:

- Assets get sorted to [1, 2, 3]
- Pegs stayed as [Peg3, Peg1, Peg2] (wrong!)
- Result: Asset 1 incorrectly gets Peg3 instead of Peg1

This caused incorrect price calculations and potential trading failures in stableswap pools using pegged assets.

## How Has This Been Tested?

- Added 7 comprehensive tests in `pallets/stableswap/src/tests/peg.rs` covering:
  - Unsorted assets with pegs
  - Reverse order assets
  - 2-asset and 5-asset pools
  - Complex ordering scenarios
  - Index-based peg retrieval validation
  - Asset-peg association verification
- All 170 existing stableswap tests pass
- Formatting and clippy checks pass

## Checklist

### DIFF
--- a/pallets/stableswap/src/lib.rs
+++ b/pallets/stableswap/src/lib.rs
@@ -1449,13 +1449,47 @@ impl<T: Config> Pallet<T> {
 
 		let block_number = T::BlockNumberProvider::current_block_number();
 
-		// Create a mapping from original asset position to sorted position
-		// to reorder peg sources correctly
-		let asset_indices: sp_std::vec::Vec<(T::AssetId, usize)> =
-			assets.iter().enumerate().map(|(idx, asset)| (*asset, idx)).collect();
-
 		let mut pool_assets = assets.to_vec();
-		pool_assets.sort();
+
+		// If peg_info is provided, sort assets together with their peg sources and current pegs
+		// to maintain correct associations
+		let sorted_peg_info = if let Some(p) = peg_info {
+			ensure!(p.current.len() == assets.len(), Error::<T>::IncorrectInitialPegs);
+			ensure!(p.source.len() == assets.len(), Error::<T>::IncorrectInitialPegs);
+
+			// Zip assets with their corresponding peg sources and current pegs, then sort together
+			let mut assets_with_pegs: sp_std::vec::Vec<(T::AssetId, PegSource<T::AssetId>, PegType)> =
+				pool_assets
+					.iter()
+					.cloned()
+					.zip(p.source.iter().cloned())
+					.zip(p.current.iter().cloned())
+					.map(|((asset, source), current)| (asset, source, current))
+					.collect();
+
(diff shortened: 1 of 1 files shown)


v9ai:
Please have a look here 
@martinfridrich 
@Roznovjak

pallets/stableswap/src/lib.rs
```diff
@@ -1496,7 +1534,14 @@ impl<T: Config> Pallet<T> {
 			} else {
 				return Err(Error::<T>::UnknownDecimals.into());
 			}
-			PoolPegs::<T>::insert(share_asset, p);
+
+			let reordered_peg_info = PoolPegInfo {
+				source: crate::types::BoundedPegSources::truncate_from(sorted_sources),
```
dmoka, line 1539: truncate_from silently removes value that doenst fit into the bounded collection.

while we are probably fine, i would be more explicit here, with something like:

```
source: BoundedPegSources::try_from(sorted_sources)
    .map_err(|_| Error::<T>::IncorrectInitialPegs)?,
```

So we get explicit error if something goes wrong.

pallets/stableswap/src/lib.rs
```diff
@@ -1496,7 +1534,14 @@ impl<T: Config> Pallet<T> {
...
+				max_peg_update,
+				current: crate::types::BoundedPegs::truncate_from(sorted_current),
+			};
```
dmoka, line 1542: AFAIK PoolPegInfo has a `updated_at` field as well, what is the reason we don't set that here?

pallets/stableswap/src/tests/peg.rs
```diff
@@ -918,7 +921,7 @@ fn remove_liquidity_for_one_asset_should_work_correctly_with_different_pegs() {
 			let bob_shares = Tokens::free_balance(pool_id, &BOB);
 			assert_eq!(bob_shares, 0);
 			let bob_b_balance = Tokens::free_balance(asset_b, &BOB);
-			assert_eq!(bob_b_balance, 9311925821564); // not same as python because fees are not recalculated in same block
+			assert_eq!(bob_b_balance, 9649832005958); // same as python
```
dmoka, line 924: what is the reason that now the actual value became the same as python?

PR #1267, opened 2025-11-19
test: reproduce failing set currency in utility batch

dmoka:
Since we introduced that assets with a DOT-XYK pool are also accepted as fee currencies, set_currency can no longer be misused with utility.batch

In this PR, we proves this in an integration test


enthusiastmartin (approved):

dmoka:
Tried to reproduce with this steps, but was not possible, it is because we introduced the EitherAcceptedCurrencyOrHasXYKDotPool condition when charging fee.

Steps:

1. take some mainnet snapshot - or use chopsticks
2. choose a fee currency - some asset that it is in omnipool and can be fee payment asset
3. remove it from omnipool and from accepted currencies - or maybe it is enough just to remove it from accepted currencies.
4. create a batch call - where the first call is set_currency with that chosen currency
5. execute batch call
6. i believe you will pay the tx in that currency. if you use just batch, the set currency would fail, but any other call in that brach can be sucesful.

PR #1271, opened 2025-11-20
feat: Remove XYK discount fee

### LINKED ISSUE #1208: remove xyk's discount parameter
the discount param has never been used, It cannot be even used. Does not do much anyway.

Dont break API, keep the discount parameter, but it will be ignored.

Roznovjak:
Remove discount fee and methods from AMM trait that are not needed.
Fixes #1208

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -193,7 +193,7 @@ pub mod pallet {
 
 		/// Max fraction of pool to buy in single transaction has been exceeded.
 		MaxOutRatioExceeded,
-		
+
 		/// Max fraction of pool to sell in single transaction has been exceeded.
 		MaxInRatioExceeded,
 


pallets/xyk/src/lib.rs
```diff
@@ -200,6 +197,7 @@ pub mod pallet {
 		FeeAmountInvalid,
 
 		/// Overflow
+		/// Not in use. The discount feature has been deprecated.
 		CannotApplyDiscount,
```
enthusiastmartin: so you can delete the error, no ? or is it because of the possible error indices? i dont think that 's an issue here.

pallets/xyk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-xyk"
-version = "7.0.0"
+version = "7.1.0"
```
enthusiastmartin: version 8 - this breaks API ( pallet's config, amm trait etc).

traits/src/lib.rs
enthusiastmartin: this no longer valid - it does not handle trades. 

why it cannot be remove completely ?  

i have the feeling that the remaining functions - there is already something similar providing this ? maybe not.

Roznovjak: I'm not aware of anything that provides these methods/info. It can be removed completely and the methods can be implemented directly in the pallets, but xyk-liquidity-mining tests would need significant refactor.

Roznovjak: renamed

pallets/xyk/Cargo.toml
Roznovjak: updated

pallets/xyk/src/lib.rs
Roznovjak: yes, I didn't want to change the indices. I've just noticed that there are already some errors in the pallet with the same message.

enthusiastmartin: you can remove it. i asked around - nothing really seems to be depend on it.

PR #1276, opened 2025-11-28
feat: ema oracle events

talhadaar:
Introduced EMA oracle event when price changes at the end of the block.

### DIFF
--- a/pallets/ema-oracle/src/lib.rs
+++ b/pallets/ema-oracle/src/lib.rs
@@ -184,7 +184,7 @@ pub mod pallet {
 		OracleUpdated {
 			source: Source,
 			assets: (AssetId, AssetId),
-			update: OracleEntry<BlockNumberFor<T>>,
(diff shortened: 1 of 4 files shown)


pallets/ema-oracle/src/lib.rs
```diff
@@ -435,11 +441,17 @@ impl<T: Config> Pallet<T> {
...
+				Self::update_oracle(src.clone(), assets.clone(), period, oracle_entry.clone());
 			}
 			// As we use (the old value of) the `LastBlock` entry to update the other oracles it
 			// gets updated last.
 			Self::update_oracle(src, assets, OraclePeriod::LastBlock, oracle_entry.clone());
+
+			Self::deposit_event(Event::<T>::OracleUpdated {
+				source: src,
+				assets,
+				update: oracle_entry,
```
enthusiastmartin: i believe we dont want to entry but the updated entry.

talhadaar: The idea was to emit event to say the oracle was 'updated by this much', we want to instead only emit the updated oracle?

PR #1278, opened 2025-12-05
feat: claim account

Roznovjak:


pallets/evm-accounts/src/lib.rs
```diff
@@ -314,10 +366,107 @@ pub mod pallet {
...
+
+		ensure!(
+			!frame_system::Pallet::<T>::account_exists(&account),
+			Error::<T>::AccountAlreadyExists
+		);
+
+		ensure!(
+			signature.verify(msg.as_slice(), &account.clone().into()),
+			Error::<T>::InvalidSignature
+		);
```
enthusiastmartin, line 457: move this as first check. let's ensure signature is valid first.

pallets/evm-accounts/src/lib.rs
```diff
@@ -376,3 +525,14 @@ where
...
+#[cfg(feature = "std")]
+/// Used for testing purposes.
+/// Signs a message for `claim_account` with a given pair of keys.
+pub fn sign_message<T: Config>(pair: sp_core::sr25519::Pair, account: &T::AccountId, asset_id: T::AssetId) -> Signature
+where
+	T::AccountId: AsRef<[u8; 32]> + frame_support::traits::IsType<AccountId32>,
+{
+	let signature = pair.sign(Pallet::<T>::create_claim_account_message(account, asset_id).as_slice());
+	Signature::Sr25519(signature)
+}
```
enthusiastmartin, line 540: why we need it here when it is only for testing purposes?

pallets/evm-accounts/src/lib.rs
```diff
@@ -314,10 +366,107 @@ pub mod pallet {
...
+			asset_id: T::AssetId,
+			signature: Signature,
+		) -> DispatchResult {
+			ensure_none(origin)?;
+
+			Self::verify_claim_account(&account, asset_id, signature)?;
+
+			Self::do_bind_evm_address(&account)?;
+
+			T::FeeCurrency::set(&account, asset_id)?;
```
enthusiastmartin, line 395: i am wondering if we should let it succeed even if this fails. 

and it is actually a little concern. 

reasoning is that there might be a situation where an asset that you have cannot be used as fee payment asset, but you stil want to bind and claim the account. 
this would prevent it - and you could just actually spam it forever for free, as it would pass the validate_unsigned but fails here on execution

pallets/evm-accounts/src/lib.rs
Roznovjak: so it can be used in integration tests.

enthusiastmartin (approved):

pallets/evm-accounts/src/lib.rs
infosec-us-team: Good instincts, @enthusiastmartin.

After your comment, @Roznovjak added a new commit to solve the particular case you mentioned, "*an asset that you have cannot be used as fee payment asset*", but it does not solve the primary concern, which is: *passing the `validate_unsigned` while failing execution*. As you mentioned, that would allow mass spam of unpaid transactions.

Another way to bypass @Roznovjak's fix/patch:

In `do_bind_evm_address`, the very first check is to ensure the address is not an EVM account:
https://github.com/galacticcouncil/hydration-node/blob/b36268566af0da286ae5297305bca7dab6e06477/pallets/evm-accounts/src/lib.rs#L418-L422

The `is_evm_account` function checks if the account ends and begins with some hardcoded values:

https://github.com/galacticcouncil/hydration-node/blob/b36268566af0da286ae5297305bca7dab6e06477/pallets/evm-accounts/src/lib.rs#L505-L508

https://github.com/galacticcouncil/hydration-node/blob/b36268566af0da286ae5297305bca7dab6e06477/pallets/evm-accounts/src/lib.rs#L481-L483

Validation can be passed while failing during execution by submitting an unsigned extrinsic  to:
https://github.com/galacticcouncil/hydration-node/blob/b36268566af0da286ae5297305bca7dab6e06477/pallets/evm-accounts/src/lib.rs#L390-L397

With an AccountId that fits the following two constraints:
```rust
&account_id[0..4] == b"ETH\0" && account_id[24..32] == [0u8; 8]
```

Because `!Self::is_evm_account(who.clone())` is enforced inside `do_bind_evm_address` but not inside `verify_claim_account`.

If there's something we are missing, please let us know. It will help us share better feedback (from a security standpoint) in future pull requests.

infosec-us-team: We are glad to see that after we took the time to drop feedback (out of goodwill towards Hydration), a new PR with more validations was merged. @enthusiastmartin @Roznovjak 

https://github.com/galacticcouncil/hydration-node/pull/1283

Roznovjak: Thank you for your time and effort. We appreciate it.

PR #1280, opened 2025-12-08
fix: prevent nonce reset

dmoka:


pallets/evm-accounts/src/lib.rs
```diff
@@ -321,6 +329,22 @@ impl<T: Config> Pallet<T> {
 	fn _is_evm_account(account_id: &[u8; 32]) -> bool {
 		&account_id[0..4] == b"ETH\0" && account_id[24..32] == [0u8; 8]
 	}
+
+	/// Checks if an account has been marked as an EVM account.
+	pub fn is_marked_as_evm_account(account: &T::AccountId) -> bool {
```
enthusiastmartin: no need to have special function for this.

pallets/evm-accounts/src/lib.rs
```diff
@@ -321,6 +329,22 @@ impl<T: Config> Pallet<T> {
...
+		EvmAccounts::<T>::contains_key(account)
+	}
+
+	/// Marks an account as an EVM account.
+	/// This should only be called once per account to avoid unnecessarily
+	/// increasing sufficients multiple times.
+	pub fn mark_as_evm_account(account: &T::AccountId) {
```
enthusiastmartin, line 489: i would also ensure it is evm acccount here. there is a helper is_evm_account 

add a debug assert there if it is not - to see if this can be reached somehow by non-evm accounts.

enthusiastmartin (changes requested):

runtime/hydradx/src/migrations/reset_max_nonces.rs
```diff
@@ -0,0 +1,1995 @@
...
+						log::info!(
+							target: "runtime::reset_max_nonces",
+							"Reset {} nonces so far...",
+							reset_count
+						);
+					}
+				} else {
+					skipped_count += 1;
+				}
+			}
```
enthusiastmartin: this is quite hard to read. 

i am not really sure why we need to extra storage to check if this was executed or not. 

i dont think it is necessary. 

suggestion to rewrite the code. 

```suggestion
			for (account_str, new_nonce) in AFFECTED_ACCOUNTS {
    let Ok(account_id) = AccountId::from_ss58check(account_str) else {
        log::warn!(
            target: "runtime::reset_max_nonces",
            "Failed to decode address: {account_str}"
        );
        skipped_count += 1;
        continue;
    };

    reads += 1;
    
    let was_reset = frame_system::Account::<T>::mutate(&account_id, |info| {
        (info.nonce == u32::MAX).then(|| info.nonce = *new_nonce).is_some()
    });

    match was_reset {
        true => {
            writes += 1;
            reset_count += 1;
            if reset_count % 100 == 0 {
                log::info!(
                    target: "runtime::reset_max_nonces",
                    "Reset {reset_count} nonces so far..."
                );
            }
        }
        false => skipped_count += 1,
    }
}
```

enthusiastmartin (changes requested):

PR #1295, opened 2025-12-17
feat: DCA with extra gas

dmoka:
todos:
- [x] test on zombienet with atoken
- [x] fix taking extra fee
- [ ] test by @Ben
- [ ] rebenchmark

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5598,7 +5598,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "379.0.0"
+version = "380.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx-runtime"
-version = "379.0.0"
+version = "380.0.0"
 authors = ["GalacticCouncil"]
 edition = "2021"
 license = "Apache 2.0"
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -122,7 +122,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("hydradx"),
 	impl_name: create_runtime_str!("hydradx"),
 	authoring_version: 1,
-	spec_version: 379,
+	spec_version: 380,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,


pallets/dca/src/lib.rs
```diff
@@ -553,6 +571,7 @@ pub mod pallet {
 			ScheduleOwnership::<T>::insert(who.clone(), next_schedule_id, ());
 			RemainingAmounts::<T>::insert(next_schedule_id, reserve_amount);
 			RetriesOnError::<T>::insert(next_schedule_id, 0);
+			ScheduleExtraGas::<T>::insert(next_schedule_id, 0); //TODO: we might dont need to set it as 0 is default?!
```
enthusiastmartin: no need

traits/src/evm.rs
```diff
@@ -122,3 +124,10 @@ pub trait Erc20OnDust<AccountId, CurrencyId> {
 		currency_id: CurrencyId,
 	) -> frame_support::dispatch::DispatchResult;
 }
+
+pub trait ExtraGasSupport {
+	fn set_extra_gas(gas: u64);
+	fn clear_extra_gas();
+
+	fn out_of_gas_error() -> DispatchError;
```
enthusiastmartin, line 136: add descriptions

runtime/hydradx/src/evm/evm_error_decoder.rs
```diff
@@ -23,6 +24,28 @@ impl Convert<CallResult, DispatchError> for EvmErrorDecoder {
...
+		// EVM Subcalls exits with Revert with empty data on gas exhaustion, so we check for that case
+		if let ExitReason::Revert(ExitRevert::Reverted) = call_result.exit_reason {
+			if call_result.value.is_empty() {
+				let threshold = call_result
+					.gas_limit
+					.checked_mul(U256::from(90))
+					.and_then(|v| v.checked_div(U256::from(100)))
+					.unwrap_or(U256::zero());
+
+				if call_result.gas_used >= threshold {
```
enthusiastmartin: this is actually possible to rewrite in simpler, cleaner way without divisions and checked math 

```suggestion
				if call_result.gas_used.saturating_mul(U256::from(100)) 
    >= call_result.gas_limit.saturating_mul(U256::from(90)) 
```

that should do the same .

pallets/dca/src/lib.rs
```diff
@@ -1210,6 +1236,20 @@ impl<T: Config> Pallet<T> {
 		}
 	}
 
+	/// Calculates trade weight including extra gas for a schedule.
+	fn get_trade_weight_with_extra_gas(schedule_id: ScheduleId, order: &Order<T::AssetId>) -> Weight {
```
enthusiastmartin: is this necessary ? 

wouldnt it be better just to include the extra weight implicitly in the existing get_trade_weight?

i dont think there is a scenario where you would not include the extra weight. you always want to include it. 

that means there will just one way to get it.

enthusiastmartin (changes requested):

pallets/dca/src/lib.rs
dmoka: During schedule creation, we use trade weight to calculate transacation fee for validation purposes. At this point we cant include the gas extra weight as we dont have any schedule id.

But you are right, i can merge the two functions to make it cleaner and more centralized.

pallets/dca/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-dca'
-version = "1.9.5"
+version = "1.9.6"
```
martinfridrich: This should be 2.0 IMO, you didn't change API but you changed internal behaviour in breaking way.

dmoka: lets bump the minor as its not about changing working behavior, but fixing failing behavior.

PR #1299, opened 2025-12-18
feat: Allowance for Multicurrency

iamyxsh:
## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5598,7 +5598,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "374.0.0"
+version = "375.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx-runtime"
-version = "374.0.0"
+version = "375.0.0"
 authors = ["GalacticCouncil"]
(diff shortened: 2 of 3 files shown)


runtime/hydradx/src/evm/precompiles/multicurrency.rs
```diff
@@ -273,29 +290,40 @@ where
...
+
+		if !spender_is_approved {
+			let allowed: Balance = pallet_evm_accounts::Pallet::<Runtime>::get_allowance(asset_id.into(), from, origin);
+
+			if allowed < amount {
+				return Err(revert("ERC20: insufficient allowance"));
+			}
+
+			if allowed != Balance::MAX {
+				pallet_evm_accounts::Pallet::<Runtime>::set_allowance(asset_id.into(), from, origin, allowed - amount);
```
enthusiastmartin: use saturating sub here.

runtime/hydradx/src/evm/precompiles/multicurrency.rs
```diff
@@ -273,29 +290,40 @@ where
...
+			);
+
+		if !spender_is_approved {
+			let allowed: Balance = pallet_evm_accounts::Pallet::<Runtime>::get_allowance(asset_id.into(), from, origin);
+
+			if allowed < amount {
+				return Err(revert("ERC20: insufficient allowance"));
+			}
+
+			if allowed != Balance::MAX {
```
enthusiastmartin, line 308: this is a problem

if you set allowance to max value, then the spender can use but because of this condition - it wont decrease

Issue #1294, opened 2025-12-17
add new extrinsic for removing stableswap liquidity from omnipool lp/omnipool lp in farm


DuriSK:
It would be great to have extrinsic(s) which will calculate for UI necessary balances when removing stableswap shares from either omnipool position or omnipool farming position.
This is currently handled in 3 signing steps.

Revert OmnipoolLiquidityMining.addLiquidityStableswapOmnipoolAndJoinFarms
depositId
list of farms (because you can be in multiple farms)

Omnipool liquidity (in farm)=>omnipoolLiquidityMining.withdrawShares(args)=>Omnipool LP position(not in farm)=>omnipool.removeLiquidityWithLimit(args)=>stableswap shares=>remove proportionally/single asset=> assets

example of proportional liquidity removal (from stableswap)l:

{
  "args": {
    "pool_id": "102",
    "share_amount": "9,868,770,921,034,374,665",
    "min_amounts_out": [
      {
        "assetId": "10",
        "amount": "4,917,091"
      },
      {
        "assetId": "22",
        "amount": "4,983,877"
      }
    ]
  },
  "method": "removeLiquidity",
  "section": "stableswap"
}

example of single side removal(from stableswap):

{
  "args": {
    "pool_id": "102",
    "asset_id": "22",
    "share_amount": "9,868,770,921,034,374,665",
    "min_amount_out": "9,898,643"
  },
  "method": "removeLiquidityOneAsset",
  "section": "stableswap"
}

1. omnipool position in farm => omnipool position (not in farm)=>stableswap shares=>asset
2. omnipool position (not in farm)=>stableswap shares=>asset

2 options how to withdaw shares from stablepool :
- proportionally
- single asset

Issue #1284, opened 2025-12-12
Add option to specify slippage limit to addLiquidityStableswapOmnipoolAndJoinFarms wrapper


DuriSK:
Comparing this function _to addLiquidityAndJoinFarms_ it would be useful if it would be possible to add slippage (min amoun) limit also to this stableswap function.
Add min_shares_limit

dmoka:
pub fn add_liquidity_stableswap_omnipool_and_join_farms(
			origin: OriginFor<T>,
			stable_pool_id: T::AssetId,
			stable_asset_amounts: BoundedVec<AssetAmount<T::AssetId>, ConstU32<MAX_ASSETS_IN_POOL>>,
			farm_entries: Option<BoundedVec<(GlobalFarmId, YieldFarmId), T::MaxFarmEntriesPerDeposit>>,
			min_shares_limit: Option<Balance>,
		)

dmoka:
Fixed it as part of this PR: https://github.com/galacticcouncil/hydration-node/pull/1300

PR #1300, opened 2025-12-19
feat: remove omnipool stable position

dmoka:
Fixes: https://github.com/galacticcouncil/hydration-node/issues/1294 and https://github.com/galacticcouncil/hydration-node/issues/1284

TODOS:

- [x] add new extrinsic
- [x] review
- [ ] test by FE on PASEO as we have PAPI there. We should do this when current release is done, as we dont want to add anytgin to paseo till then
- [ ] coordinate deployment with FE
- [x] add benchmark tests and rebench


pallets/omnipool-liquidity-mining/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-omnipool-liquidity-mining"
-version = "2.7.1"
+version = "2.8.0"
```
enthusiastmartin: it should be really be 3.0.0 because it breaks the api by adding extra parameter.

pallets/omnipool/src/lib.rs
```diff
@@ -2258,6 +2119,163 @@ impl<T: Config> Pallet<T> {
...
 	}
 
+	/// Internal method to remove liquidity with limit.
+	/// Returns the actual amount of asset transferred to the user.
+	///
+	/// This is the core logic for liquidity removal, extracted to allow
+	/// other pallets to call it and receive the transferred amount.
+	#[require_transactional]
+	pub fn do_remove_liquidity_with_limit(
+		origin: OriginFor<T>,
```
enthusiastmartin, line 2140: account instead of origin. origin can be ensured just before. 
also usage of this in the omnipool-lm favors this approach.

pallets/omnipool/src/lib.rs
```diff
@@ -2258,6 +2119,163 @@ impl<T: Config> Pallet<T> {
 		Ok(instance_id)
 	}
 
+	/// Internal method to remove liquidity with limit.
+	/// Returns the actual amount of asset transferred to the user.
+	///
+	/// This is the core logic for liquidity removal, extracted to allow
+	/// other pallets to call it and receive the transferred amount.
+	#[require_transactional]
+	pub fn do_remove_liquidity_with_limit(
```
enthusiastmartin: it does not need to be have the "with_limit" in the name. 

just call it "do_remove_liquidity".

same would apply for do_add_liquidity. 

the with_limit is legacy thing where we wanted to keep existing interface and provide new interface with slippage protection. but that interface is deprecated and can be really removed.

traits/src/stableswap.rs
```diff
@@ -9,6 +9,21 @@ pub trait StableswapAddLiquidity<AccountId, AssetId, Balance> {
...
+		share_amount: Balance,
+		min_amount_out: Balance,
+	) -> Result<Balance, DispatchError>;
+
+	fn remove_liquidity(
+		who: AccountId,
+		pool_id: AssetId,
+		share_amount: Balance,
+		min_amounts_out: Vec<AssetAmount<AssetId>>,
+	) -> Result<(), DispatchError>;
```
enthusiastmartin, line 26: this does not make much sense here.  the trait is called AddLiquidity.

pallets/omnipool-liquidity-mining/Cargo.toml
dmoka: does it really break the API? I mean, it is just an optional param in the end, AFAIK it doesn't break API.

pallets/omnipool/src/lib.rs
dmoka: the thing is that on_liquidity_changed requires origin, so i think we need to keep it as is, unfortunately.

traits/src/stableswap.rs
dmoka: good point. Since this trait is only used in omnipool liqudity mining, i would rename it to "StableswapLiquidityMutation'

pallets/omnipool-liquidity-mining/Cargo.toml
enthusiastmartin: yes. exactly the extra parameter is the breaking change. that would mean, whatever is using this must adjust. 

let's also ensure this change with FE.

pallets/omnipool/src/lib.rs
enthusiastmartin: ok. not a fan, but dont want to change anything which is not related.

pallets/liquidity-mining/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "4.4.8"
+version = "4.4.9"
```
martinfridrich: This should not be 4.5.0, you added feature(non-breaking change), it's not a patch

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -87,6 +87,7 @@ type PeriodOf<T> = BlockNumberFor<T>;
 pub mod pallet {
 	use super::*;
 	use frame_support::pallet_prelude::*;
+	use primitives::ItemId;
```
martinfridrich:
```suggestion
```

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -1076,6 +1084,100 @@ pub mod pallet {
...
+			ensure!(!stableswap_min_amounts_out.is_empty(), Error::<T>::NoAssetsSpecified);
+
+			if let Some(deposit_id) = deposit_id {
+				Self::ensure_nft_owner(origin.clone(), deposit_id)?;
+
+				let stored_position_id = OmniPositionId::<T>::get(deposit_id)
+					.ok_or(Error::<T>::InconsistentState(InconsistentStateError::MissingLpPosition))?;
+				ensure!(
+					stored_position_id == position_id,
+					Error::<T>::InconsistentState(InconsistentStateError::MissingLpPosition)
```
martinfridrich: This shouldn't be `InconsistentState` error. `InconsistentState` should be used for things that should never happen, it basically mean something is wrong with the chain.

This error can easily happen by user mixing input params(`position_id` and `deposit_id`) and it doesn't mean issue with the chain.

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -1076,6 +1084,100 @@ pub mod pallet {
...
+
+				Self::exit_farms(origin.clone(), deposit_id, yield_farm_ids)?;
+			}
+
+			let omnipool_position = OmnipoolPallet::<T>::load_position(position_id, who.clone())?;
+			let omnipool_shares_to_remove = omnipool_position.shares;
+			let stable_pool_id = omnipool_position.asset_id;
+
+			let actual_stable_shares_received = OmnipoolPallet::<T>::do_remove_liquidity(
+				origin.clone(),
```
martinfridrich:
```suggestion
				origin,
```

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -1076,6 +1084,100 @@ pub mod pallet {
...
+				position_id,
+				omnipool_shares_to_remove,
+				omnipool_min_limit,
+			)?;
+
+			if stableswap_min_amounts_out.len() == 1 {
+				let asset_amount = &stableswap_min_amounts_out[0];
+
+				T::Stableswap::remove_liquidity_one_asset(
+					who.clone(),
```
martinfridrich:
```suggestion
					who,
```

pallets/omnipool-liquidity-mining/Cargo.toml
martinfridrich: I agree with Martin, this is breaking change.

pallets/omnipool-liquidity-mining/src/lib.rs
```diff
@@ -1076,6 +1084,100 @@ pub mod pallet {
...
+				.saturating_add(<T as Config>::WeightInfo::price_adjustment_get().saturating_mul(T::MaxFarmEntriesPerDeposit::get() as u64))
+		})]
+		pub fn remove_liquidity_stableswap_omnipool_and_exit_farms(
+			origin: OriginFor<T>,
+			position_id: T::PositionItemId,
+			omnipool_min_limit: Balance,
+			stableswap_min_amounts_out: BoundedVec<AssetAmount<T::AssetId>, ConstU32<MAX_ASSETS_IN_POOL>>,
+			deposit_id: Option<DepositId>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin.clone())?;
```
martinfridrich:
```suggestion
			let who = if let Some(deposit_id) = deposit_id {
			  let who = Self::ensure_nft_owner(origin.clone(), deposit_id)?;
			  ... whole if bellow
			  who
			} else {
			  ensure_signed(origin.clone())?
			}
			
```
I would do this, it saves cloning and signature verification in case when `deposit_id` is provided.

pallets/omnipool-liquidity-mining/src/tests/remove_liquidity_stableswap_omnipool_and_exit_farms.rs
```diff
@@ -0,0 +1,861 @@
...
+					cap: DEFAULT_WEIGHT_CAP,
+					tradable: Tradability::default(),
+				}
+			);
+
+			let hdx_balance_after = Tokens::free_balance(HDX, &LP1);
+
+			// Verify user received HDX rewards (if any were generated)
+			// Note: rewards might be 0 in test depending on block time configuration
+			if hdx_balance_after > hdx_balance_before {
```
martinfridrich: From this test we don't know if rewards distribution are broken or not.
Especially we won't notice it in the future when we will change something.

It's up to you set up test's env the way you need to test it.

pallets/xyk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-xyk"
-version = "8.0.0"
+version = "8.0.1"
```
martinfridrich: Why did you bump this? I don't see any changes in xyk

runtime/hydradx/src/benchmarking/omnipool_liquidity_mining.rs
```diff
@@ -924,7 +925,171 @@ runtime_benchmarks! {
...
+			 Some(OmnipoolWarehouseLM::deposit_id() + 1)
+		} else {
+			None
+		};
+		let farms_to_join = if with_farms {
+			Some(farms.try_into().unwrap())
+		} else {
+			None
+		};
+		OmnipoolLiquidityMining::add_liquidity_stableswap_omnipool_and_join_farms(RawOrigin::Signed(lp_provider.clone()).into(),pool_id, added_liquidity.try_into().unwrap(), farms_to_join, None).unwrap();
```
martinfridrich, line 1060: Is this really worst case? 
I don't see LM's period change after this so no rewards will be claimed.

pallets/xyk/Cargo.toml
dmoka: yeah, not sure, CI noted it, you remember i asked you the same question half a year ago or so but we could not figoure out. It probably comes the the fact that xyk-lm and xy has connections.

dmoka: ok lets revert it, we can merge the PR anyway

PR #1308, opened 2025-12-24
fix: pin srtool docker image version to 1.81.0

1xGiraffe:
## Description

Fixes the srtool build workflow by using Docker directly instead of srtool-cli, with a pinned image version matching `rust-toolchain` (1.81.0).

srtool-cli ignores `rust-toolchain` and always uses the latest container image (1.88.0), causing build failures:

```
error: `#[no_mangle]` cannot be used on internal language items
```

### Changes

- Use `docker run` directly with pinned `paritytech/srtool:1.81.0` image
- Add `SRTOOL_VERSION` env variable for easy version management
- Update Makefile `srbuild` target to use Docker directly
- Remove `srtool_image` workflow input (see note below)

### Note on Removed Input

The `srtool_image` workflow input has been removed. It never worked as intended - it controlled the srtool-cli git tag (`cargo install --tag`), not the Docker image version.

This is **not classified as a breaking change** because:
1. This PR actually fixes the broken action
2. If `srtool_image` is specified via `workflow_dispatch`, it will simply be silently ignored by GitHub Actions

## Related Issue

Workaround for [chevdor/srtool-cli#40](https://github.com/chevdor/srtool-cli/issues/40):

> Currently the CLI always uses the latest available srtool container image. It ignores rust-toolchain.toml and there is no ability to specify what srtool container image using --image parameter.

## Motivation and Context

**Deterministic builds are critical for blockchain runtimes.** The `authorize_upgrade` hash must be reproducible so validators and the community can verify the runtime WASM matches the source code.

By pinning the Docker image version to match `rust-toolchain`, we ensure consistent, reproducible builds.

## How Has This Been Tested?

- Tested GitHub Actions workflow in a fork
- Local Docker build completes successfully with correct JSON output

## Checklist:


F3Joule (approved):
lgtm

1xGiraffe:
Good catch! Had I known about `SRTOOL_TAG` earlier, I probably would've gone that route since it requires fewer changes. The env is actually undocumented - at least I couldn't find any reference. Did you?

On reflection, using Docker directly might actually be better:

- **Security**: `srtool-cli` is maintained by an individual, while the Docker image is maintained by Parity. Minimizing supply chain dependencies might matter.
- **Reliability**: `SRTOOL_TAG` is undocumented and could change without notice
- **Performance**: No `cargo install` step in CI

That said, happy to switch to `SRTOOL_TAG` and back to `srtool-cli` if that's the preferred approach. Just give me a heads up and I'll update the PR.

PR #1314, opened 2025-12-31
fix(omnipool): allow tolerance for atoken fee taken

enthusiastmartin:
This fix allows accepted tolerance of 1 unit when a part of the fee is taken out and sent to staking/referral pots.


martinfridrich:
There are additional storage ops, don't forget to rebenchmark before merge

Issue #1336, opened 2026-01-19
Omnipool -  H2O Sell Routing Change


enthusiastmartin:
# H2O Sell Routing — Route to Treasury Account

## Overview

When a user sells H2O (hub asset) to the Omnipool, instead of adding the H2O amount to the traded asset's subpool hub reserve, **send the entire H2O amount to a designated treasury account**.

Note: H2O can only be sold to the Omnipool, not bought from it.

## Current Behavior

1. User initiates a trade (sell or buy) swapping H2O for another asset
2. The corresponding asset subpool is updated:
   * Hub reserve (H2O) increases
   * Asset reserve decreases

## New Behavior

1. Update the asset subpool as before, but **do not increase its hub reserve**
2. **Transfer the H2O amount to the treasury account**
3. The H2O leaves the Omnipool's hub reserve accounting entirely

### Affected functions

Selling H2O is handled in a separate code path. When the sell side of a trade is H2O (hub asset), one of these functions is called:

* `sell_hub_asset`
* `buy_asset_for_hub_asset`

### Configuration

Add a new config parameter to the Omnipool pallet:

```rust
/// Destination account when hub asset is sold
type HubDestination: Get<Self::AccountId>;
```

### Changes required (sell side)

1. Around line 1793 — where we calculate the new asset state after the trade
2. Before creating the new asset state:
   * Store the `delta_hub_reserve` value
   * Replace `delta_hub_reserve` in `state_changes` with `Increase(0)`
3. Create the new asset state as before (now with zero hub reserve change)
4. **Transfer the stored `delta_hub_reserve` amount of H2O from the Omnipool account to the treasury account**
5. Call `on_hub_asset_trade` hook with accurate pool change information:
   * For the traded asset subpool: report hub reserve change as 0

Similar approach for the buy side.

**Note:** The trade event does not change and still includes the full H2O amount traded.

### Unit tests

* Verify H2O is transferred to the treasury account, not added to any subpool
* Verify asset subpool reserve changes correctly (asset decreases, hub reserve unchanged)
* Verify the Omnipool account's H2O balance decreases by the correct amount
* Verify the treasury account's H2O balance increases by the correct amount
* Verify `on_hub_asset_trade` is called with correct parameters (zero hub change for traded asset)

### Integration tests

* End-to-end H2O sell trades
* Verify treasury account receives the correct H2O amount
* Verify traded asset subpool hub reserve remains unchanged

### Invariant

[...]

PR #1340, opened 2026-01-23
feat: route h2o trade to hdx subpool

dmoka:
implements https://github.com/galacticcouncil/hydration-node/issues/1336

### DIFF
--- a/pallets/omnipool/Cargo.toml
+++ b/pallets/omnipool/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-omnipool"
-version = "5.5.0"
+version = "6.0.0"
 authors = ['GalacticCouncil']
 edition = "2021"
 license = "Apache-2.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11004,7 +11004,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-omnipool"
-version = "5.5.0"
+version = "6.0.0"
 dependencies = [
  "bitflags 1.3.2",
  "frame-benchmarking",


pallets/omnipool/src/lib.rs
```diff
@@ -1608,19 +1625,43 @@ impl<T: Config> Pallet<T> {
...
+			T::HdxAssetId::get(),
+			&hdx_state_before,
+			&hdx_state_after,
+			&hdx_delta_changes,
+			false,
+		);
+
+		T::OmnipoolHooks::on_hub_asset_trade(origin, hdx_info)?;
+
+		// Store HDX fee to keep it in sync with the oracle update
```
enthusiastmartin:
```suggestion
		// Update HDX dynamic fee to keep it in sync with the oracle update
		// Fees must be updated on any liquidity change
```

pallets/omnipool/README.md
```diff
@@ -24,15 +24,24 @@ and in return they will receive the token of their choice in the appropriate qua
 Omnipool is implemented with concrete Balance type: u128.
 
 #### Imbalance mechanism
+
```
enthusiastmartin: i guess this readme is really out of date anyway - eg. the imbalance is no longer a thing. it has been removed.

so, i would no bother changing it now unless completely redone.

integration-tests/src/oracle.rs
```diff
@@ -602,3 +602,152 @@ fn bifrost_oracle_update_fail_should_charge_fee() {
...
+			RuntimeOrigin::signed(ALICE.into()),
+			LRNA, // selling hub asset
+			DAI,
+			10 * UNITS,
+			0,
+		));
+
+		// Finalize block to trigger oracle update
+		hydradx_runtime::System::reset_events();
+		hydradx_run_to_next_block();
```
martinfridrich: This doesn't look correct: reset events, move to next block and assert events from previous block

`reset_events()` should be called at the end of the block, prior initialize on new block but this basically calls it on current block before on_finalize.
This way events from on_finalize are in next block instead of current.

integration-tests/src/oracle.rs
```diff
@@ -602,3 +602,152 @@ fn bifrost_oracle_update_fail_should_charge_fee() {
...
+			.collect();
+
+		// Should have oracle updates for:
+		// 1. DAI/LRNA pair (from the trade)
+		// 2. HDX/LRNA pair (from H2O routing to HDX subpool)
+		let has_dai_lrna_update = oracle_updated_events.iter().any(|e| {
+			matches!(
+				e,
+				hydradx_runtime::RuntimeEvent::EmaOracle(pallet_ema_oracle::Event::OracleUpdated {
+					assets: (LRNA, DAI) | (DAI, LRNA),
```
martinfridrich: Why "or"? You are doing single trade and you know assets' order.
I'm not saying to assert on value that can change but assets' order will be always the same until we change the test.

integration-tests/src/oracle.rs
```diff
@@ -602,3 +602,152 @@ fn bifrost_oracle_update_fail_should_charge_fee() {
...
+		));
+
+		hydradx_runtime::System::reset_events();
+		hydradx_run_to_next_block();
+
+		// Act
+		assert_ok!(hydradx_runtime::Omnipool::buy(
+			RuntimeOrigin::signed(ALICE.into()),
+			DAI,  // buying DAI
+			LRNA, // paying with hub asset
```
martinfridrich:
```suggestion
			DAI,
			LRNA,
```
Please remove all useless comments,I won't comment on all of them

pallets/omnipool/src/lib.rs
```diff
@@ -937,8 +944,11 @@ pub mod pallet {
 
 			T::OmnipoolHooks::on_trade(origin.clone(), info_in, info_out)?;
 
-			let protocol_fees =
-				Self::process_protocol_fee(state_changes.fee.protocol_fee, state_changes.fee.burned_protocol_fee)?;
+			let protocol_fees = Self::process_protocol_fee(
+				origin.clone(),
```
martinfridrich:
```suggestion
				origin,
```

pallets/omnipool/README.md
dmoka: reverted back, but i left only the part "#### Hub Asset Trading". I dont think it can harm, it is still some valid info for us.

pallets/omnipool/src/lib.rs
```diff
@@ -1164,8 +1174,11 @@ pub mod pallet {
...
```
martinfridrich:
```suggestion
				origin,
```

pallets/omnipool/src/lib.rs
```diff
@@ -1608,19 +1625,44 @@ impl<T: Config> Pallet<T> {
...
 			let asset_state = maybe_asset.as_mut().ok_or(Error::<T>::AssetNotFound)?;
 			asset_state.hub_reserve = asset_state
 				.hub_reserve
 				.checked_add(amount)
 				.ok_or(ArithmeticError::Overflow)?;
 			Ok(())
-		})
+		})?;
+
+		let hdx_state_after = Self::load_asset_state(T::HdxAssetId::get())?;
```
martinfridrich, line 1646: This can be calculated from `hdx_state_before`, only thing that changes is `hub_reserve`

pallets/omnipool/src/tests/invariants.rs
```diff
@@ -49,6 +49,69 @@ macro_rules! assert_asset_invariant_not_decreased {
...
+			"Invariant decreased for {} - {:?} >= {:?}",
+			$desc,
+			new_s,
+			old_s
+		);
+	}};
+}
+
+/*
+fn assert_invariants_after_trade(
```
martinfridrich: Why do we need this commented test?

martinfridrich (commented):
Do we have test where we trade H2O for HDX?

pallets/omnipool/src/tests/invariants.rs
enthusiastmartin: this was removed in my PR, but it seems this one brings it back.. can be removed

pallets/omnipool/README.md
```diff
@@ -39,6 +39,12 @@ Omnipool pallet supports multiple hooks which are triggerred on certain operatio
 
 This is currently used to update on-chain oracle and in the circuit breaker.
 
+#### Hub Asset Trading
+
+When Hub Asset (LRNA) is sold to the Omnipool, the incoming LRNA is routed to the HDX subpool instead of the traded
```
mrq1911: LRNA -> H2O

pallets/omnipool/src/lib.rs
```diff
@@ -1728,14 +1777,22 @@ impl<T: Config> Pallet<T> {
...
 		);
 
+		// Emit trade event for H2O routing to HDX subpool
+		pallet_broadcast::Pallet::<T>::deposit_trade_event(
+			who.clone(),
+			Self::protocol_account(),
+			pallet_broadcast::types::Filler::Omnipool,
+			pallet_broadcast::types::TradeOperation::ExactIn,
+			vec![Asset::new(T::HubAssetId::get().into(), hub_reserve_delta)],
+			vec![Asset::new(T::HdxAssetId::get().into(), Balance::zero())],
```
mrq1911: this is not a swap, use transfer event instead

mrq1911: swap happened before buying that H2O

pallets/omnipool/src/lib.rs
```diff
@@ -1845,14 +1909,22 @@ impl<T: Config> Pallet<T> {
...
-			vec![Asset::new(
-				T::HubAssetId::get().into(),
-				*state_changes.asset.delta_hub_reserve,
-			)],
+			vec![Asset::new(T::HubAssetId::get().into(), hub_reserve_delta)],
 			vec![Asset::new(asset_out.into(), *state_changes.asset.delta_reserve)],
 			trade_fees,
 		);
 
+		// Emit trade event for H2O routing to HDX subpool
```
mrq1911: transfer event instead

mrq1911 (changes requested):

integration-tests/src/oracle.rs
dmoka: yeah right, the reset was not needed, removed

pallets/omnipool/src/lib.rs
dmoka: that could also work, but i feel this current approach is safer on the long term. it is just one storage read. But if you insist, I can change it ofc...

dmoka:
> Do we have test where we trade H2O for HDX?

good catch, added both integration and unit tests. 

We dont do any special handling here, it is a sort of "extra" work as we route from HDX to HDX pool, but the result would be the same we would skip rerouting for such trades. We choose this for code simplicity.

integration-tests/src/utility.rs
```diff
@@ -5,10 +5,10 @@ use frame_support::assert_ok;
 use pallet_broadcast::types::Destination;
 
 use crate::assert_balance;
-use hydradx_runtime::LBP;
 use hydradx_runtime::XYK;
 use hydradx_runtime::{Currencies, Omnipool, Runtime};
 use hydradx_runtime::{RuntimeCall, Utility};
+use hydradx_runtime::{RuntimeEvent, LBP};
```
dmoka:
```suggestion
use hydradx_runtime::{LBP};
```

integration-tests/src/liquidation.rs
```diff
@@ -195,6 +196,8 @@ fn liquidation_should_work() {
 	// Snapshot contains the storage of EVM, AssetRegistry, Timestamp, Omnipool and Tokens pallets
 	hydra_live_ext(PATH_TO_SNAPSHOT).execute_with(|| {
 		// Arrange
+		deposit_hdx_to_protocol_account();
```
enthusiastmartin, line 199: why this is here in this PR ?

pallets/omnipool/src/lib.rs
```diff
@@ -1675,7 +1721,11 @@ impl<T: Config> Pallet<T> {
 		);
 
 		let (taken_fee, trade_fees) = Self::process_trade_fee(who, asset_out, state_changes.fee.asset_fee)?;
-		let state_changes = state_changes.account_for_fee_taken(taken_fee);
+		let mut state_changes = state_changes.account_for_fee_taken(taken_fee);
+
+		// Store original hub reserve delta for routing to HDX subpool, then zero it
+		let hub_reserve_delta = *state_changes.asset.delta_hub_reserve;
+		state_changes.asset.delta_hub_reserve = BalanceUpdate::Increase(Balance::zero());
```
enthusiastmartin, line 1730: one additonal check i would add here

because here - we basically loose the information that the hub reserve actually Increases. 

i would ensure that the delta_hub_reserve is actually Increase variant. and fail if not.

pallets/omnipool/src/lib.rs
```diff
@@ -1792,7 +1848,11 @@ impl<T: Config> Pallet<T> {
...
```
enthusiastmartin, line 1861: same here - ensure that it is increase.

integration-tests/src/circuit_breaker.rs
```diff
@@ -121,7 +121,7 @@ fn sell_lrna_in_omnipool_should_fail_when_min_trade_limit_per_block_exceeded() {
 		let sell_amount = 300000 * UNITS;
 
 		//We need to split in multiple sells to avoid max in ratio error
-		for _ in 1..=3 {
+		for _ in 1..=2 {
```
enthusiastmartin, line 124: why this change here?

integration-tests/src/liquidation.rs
dmoka: so all in these tests the `process_protocol_fee` hence `increase_hdx_subpool_hub_reserve` called under the hood which updates hdx oracle.

the problem is that in this liquidation snapshot, the System pallet is not included, so it leads to 0 liquidty for HDX, which leads to OnTradeValueZero error from the oracle pallet when the oracle is updated.

Was not able to recreate the exact snapshot including the System pallet, so this was the easiest patch for that.

pallets/omnipool/src/lib.rs
dmoka: do you mean this ?

```
		ensure!(
			matches!(state_changes.asset.delta_hub_reserve, BalanceUpdate::Increase(_)),
			Error::<T>::HubAssetUpdateError
		);
```

integration-tests/src/circuit_breaker.rs
dmoka: Because the PR routes h2o to treasury and each subsequent h2o sell gets more HDX out. so the cumulative hdx outflow hits the 50% circuit breaker limit earlier, so in the 3rd sell, and not in the 4rd sell

pallets/omnipool/src/lib.rs
dmoka: anyway, added, pls check the code if that is what you meant

pallets/omnipool/src/lib.rs
```diff
@@ -45,6 +45,13 @@
 //!
 //! This is currently used to update on-chain oracle and in the circuit breaker.
 //!
+//! ### Hub Asset Trading
+//!
+//! When Hub Asset (LRNA) is sold to the Omnipool, the incoming LRNA is routed to the HDX subpool
+//! instead of the traded asset's subpool. This increases the HDX subpool's hub reserve while
+//! leaving the traded asset's hub reserve unchanged. The routing strengthens the HDX price by
+//! increasing its backing with hub asset.
```
enthusiastmartin: no longer true

pallets/omnipool/README.md
```diff
@@ -39,6 +39,12 @@ Omnipool pallet supports multiple hooks which are triggerred on certain operatio
 
 This is currently used to update on-chain oracle and in the circuit breaker.
 
+#### Hub Asset Trading
+
+When Hub Asset (H2O) is sold to the Omnipool, the incoming H2O is routed to the HDX subpool instead of the traded
```
enthusiastmartin: needs update

pallets/omnipool/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-omnipool"
-version = "5.4.1"
+version = "5.5.0"
```
enthusiastmartin: 6.0.0

enthusiastmartin (changes requested):

Issue #1346, opened 2026-01-30
Bug:  Clean storage of HyperBridge


dmoka:
we have a lot of storage wastes coming from Ismp pallets.

We need to clean them up.

we need 2 task:
- clean them up
- open issue (or PR?) on hyperbridge repo to make sure it resolved asap

<img width="775" height="559" alt="Image" src="https://github.com/user-attachments/assets/470352d0-191a-422b-ac91-010a5a1d460f" />

dmoka:
helper script to analyze state of our storage [analyze-storage.js](https://github.com/user-attachments/files/25005279/analyze-storage.js), leading to summary as:

<img width="775" height="559" alt="Image" src="https://github.com/user-attachments/assets/3c2d5d2f-cec6-4311-888d-1326169dde50" />

PR #1357, opened 2026-02-06
fix: disable hyperbridge

F3Joule:
## Description
Due to bloating caused by ismp, this PR temporarily disables hyperbridge, so new state commitments doesn't appear in storage.

## Related Issue
#1346


.github/workflows/tests.yml
```diff
@@ -40,8 +40,6 @@ jobs:
         uses: jwlawson/actions-setup-cmake@v1.12
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          rustflags: '-D warnings'
```
enthusiastmartin, line 44: why we are removing this as part of thsi Pr?

mrq1911: i have moved it to the makefile

PR #1358, opened 2026-02-09
fix: dust account

dmoka:
For unbound evm accounts, the substrate account ID and the EVM-derived account ID differ, so we need to whitelsit to right account id.


pallets/duster/src/lib.rs
```diff
@@ -159,12 +159,16 @@ pub mod pallet {
 	}
 
 	#[pallet::call]
-	impl<T: Config> Pallet<T> {
+	impl<T: Config> Pallet<T>
+	where
+		T::AccountId: AsRef<[u8; 32]>,
```
dmoka: we dont need this anymore, can be removed

PR #1360, opened 2026-02-10
feat: ICE

enthusiastmartin:
This PR introduces foundation of ICE - Intent Composing Engine.


pallets/ice/amm-simulator/src/aave.rs
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	fn simulate_buy(
+		asset_in: AssetId,
+		asset_out: AssetId,
+		amount_out: Balance,
+		_max_amount_in: Balance,
+		snapshot: &Self::Snapshot,
+	) -> Result<(Self::Snapshot, TradeResult), SimulatorError> {
+		if snapshot.reserves.get(&asset_in).is_none() && snapshot.reserves.get(&asset_out).is_none() {
```
dmoka: what is the reason that we use `AND` here?

I think if one of them is missing, than we should return error, so using `OR`

pallets/ice/amm-simulator/src/aave.rs
```diff
@@ -0,0 +1,326 @@
...
+			return Err(SimulatorError::AssetNotFound);
+		}
+
+		Ok((
+			snapshot.clone(),
+			TradeResult {
+				amount_in: amount_out,
```
dmoka, line 418: we do 1:1, which is probably fine, but we also have the famous atoken rounding issue in aave, so in the router implementation we add a buffer on top of it, so it is not exactly 1:1


https://github.com/galacticcouncil/hydration-node/blob/cfb7a0a02b3d0c0d040366d6156cce5bc6f319a7/runtime/hydradx/src/evm/aave_trade_executor.rs#L456-L468

Wondering if we should do something similar here, or we should be fine?!

pallets/intent/src/lib.rs
```diff
@@ -0,0 +1,600 @@
...
+	// Function updates intent's `SwapData` and returns `true` if intent was fully resolved.
+	fn resolve_swap_intent(intent: &mut SwapData, resolve: &SwapData) -> Result<bool, DispatchError> {
+		match intent.swap_type {
+			SwapType::ExactIn => {
+				intent.amount_in = intent
+					.amount_in
+					.checked_sub(resolve.amount_in)
+					.ok_or(Error::<T>::ArithmeticOverflow)?;
+
+				intent.amount_out = intent.amount_out.saturating_sub(resolve.amount_out);
```
dmoka: i would use checked math here as we do in other places within this method

ice/ice-solver/src/v1/solver.rs
```diff
@@ -0,0 +1,664 @@
...
+						} else {
+							swap.amount_in
+						}
+					}
+				};
+				*available.entry(swap.asset_in).or_default() += amount_in;
+			}
+
+			for trade in &executed_trades {
+				let asset_in = trade.route.first().map(|t| t.asset_in).unwrap_or(0);
```
dmoka: any specific reason that we unwrap this to 0?

Also for asset out in the next line

ice/ice-solver/src/v1/solver.rs
```diff
@@ -0,0 +1,664 @@
...
+				let remaining = remaining_avail.get(&asset_out).copied().unwrap_or(0);
+				let total_demand = exactin_demand.get(&asset_out).copied().unwrap_or(0);
+
+				// Scale down proportionally if total ExactIn demand exceeds remaining availability
+				let actual_out = if total_demand > remaining && total_demand > 0 {
+					U256::from(ideal_amount)
+						.checked_mul(U256::from(remaining))
+						.and_then(|n| n.checked_div(U256::from(total_demand)))
+						.map(|r| r.as_u128())
+						.unwrap_or(ideal_amount)
```
dmoka: is it a good idea to fallback to ideal_amount in case of any unwrap failure?

ice/ice-solver/src/v1/solver.rs
```diff
@@ -0,0 +1,664 @@
...
+								direction: SwapType::ExactIn,
+								amount_in: trade_execution.amount_in,
+								amount_out: trade_execution.amount_out,
+								route: trade_execution.route,
+							});
+
+							state = new_state;
+						}
+						Err(_) => {
+							continue;
```
dmoka: i see that we just continue on error, also at line 184 and 205

Would not be better to at least log some error info here?

pallets/ice/amm-simulator/src/stableswap.rs
```diff
@@ -0,0 +1,670 @@
...
+		asset_in: AssetId,
+		asset_out: AssetId,
+		snapshot: &Self::Snapshot,
+	) -> Result<Ratio, SimulatorError> {
+		let (pool_id, pool_snapshot) = find_pool(asset_in, asset_out, snapshot)?;
+
+		if asset_in == pool_id {
+			// Price = how much asset_out you get per 1 share
+			// Using a small simulation to determine spot price
+			let test_shares = pool_snapshot.share_issuance / 10000; // 0.01% of total shares
```
dmoka: 10000 is a magic number, i would make a constant for it, or at least a variable

dmoka: if share_issuance is smaller than 10000, this will be zero and we treat it `InsufficientLiquidity`, i assume it is indended?

ice/ice-solver/src/v1/solver.rs
```diff
@@ -0,0 +1,664 @@
...
+				let asset_out = trade.route.last().map(|t| t.asset_out).unwrap_or(0);
+
+				if let Some(bal) = available.get_mut(&asset_in) {
+					*bal = bal.saturating_sub(trade.amount_in);
+				}
+				*available.entry(asset_out).or_default() += trade.amount_out;
+			}
+
+			let mut ideal_resolutions: Vec<(usize, ResolvedIntent)> = Vec::new();
+			let mut total_output_per_asset: BTreeMap<AssetId, Balance> = BTreeMap::new();
```
dmoka: what is the purpose of this `total_output_per_asset`?

I see it is populated at line 285, but that's it, we dont use anywhere else...

pallets/ice/amm-simulator/src/stableswap.rs
```diff
@@ -0,0 +1,670 @@
...
+
+	fn snapshot() -> Self::Snapshot {
+		let mut pools = BTreeMap::new();
+
+		for (pool_id, pool) in DP::pools() {
+			// TODO: we skip incorrect pools - this was likely due to incorrect snapshots used in tests
+			// but verify!
+			if let Some(peg_info) = DP::pool_pegs(pool_id) {
+				if peg_info.current.len() != pool.assets.len() {
+					continue;
```
dmoka, line 92: if possible and useful, I would log here.

pallets/ice/amm-simulator/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+		let mut current_amount = amount_in;
+		let original_amount_in = amount_in;
+
+		for trade in route.iter() {
+			let (new_state, result) = C::Simulators::simulate_sell(
+				trade.pool,
+				trade.asset_in,
+				trade.asset_out,
+				current_amount,
+				0, // No limit check on intermediate hops
```
dmoka, line 95: any reason for no limit check on intermediate hops? What happens if the solver propose a route that would fail onchain in an intermediate hop? Or i might have misunderstood this part...

anyway, if we keep it 0, i would comment why we use no limit

pallets/lazy-executor/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn offchain_worker(block_number: BlockNumberFor<T>) {
+			log::debug!(target: LOG_TARGET, "run offchain worker on block: {:?}", block_number);
+
+			let mut next_id = Self::dispatch_next_id();
+			for i in 0..Self::max_txs_per_block() {
+				next_id = if let Some(n) = next_id.checked_add(i as u128) {
```
dmoka: what is the reason that we add the `i` to get next_id each time? By doing so we will have offsets in the ids.

I would expect just to use `checked_add(1)`

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+
+		for cp in clearing_prices {
+			ensure!(!cp.1.n.is_zero() && !cp.1.d.is_zero(), Error::<T>::InvalidPriceRatio);
+		}
+
+		Ok(())
+	}
+
+	/// Function validates if intent was resolved based on clearing price.
+	fn validate_price_consitency(
```
dmoka:
```suggestion
	fn validate_price_consistency(
```

pallets/intent/src/lib.rs
```diff
@@ -0,0 +1,600 @@
...
+
+			for (i, intent_id) in expired.iter().enumerate() {
+				if i >= 10 {
+					break;
+				}
+
+				let call = Call::cleanup_intent { id: *intent_id };
+				let tx = T::create_bare(call.into());
+				if let Err(e) = SubmitTransaction::<T, Call<T>>::submit_transaction(tx) {
+					log::error!(target: OCW_LOG_TARGET, "to sumbmit cleanup_intent call, err: {:?}", e);
```
dmoka:
```suggestion
					log::error!(target: OCW_LOG_TARGET, "to submit cleanup_intent call, err: {:?}", e);
```

pallets/lazy-executor/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+				} else {
+					log::debug!(target: LOG_TARGET, "queue is empty");
+					break;
+				};
+
+				if CallQueue::<T>::get(next_id).is_some() {
+					let call = Call::dispatch_top {};
+					let tx = T::create_bare(call.into());
+					let r = SubmitTransaction::<T, Call<T>>::submit_transaction(tx);
+					log::debug!(target: LOG_TARGET, "sutmitted dispatch_top transaction, result: {:?}", r,);
```
dmoka:
```suggestion
					log::debug!(target: LOG_TARGET, "submitted dispatch_top transaction, result: {:?}", r,);
```

pallets/ice/src/traits.rs
```diff
@@ -0,0 +1,3 @@
+// This file can be removed or kept for backwards compatibility.
```
dmoka, line 1: i believe this traits file can be removed

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+use sp_runtime::traits::Zero;
+use sp_std::borrow::ToOwned;
+use sp_std::collections::btree_map::BTreeMap;
+use sp_std::collections::btree_set::BTreeSet;
+use sp_std::vec::Vec;
+
+pub use pallet::*;
+pub use weights::WeightInfo;
+
+//TODO: make sure tx is always first in the block(same as liquidations), this is tmp
```
dmoka: todo

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+			// V1 solver may produce solutions with no trades (perfect CoW matching)
+			ensure!(!solution.resolved_intents.is_empty(), Error::<T>::InvalidSolution);
+
+			Self::validate_clearing_prices(&solution.clearing_prices)?;
+
+			let mut processed_intents: BTreeSet<IntentId> = BTreeSet::new();
+			let holding_pot = Self::get_pallet_account();
+			let holding_origin: OriginFor<T> = Origin::<T>::Signed(holding_pot.clone()).into();
+
+			// TODO: this is not most perform solution, verify it works and optimize
```
dmoka: todo

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+		ensure!(intent.amount_in() >= ed_in, Error::<T>::InvalidAmount);
+		ensure!(intent.amount_out() >= ed_out, Error::<T>::InvalidAmount);
+
+		Ok(())
+	}
+
+	/// Function validates provided solution and returns solution's score if solution is
+	/// valid.
+	fn validate_unsigned_solution(solution: &Solution) -> Result<(), DispatchError> {
+		//TODO:
```
dmoka, line 856: todo

pallets/ice/amm-simulator/src/aave.rs
```diff
@@ -0,0 +1,326 @@
...
+			SimulatorError::Other
+		})?;
+
+		// Ensure sufficient length
+		ensure!(decoded.len() == param_types.len(), {
+			log::error!(target: LOG_TARGET, "invalid reserve data");
+			SimulatorError::Other
+		});
+
+		let a_token = EvmAddress::from_slice(decoded[8].clone().into_address().unwrap_or_default().as_ref());
```
dmoka, line 300: if there is no address, should not we return simulator error, instead of `unwrap_or_default`?

runtime/hydradx/src/assets.rs
```diff
@@ -1871,6 +1876,63 @@ impl pallet_hsm::Config for Runtime {
...
+	type Currency = Currencies;
+	type MaxAllowedIntentDuration = MaxIntentDuration;
+	type TimestampProvider = Timestamp;
+	type HubAssetId = LRNA;
+	type WeightInfo = weights::pallet_intent::HydraWeight<Runtime>;
+}
+
+parameter_types! {
+	pub const IcePalletId: PalletId = PalletId(*b"ice_ice#");
+	pub const SimulatorHubAsset: AssetId = 0;
```
dmoka: is HDX the hub asset for ice?  i mean, for Omnipool we use LRNA with asse `1`, thats why it is strange.

pallets/lazy-executor/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+				}
+
+				ensure!(
+					CallQueue::<T>::contains_key(Self::dispatch_next_id()),
+					InvalidTransaction::Call
+				);
+
+				return ValidTransaction::with_tag_prefix(OCW_TAG_PREFIX)
+					.priority(T::UnsignedPriority::get())
+					.and_provides(OCW_PROVIDES.to_vec())
```
dmoka: since OCW_PROVIDES is fixed, would this mean that only one dispatch_top can be in the the tx pool? Is it acceptable, or we might wanna support more?

pallets/lazy-executor/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+			let mut next_id = Self::dispatch_next_id();
+			for i in 0..Self::max_txs_per_block() {
+				next_id = if let Some(n) = next_id.checked_add(i as u128) {
+					n
+				} else {
+					log::debug!(target: LOG_TARGET, "queue is empty");
+					break;
+				};
+
+				if CallQueue::<T>::get(next_id).is_some() {
```
dmoka: maybe contains_key is faster?!

pallets/lazy-executor/src/tests/mod.rs
```diff
@@ -0,0 +1,11 @@
+use mock::System;
+
+mod add_to_queue;
+pub(crate) mod mock;
+#[allow(clippy::module_inception)]
+// mod tests;
```
dmoka:
```suggestion
mod tests;
```

pallets/ice/amm-simulator/src/stableswap.rs
```diff
@@ -0,0 +1,670 @@
...
+	);
+
+	let pool_id = find_pool_id_for_snapshot(pool_snapshot, snapshot)?;
+	let updated_snapshot = snapshot.clone().with_updated_pool(pool_id, updated_pool);
+
+	Ok((updated_snapshot, TradeResult::new(amount_in, amount_out)))
+}
+
+fn simulate_regular_buy(
+	_pool_id: AssetId,
```
dmoka: if this param not used, we could just remove it

runtime/hydradx/src/ice_simulator_provider.rs
```diff
@@ -0,0 +1,127 @@
...
+use primitives::EvmAddress;
+use sp_core::U256;
+
+pub struct Aave<T>(PhantomData<T>);
+
+impl<T> AaveDataProvider for Aave<T>
+where
+	T: frame_system::Config + pallet_liquidation::Config + pallet_evm::Config + pallet_dispatcher::Config,
+	BalanceOf<T>: TryFrom<U256> + Into<U256>,
+	T::AddressMapping: AddressMapping<T::AccountId>,
```
dmoka, line 108:
```suggestion
```

ice/ice-solver/src/v1/solver.rs
```diff
@@ -0,0 +1,664 @@
...
+		Ok(Solution {
+			resolved_intents: ResolvedIntents::truncate_from(resolved_intents),
+			trades: SolutionTrades::truncate_from(executed_trades),
+			clearing_prices,
+			score: total_score,
+		})
+	}
+
+	fn collect_unique_assets(intents: &[Intent]) -> Vec<AssetId> {
+		let mut assets: Vec<AssetId> = Vec::new();
```
dmoka: since we want unique assets, we could possibly make this function simpler and faster if we would use BTreeSet.

integration-tests/src/solver.rs
```diff
@@ -0,0 +1,1805 @@
...
+			assert_eq!(
+				bnc_received, swap_data.amount_out,
+				"BNC received should equal resolved amount_out"
+			);
+		});
+}
+
+/// Test partial CoW match: Alice sells large HDX, Bob sells small BNC (opposite directions)
+#[test]
+fn solver_v1_two_intents_partial_cow_match() {
```
dmoka: i could not find one, maybe we could add a test for full cow match too

pallets/ice/src/tests/mod.rs
```diff
@@ -0,0 +1,8 @@
+use crate::*;
+use ice_support::AssetId;
+use pretty_assertions::assert_eq;
+
```
dmoka: generally i miss some property based tests for ICE related invariants. Not sure what would be the best properties, maybe  @jepidoptera knows more,  but would be useful to add such to catch issues early

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,515 @@
...
+				});
+
+				let intent = pallet_intent::Pallet::<T>::get_intent(id).ok_or(Error::<T>::IntentNotFound)?;
+				let surplus = intent.data.surplus(resolve).ok_or(Error::<T>::ArithmeticOverflow)?;
+				exec_score = exec_score.checked_add(surplus).ok_or(Error::<T>::ArithmeticOverflow)?;
+
+				pallet_intent::Pallet::<T>::intent_resolved(&owner, resolved_intent)?;
+			}
+
+			ensure!(solution.score == exec_score, Error::<T>::ScoreMismatch);
```
dmoka, line 429: maybe we could add a debug assert here, this seems to be important:

```
		debug_assert_eq!(
			solution.score, exec_score,
			"ICE: score mismatch - solution says {}, execution computed {}",
			solution.score, exec_score
		);
```

pallets/intent/src/lib.rs
dmoka: since it is an error scenario we could consider adding a debug_assert so fuzzer might catch issues

pallets/lazy-executor/src/lib.rs
dmoka: since it is an error scenario we could consider adding a debug_assert so fuzzer might catch issues

ice/ice-solver/src/v1/solver.rs
dmoka: if its an unexpected error, we could anyway add a `debug_assert!()`

pallets/ice/amm-simulator/src/stableswap.rs
dmoka: if its an unexpected error, we could add a debug assert

pallets/ice/amm-simulator/src/aave.rs
martinfridrich: As note on line 120 says, this is not full implementation of avve simulator.
I don't think it's necessary in this version

pallets/intent/src/lib.rs
martinfridrich: Intent can be resolved better than limit and in that case `resolve.amount_out` would > `intent.amount_out`  checked math would overflow even when intent would be resolved correctly

pallets/ice/src/lib.rs
martinfridrich: yes, we decided to not do it in this version

pallets/ice/src/lib.rs
martinfridrich: yes, we decided to not do it in this version

pallets/ice/amm-simulator/src/aave.rs
martinfridrich: It's same as `AaveTradeExecutor` https://github.com/galacticcouncil/hydration-node/blob/139ded5172c01bbdec1685e5e8ef94c843d257e6/runtime/hydradx/src/evm/aave_trade_executor.rs#L245

pallets/ice/src/tests/mod.rs
martinfridrich: What kind of invariats do you mean?
ICE doesn't do any pool state change nor anything similar. 
Basically if solution is executed and intents are resolved in users' limits it's valid. 
We can add that it can't trade more than necessary so we won't have leftovers on ice's pot account but AFAIK we don't want to do it in this version, solver is this WIP

pallets/ice/src/lib.rs
martinfridrich: I don't think it's good idea. 
`solution.score` is basically user input. It would be easy for fuzzer to craft invalid solution and we would end up with false positive

pallets/ice/amm-simulator/src/aave.rs
martinfridrich: As comment says this is tmp. impl. of simulator
Full validation is more complicated than using `||`

ice/ice-solver/src/v1/solver.rs
martinfridrich: solver is WIP

ice/ice-solver/src/v1/solver.rs
martinfridrich: solver is WIP

pallets/ice/amm-simulator/src/stableswap.rs
martinfridrich: yes

integration-tests/src/solver.rs
martinfridrich: https://github.com/galacticcouncil/hydration-node/blob/b022454d33240de3f18b8ce5680671ee12a36d25/integration-tests/src/solver.rs#L377

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,567 @@
...
+
+			// TODO: this is not most perform solution, verify it works and optimize
+
+			for ResolvedIntent { id, data: intent } in &solution.resolved_intents {
+				Self::validate_intent_amounts(intent)?;
+
+				let owner = pallet_intent::Pallet::<T>::intent_owner(id).ok_or(Error::<T>::IntentOwnerNotFound)?;
+				pallet_intent::Pallet::<T>::unlock_funds(&owner, intent.asset_in(), intent.amount_in())?;
+
+				log::debug!(target: LOG_TARGET, "{:?}: sumbit_solution(), unlock and transfer amounts, owner: {:?}, asset: {:?}, amount: {:?}",
```
dmoka, line 227:
```suggestion
				log::debug!(target: LOG_TARGET, "{:?}: submit_solution(), unlock and transfer amounts, owner: {:?}, asset: {:?}, amount: {:?}",
```

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,567 @@
...
+				// near-perfect intent matching can leave a negligible AMM remainder.
+				if t.amount_in < ed_in || t.amount_out < ed_out {
+					log::debug!(target: LOG_TARGET, "{:?}: submit_solution(), skipping dust trade: amount_in {:?} (ed {:?}), amount_out {:?} (ed {:?})",
+						LOG_PREFIX, t.amount_in, ed_in, t.amount_out, ed_out);
+					continue;
+				}
+
+				match t.direction {
+					SwapType::ExactOut => {
+						log::debug!(target: LOG_TARGET, "{:?}: sumbit_solution(), buying, asset_in: {:?}, asset_out: {:?}, amount_out: {:?}, max_amount_in: {:?}, route: {:?}",
```
dmoka, line 256:
```suggestion
						log::debug!(target: LOG_TARGET, "{:?}: submit_solution(), buying, asset_in: {:?}, asset_out: {:?}, amount_out: {:?}, max_amount_in: {:?}, route: {:?}",
```

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,567 @@
...
+							holding_origin.clone(),
+							asset_in,
+							asset_out,
+							t.amount_out.into(),
+							t.amount_in.into(),
+							t.route.clone(),
+						)?;
+					}
+					SwapType::ExactIn => {
+						log::debug!(target: LOG_TARGET, "{:?}: sumbit_solution(), selling, asset_in: {:?}, asset_out: {:?}, amount_in: {:?}, min_amount_out: {:?}, route: {:?}",
```
dmoka, line 269:
```suggestion
						log::debug!(target: LOG_TARGET, "{:?}: submit_solution(), selling, asset_in: {:?}, asset_out: {:?}, amount_in: {:?}, min_amount_out: {:?}, route: {:?}",
```

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,567 @@
...
+				log::debug!(target: LOG_TARGET, "{:?}: sumbit_solution(), resolving intent, id: {:?}", LOG_PREFIX, id);
+
+				ensure!(processed_intents.insert(*id), Error::<T>::DuplicateIntent);
+
+				let owner = pallet_intent::Pallet::<T>::intent_owner(id).ok_or(Error::<T>::IntentOwnerNotFound)?;
+
+				let fee_amount = Self::protocol_fee().mul_floor(resolve.amount_out());
+				let payout = resolve.amount_out().saturating_sub(fee_amount);
+
+				log::debug!(target: LOG_TARGET, "{:?}: sumbit_solution(), transferring, id: {:?}, to: {:?}, amount: {:?}, fee: {:?}", LOG_PREFIX, id, owner, payout, fee_amount);
```
dmoka, line 297:
```suggestion
				log::debug!(target: LOG_TARGET, "{:?}: submit_solution(), transferring, id: {:?}, to: {:?}, amount: {:?}, fee: {:?}", LOG_PREFIX, id, owner, payout, fee_amount);
```

pallets/ice/src/lib.rs
```diff
@@ -0,0 +1,567 @@
...
+
+				<T as Config>::Currency::transfer(resolve.asset_out(), &holding_pot, &owner, payout, AllowDeath)?;
+
+				Self::validate_price_consistency(&mut exec_prices, resolve)?;
+
+				let intent = pallet_intent::Pallet::<T>::get_intent(id).ok_or(Error::<T>::IntentNotFound)?;
+				let surplus = pallet_intent::Pallet::<T>::compute_surplus(&intent, resolve)
+					.ok_or(Error::<T>::ArithmeticOverflow)?;
+				log::debug!(target: LOG_TARGET, "{:?}: sumbit_solution(), id: {:?}, surplus: {:?}", LOG_PREFIX, id, surplus);
```
dmoka, line 306:
```suggestion
				log::debug!(target: LOG_TARGET, "{:?}: submit_solution(), id: {:?}, surplus: {:?}", LOG_PREFIX, id, surplus);
```

pallets/intent/src/lib.rs
```diff
@@ -0,0 +1,871 @@
...
+
+			for (i, intent_id) in expired.iter().enumerate() {
+				if i >= 10 {
+					break;
+				}
+
+				let call = Call::cleanup_intent { id: *intent_id };
+				let tx = T::create_bare(call.into());
+				if let Err(e) = SubmitTransaction::<T, Call<T>>::submit_transaction(tx) {
+					debug_assert!(false, "laxy-executorn: failed to submit dispatch_top transaction");
```
dmoka, line 351:
```suggestion
					debug_assert!(false, "laxy-executor: failed to submit dispatch_top transaction");
```

runtime/hydradx/src/assets.rs
```diff
@@ -1838,6 +1844,91 @@ impl pallet_hsm::Config for Runtime {
...
+impl pallet_lazy_executor::Config for Runtime {
+	type RuntimeCall = RuntimeCall;
+	type UnsignedLongevity = ConstU64<2>;
+	type UnsignedPriority = ConstU64<100>;
+	type WeightInfo = weights::pallet_lazy_executor::HydraWeight<Runtime>;
+}
+
+parameter_types! {
+	//24 hours
+	pub const MaxIntentDuration: u64  = 24 * 3_600 * 1_000;
```
mrq1911, line 1856: there is still an limit?

runtime/hydradx/src/assets.rs
```diff
@@ -1838,6 +1844,91 @@ impl pallet_hsm::Config for Runtime {
...
+	type HubAssetId = LRNA;
+	type OraclePriceProvider = OraclePriceProvider<AssetId, EmaOracle, LRNA>;
+	type BlockNumberProvider = System;
+	type MinDcaPeriod = MinimalPeriod;
+	type MaxIntentsPerAccount = sp_core::ConstU32<100>;
+	type WeightInfo = weights::pallet_intent::HydraWeight<Runtime>;
+}
+
+parameter_types! {
+	pub const IcePalletId: PalletId = PalletId(*b"ice_ice#");
```
mrq1911, line 1874:
```suggestion
	pub const IcePalletId: PalletId = PalletId(*b"iceicebb");
```

pallets/ice/src/tests/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod mock;
+mod ocw;
```
mrq1911, line 2: cow?

pallets/intent/src/lib.rs
```diff
@@ -0,0 +1,871 @@
...
+	pub(super) type IntentOwner<T: Config> = StorageMap<_, Blake2_128Concat, IntentId, T::AccountId>;
+
+	#[pallet::storage]
+	/// Intent id sequencer
+	#[pallet::getter(fn next_incremental_id)]
+	pub(super) type NextIncrementalId<T: Config> = StorageValue<_, IncrementalIntentId, ValueQuery>;
+
+	#[pallet::storage]
+	/// Reverse index mapping account to its intent ids for easy lookup.
+	pub type AccountIntents<T: Config> =
```
mrq1911, line 236: do we really need 3 maps? cannot we just store intents here where it preserve ownership as well? this way you are spending 3x reads and writes per intent bloat storage and take longer time to load into solver

pallets/intent/src/lib.rs
```diff
@@ -0,0 +1,871 @@
...
+		MaxIntentsReached,
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn get_intent)]
+	pub type Intents<T: Config> = StorageMap<_, Blake2_128Concat, IntentId, Intent>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn intent_owner)]
+	pub(super) type IntentOwner<T: Config> = StorageMap<_, Blake2_128Concat, IntentId, T::AccountId>;
```
mrq1911, line 227: if intent cannot change ower there is no reason to have separate map

runtime/hydradx/src/assets.rs
martinfridrich: yes, you can create intent without deadline or withing this limit

pallets/ice/src/tests/mod.rs
martinfridrich: these are offchain-worker(ocw) related tests

pallets/intent/src/lib.rs
```diff
@@ -0,0 +1,871 @@
...
+		/// Intent was resolved as part of ICE solution execution.
+		IntentResolved {
+			id: IntentId,
+			amount_in: Balance,
+			amount_out: Balance,
+			fee: Balance,
+		},
+
+		/// Portion of intent was resolved as part of ICE solution execution.
+		IntentResovedPartially {
```
dmoka, line 189:
```suggestion
		IntentResolvedPartially {
```

PR #1361, opened 2026-02-10
fix: dispenser benchmarking

iamyxsh:
## Checklist:

### DIFF
--- a/pallets/dispenser/Cargo.toml
+++ b/pallets/dispenser/Cargo.toml
@@ -57,6 +57,7 @@ std = [
     "serde_json/std",
     "pallet-currencies/std",
     "pallet-signet/std",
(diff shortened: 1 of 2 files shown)


pallets/dispenser/Cargo.toml
```diff
@@ -57,6 +57,7 @@ std = [
     "serde_json/std",
     "pallet-currencies/std",
     "pallet-signet/std",
+    "pallet-asset-registry/std",
```
enthusiastmartin, line 60: there should not be a need to be dependent on asset registry.

enthusiastmartin:
regenerate weights

enthusiastmartin:
in #1380

Issue #1352, opened 2026-02-02
Configurable oracle sources, not just for bifrost


mrq1911:
Dynamic version of bifrost oracle interface, allowing to add more sources using extrinsic. Each source would have list of origins/accounts that are able to set it.

This would allow us to use our EMA oracle logic for external oracle sources.

Also set oracle extrinsic should reserve space for calculation of its oracle on end of the block to make sure its always recalculated when updated - set oracle extrinsic shouldn't ever fail.

dmoka:
# Configurable Oracle Sources

## Overview

This feature introduces a dynamic, configurable system for external oracle sources, replacing the current hardcoded
bifrost oracle implementation. It allows governance to register multiple external oracle sources, each with its own
authorized accounts that can submit price updates.

## Motivation

Currently, the EMA oracle pallet has a hardcoded `BIFROST_SOURCE` with a dedicated `BifrostOrigin` config type. This
approach doesn't scale when multiple external oracle providers need to be integrated. By making oracle sources
configurable, we can:

- Add new external oracle sources without runtime upgrades
- Manage authorized accounts per source via governance
- Reuse the existing EMA oracle logic for all external sources

## Design Decisions

| Decision              | Choice                                      | Rationale                                                                     |
|-----------------------|---------------------------------------------|-------------------------------------------------------------------------------|
| Authorization model   | Account-based per source (StorageDoubleMap)  | O(1) lookup, no limit on accounts per source                                  |
| Price deviation check | None                                         | No price deviation check for external sources; authorized accounts are trusted, and clamped oracle will be used to prevent attacks |
| Whitelist behavior    | Bypass                                       | External sources can update any asset pair (same as current bifrost)          |
| Source management     | `AuthorityOrigin`                            | Consistent with existing whitelist management                                 |
| Migration             | Deprecate bifrost                            | Keep `update_bifrost_oracle` as deprecated, delegate to shared internal logic |
| Accumulator           | Single unbounded `Accumulator`               | No separate `ExternalAccumulator`; soft limit only for AMM trades             |
| Weight reservation    | `fractional_on_finalize_weight` per call     | Each external oracle call pays for its own `on_finalize` processing           |

## Accumulator Changes

### Make Accumulator unbounded

The existing `Accumulator` changes from `BoundedBTreeMap` to a regular `BTreeMap`:

```rust
/// Accumulator for oracle data in current block that will be recorded at the end of the block.
#[pallet::storage]
[...]

PR #1369, opened 2026-02-19
feat: configurable oracle sources

dmoka:
Spec can be found at: https://github.com/galacticcouncil/hydration-node/issues/1352#issuecomment-3848452471

todo:
- [ ] rebench

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10507,7 +10507,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-evm-accounts"
-version = "1.6.0"
+version = "1.6.1"
 dependencies = [
  "frame-benchmarking",
  "frame-support",
--- a/integration-tests/src/evm.rs
+++ b/integration-tests/src/evm.rs
@@ -355,7 +355,7 @@ mod account_conversion {
 	}
 
 	#[test]
-	fn evm_call_from_runtime_rpc_should_not_be_accepted_from_bound_addresses() {
+	fn evm_rpc_call_should_be_accepted_from_bound_addresses() {
 		TestNet::reset();
 
 		Hydra::execute_with(|| {
@@ -368,41 +368,21 @@ mod account_conversion {
 
 			let evm_address = EVMAccounts::evm_address(&Into::<AccountId>::into(ALICE));
 
-			//Act & Assert
-			assert_noop!(
-				hydradx_runtime::Runtime::call(
-					evm_address,   // from
-					DISPATCH_ADDR, // to
-					data,          // data
-					U256::from(1000u64),
-					U256::from(100000u64),
-					None,
-					None,
-					None,
-					false,
-					None,
-					None,
-				),
-				pallet_evm_accounts::Error::<Runtime>::BoundAddressCannotBeUsed
-			);
-		});
-	}
+			//Act & Assert - both estimate=false and estimate=true should work from RPC
+			assert_ok!(hydradx_runtime::Runtime::call(
+				evm_address,   // from
+				DISPATCH_ADDR, // to
+				data.clone(),  // data
+				U256::from(1000u64),
+				U256::from(100000u64),
+				None,
+				None,
+				None,
+				false,
+				None,
+				None,
+			));
 
-	#[test]
-	fn estimation_of_evm_call_should_be_accepted_even_from_bound_address() {
-		TestNet::reset();
-
-		Hydra::execute_with(|| {
-			//Arrange
-			let data =
-				hex!["4d0045544800d1820d45118d78d091e685490c674d7596e62d1f0000000000000000140000000f0000c16ff28623"]
-					.to_vec();
-
-			assert_ok!(EVMAccounts::bind_evm_address(RuntimeOrigin::signed(ALICE.into())),);
-
-			let evm_address = EVMAccounts::evm_address(&Into::<AccountId>::into(ALICE));
-
-			//Act & Assert
 			assert_ok!(hydradx_runtime::Runtime::call(
 				evm_address,   // from
 				DISPATCH_ADDR, // to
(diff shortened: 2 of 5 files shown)


runtime/hydradx/src/assets.rs
```diff
@@ -666,17 +665,9 @@ where
 pub fn bifrost_account() -> AccountId {
 	hex!["7369626cee070000000000000000000000000000000000000000000000000000"].into()
 }
```
martinfridrich:
```suggestion
```

pallets/ema-oracle/src/lib.rs
```diff
@@ -319,62 +337,133 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// DEPRECATED - Use `set_external_oracle` instead.
+		/// This is kept for backward compatibility with bifrost and will be removed in the future.
 		#[pallet::call_index(2)]
-		#[pallet::weight(<T as Config>::WeightInfo::update_bifrost_oracle())]
+		#[pallet::weight(<T as Config>::WeightInfo::update_bifrost_oracle()
+			.saturating_add(fractional_on_finalize_weight::<T>(MAX_EXTERNAL_ENTRIES_PER_BLOCK)))]
 		pub fn update_bifrost_oracle(
```
martinfridrich:
```suggestion

		#[pallet::call_index(2)]
		#[pallet::weight(<T as Config>::WeightInfo::update_bifrost_oracle()
			.saturating_add(fractional_on_finalize_weight::<T>(MAX_EXTERNAL_ENTRIES_PER_BLOCK)))]
		#[deprecated(note = "...")]
pub fn update_bifrost_oracle(
```

pallets/ema-oracle/src/lib.rs
```diff
@@ -319,62 +337,133 @@ pub mod pallet {
...
+				Error::<T>::SourceAlreadyRegistered
+			);
+			ExternalSources::<T>::insert(source, ());
+			Self::deposit_event(Event::ExternalSourceRegistered { source });
+			Ok(())
+		}
 
-			Ok(Pays::No.into())
+		#[pallet::call_index(5)]
+		#[pallet::weight(<T as Config>::WeightInfo::remove_external_source())]
```
martinfridrich: This doesn't account for cleaning multiple `AuthorizedAccounts`, `clear_prefix` is not 1 op.

pallets/ema-oracle/src/lib.rs
```diff
@@ -319,62 +337,133 @@ pub mod pallet {
...
+		source: Source,
+		asset_a: Box<polkadot_xcm::VersionedLocation>,
+		asset_b: Box<polkadot_xcm::VersionedLocation>,
+		price: (Balance, Balance),
+	) -> DispatchResultWithPostInfo {
+		ensure!(ExternalSources::<T>::contains_key(source), Error::<T>::SourceNotFound);
+		ensure!(
+			AuthorizedAccounts::<T>::contains_key(source, &who),
+			Error::<T>::NotAuthorized
+		);
```
martinfridrich: you don't really need both, `AuthorizedAccounts` can't exists without `ExternalSources`

pallets/ema-oracle/src/lib.rs
```diff
@@ -384,10 +473,13 @@ impl<T: Config> Pallet<T> {
 				entry.accumulate_volume_and_update_from(&oracle_entry);
 				Ok(())
 			} else {
-				accumulator
-					.try_insert((src, assets), oracle_entry)
-					.map(|_| ())
-					.map_err(|_| ())
+				if !ExternalSources::<T>::contains_key(src) && accumulator.len() >= T::MaxUniqueEntries::get() as usize
```
martinfridrich: I would use var for `ExternalSources::<T>::contains_key(src)`, no need to hit it 2 times, it's already accessed on 466

martinfridrich: What "soft limit up to MaxUniqueEntries" mean? 
This really depend on txs order. 
e.g. MaxUniqueEntries=5
we have 5 external calls and 1 from AMM(in this order) and AMM will fail because accumulator will be full

pallets/ema-oracle/src/tests/external_oracle.rs
```diff
@@ -0,0 +1,524 @@
...
+			(100, 99),
+		));
+
+		//Assert - accumulator has more entries than MaxUniqueEntries
+		assert_eq!(Accumulator::<Test>::get().len(), (max_entries + 1) as usize);
+	});
+}
+
+#[test]
+fn soft_limit_only_for_non_external_sources() {
```
martinfridrich: This is basically same test as above

pallets/ema-oracle/src/lib.rs
dmoka: i tried that first but `#[deprecated]` would require `#[allow(deprecated)]` at every call site. Does it worth it?

pallets/ema-oracle/src/lib.rs
dmoka: you are right, but i think it is better double check. A possibly runtime bug might cause invalid state regarding, very unlikely, but better do double safe then sorry.

pallets/ema-oracle/src/lib.rs
dmoka: soft limit meaning there is no enforced bound for the AMM oracle entries in the storage. We only do the limit here. the term soft limit might be not the best

our goal is to make external oracle updates always pass, they can never fail, otherwise it is a big issue as prices out of sync. therefore there is no contraint on external oracles, only on AMM ones.

pallets/ema-oracle/src/lib.rs
martinfridrich: If we want to remove it in the future yes. We should keep all tests so we know we didn't break it but rest should use new fn.

pallets/ema-oracle/src/lib.rs
martinfridrich: So is it ok to not have updates from AMM? Why AMMs are less important?

dmoka: if the update fails for amm, then the whole trade fails.

External oracle price updates are more important as they are less frequent and if any would fail then it would lead to invalid oracle price that can be exploited.

pallets/ema-oracle/src/lib.rs
dmoka: fair enough, let me add do that change.

pallets/omnipool/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-omnipool"
-version = "7.3.0"
+version = "7.3.1"
```
enthusiastmartin, line 3: why change since nothing changed in omnipool pallet ?

PR #1370, opened 2026-02-19
feat: Signet Initialization

iamyxsh:
created a script that will initialize the signet pallet

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,5 @@
+{
+  "dependencies": {
+    "@polkadot/api": "^16.5.4"
+  }
+}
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@polkadot/api": "^16.5.4"
+  }
+}
```
enthusiastmartin, line 5: this file should not be here. nor i think it is needed anywhere it is not repo. if it is , it should be in the corresponding subdir.

enthusiastmartin:
in #1380

Issue #1296, opened 2025-12-18
Bug: set currency charges with previous fee currency


enthusiastmartin:
Setting fee payment asset via dispatch_with_extra_gas still charges the fee in the previous set currency.

Eg. setting atoken as fee payment asset 

Note that calling just set_currency charges correctly in the selected asset. 

Possible fix:
use currency override similar to dispatch-permit.

Issue #1293, opened 2025-12-16
Bug: Bad behaviour of erc20 is being setup like fee payment


DuriSK:
I believe, if user is setting asset as fee payment, this transaction should be already paid in new currency.
This is not the case for erc20

Comparing this two operations its clear that changing to DOT, i paid already fee in DOT
When i was switching to Hollar, i was still charged in HUSDx

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2F2.lark.hydration.cloud#/explorer/query/0xb1584591b7c7fa426383fd6618c45abeadaa8b785edf4b4311be62660295552b

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2F2.lark.hydration.cloud#/explorer/query/0x637acd2dc1b7c7cf124e14407ccb7fb372101c1b00d67ba385794ca13eae1096

mrq1911:
will be fixed by #1296 ?

PR #1372, opened 2026-02-26
fix: dispatch with extra gas transaction charges previous currency

khuzama98:
## Description

Fixes fee currency resolution for `dispatch_with_extra_gas` wrapping `set_currency` calls.
When users submitted `dispatch_with_extra_gas { set_currency { currency: X } }`, the transaction fee was incorrectly charged in the **previous** account currency instead of the newly specified currency `X`. Direct `set_currency` calls worked correctly.

This affected both substrate accounts (issue #1296) and EVM accounts (issue #1293).

## Root Cause

The pallet's `resolve_currency_from_call` function had duplicate currency resolution logic that was missing the `dispatch_with_extra_gas` handler, while the runtime's `TryCallCurrency` implementation already handled it correctly. The two resolvers diverged over time.

## Fix

Replaced the duplicate pattern-matching logic in `resolve_currency_from_call` with a simple delegation to `T::TryCallCurrency::try_convert()`. This ensures both code paths use the same authoritative resolver and eliminates future divergence.

## Related Issue

Fixes: #1296 #1293 

## How Has This Been Tested?

Integration tests have been added to test the scenario in which new currencies are being charged successfully.

## Checklist:

### DIFF
--- a/integration-tests/src/xcm_aliasers.rs
+++ b/integration-tests/src/xcm_aliasers.rs
@@ -308,3 +308,95 @@ fn asset_hub_root_can_alias_ethereum_accounts() {
 		);
 	});
 }
+
+#[test]
+fn asset_hub_root_can_alias_system_parachain_descendants() {
+	Hydra::execute_with(|| {
+		let origin = Location::new(1, X1([Parachain(1000)].into()));
+
+		// All real system parachains + boundary
+		let system_parachains: [(u32, &str); 6] = [
+			(1000, "Asset Hub"),
+			(1001, "Collectives"),
+			(1002, "BridgeHub"),
+			(1004, "People"),
+			(1005, "Coretime"),
+			(1999, "boundary (last valid)"),
+		];
+
+		for (id, name) in system_parachains {
+			// AccountId32 descendant
+			let target = Location::new(
+				1,
+				X2([
+					Parachain(id),
+					AccountId32 {
+						network: None,
+						id: [1u8; 32],
+					},
+				]
+				.into()),
+			);
+			assert!(
+				<XcmConfig as xcm_executor::Config>::Aliasers::contains(&origin, &target),
+				"Asset Hub root must be able to alias to an account on {name} ({id})"
+			);
+
+			// PalletInstance descendant
+			let target = Location::new(1, X2([Parachain(id), PalletInstance(8)].into()));
+			assert!(
+				<XcmConfig as xcm_executor::Config>::Aliasers::contains(&origin, &target),
+				"Asset Hub root must be able to alias to a pallet on {name} ({id})"
+			);
+
+			// Deeper nested: PalletInstance + GeneralIndex
+			let target = Location::new(1, X3([Parachain(id), PalletInstance(8), GeneralIndex(9)].into()));
+			assert!(
+				<XcmConfig as xcm_executor::Config>::Aliasers::contains(&origin, &target),
+				"Asset Hub root must be able to alias to deeper locations on {name} ({id})"
+			);
+		}
+	});
+}
+
+#[test]
+fn asset_hub_root_cannot_alias_non_system_parachain_descendants() {
+	Hydra::execute_with(|| {
+		let origin = Location::new(1, X1([Parachain(1000)].into()));
+
+		let non_system_parachains: [u32; 4] = [2000, 2001, 3000, 9999];
+
+		for id in non_system_parachains {
+			// AccountId32 descendant
+			let target = Location::new(
+				1,
+				X2([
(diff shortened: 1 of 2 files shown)


integration-tests/src/non_native_fee.rs
```diff
@@ -170,6 +171,166 @@ fn set_currency_should_work_in_batch_transaction_when_first_tx() {
...
+
+		let bob_hdx_after = hydradx_runtime::Balances::free_balance(AccountId::from(BOB));
+		let bob_btc_after = hydradx_runtime::Tokens::free_balance(BTC, &AccountId::from(BOB));
+
+		// Fee must be charged in BTC (the new currency declared inside dispatch_with_extra_gas).
+		assert!(
+			bob_btc_after < bob_btc_before,
+			"BTC balance should decrease — fee must be charged in the new currency"
+		);
+		// HDX must not be touched.
```
dmoka:
```suggestion

```

integration-tests/src/non_native_fee.rs
```diff
@@ -170,6 +171,166 @@ fn set_currency_should_work_in_batch_transaction_when_first_tx() {
...
+			1_000_000_000_000_000_000i128,
+		));
+		assert_ok!(hydradx_runtime::Currencies::update_balance(
+			hydradx_runtime::RuntimeOrigin::root(),
+			evm_acc.clone(),
+			DAI,
+			1_000_000_000_000_000_000i128,
+		));
+
+		// EVM accounts have WETH automatically set as their fee currency on account creation
```
dmoka:
```suggestion
```

integration-tests/src/non_native_fee.rs
```diff
@@ -170,6 +171,166 @@ fn set_currency_should_work_in_batch_transaction_when_first_tx() {
...
+
+		assert_ok!(
+			pallet_transaction_payment::ChargeTransactionPayment::<hydradx_runtime::Runtime>::from(0)
+				.validate_and_prepare(Some(evm_acc.clone()).into(), &call, &info, len, 0,)
+		);
+
+		let weth_after = hydradx_runtime::Tokens::free_balance(WETH, &evm_acc);
+		let dai_after = hydradx_runtime::Tokens::free_balance(DAI, &evm_acc);
+
+		// Fee must be charged in DAI (the new currency), not WETH (EVM default).
```
dmoka:
```suggestion
```

dmoka: i am removing these comments because in the assert you place very similar error message, so the comments feel redundant. small details...

integration-tests/src/non_native_fee.rs
```diff
@@ -170,6 +171,166 @@ fn set_currency_should_work_in_batch_transaction_when_first_tx() {
...
+		assert!(
+			dai_after < dai_before,
+			"DAI balance should decrease — fee must be charged in the new currency"
+		);
+		// WETH must not be touched.
```
dmoka:
```suggestion
```

integration-tests/src/non_native_fee.rs
```diff
@@ -170,6 +171,166 @@ fn set_currency_should_work_in_batch_transaction_when_first_tx() {
...
+		let bob_btc_before = hydradx_runtime::Tokens::free_balance(BTC, &AccountId::from(BOB));
+
+		assert_ok!(
+			pallet_transaction_payment::ChargeTransactionPayment::<hydradx_runtime::Runtime>::from(0)
+				.validate_and_prepare(Some(AccountId::from(BOB)).into(), &call, &info, len, 0,)
+		);
+
+		let bob_btc_after = hydradx_runtime::Tokens::free_balance(BTC, &AccountId::from(BOB));
+
+		// BTC must not be charged — the resolver did not reach the nested set_currency.
```
dmoka:
```suggestion

```

dmoka:
I cannot run `cargo test` command from root `hydration-node/pallets/transaction-multi-payment`, could you please fix that?

you prolly need to do add std feature in cargo.toml, cor currencies

```
[dev-dependencies]
pallet-currencies = { workspace = true, features = ["std"] }
...
```

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -821,30 +821,13 @@ where
 
 impl<T, MC, DF, FR, WF> TransferFees<T, MC, DF, FR, WF>
 where
-	T: Config + pallet_utility::Config,
+	T: Config,
```
dmoka: i think you can do the same bound cleanups in the `OnChargeTransaction` impl for TransferFees

PR #1375, opened 2026-03-02
fix: Changed path, Added robust testing

iamyxsh:
- changed the path used for deriving a signing address to be same for each requester
- refactored tests for multiple account testing

## Checklist:


enthusiastmartin:
in #1380

PR #1376, opened 2026-03-03
fix: benchmarking test for dispenser pallet

iamyxsh:
## Checklist:


enthusiastmartin:
in #1380

PR #1379, opened 2026-03-06
fix: dispenser pallet derivation path and output schema

esaminu:
## Description

- Use a fixed `"dispenser"` derivation path instead of per-user path so all requests derive the same MPC-controlled ETH address
- Set `output_deserialization_schema` to valid empty JSON array (`[]`) instead of empty `Vec`
- Update faucet contract address to deployed Sepolia contract
- Fix test key derivation to include `predecessorId` and `path` params
- Fix test to read faucet address from chain instead of `.env`

## How Has This Been Tested?

End-to-end dispenser test on Chopsticks + Sepolia with MPC response server. Request ID matches, MPC signs with correct derived key, EVM transaction confirmed on Sepolia, bidirectional response received on Substrate.


enthusiastmartin:
part of #1380

PR #1380, opened 2026-03-06
fix: all the fixes for the dispenser pallet

iamyxsh:
Standardize pallet_signet and pallet_dispenser configuration: replace one-time initialize with repeatable set_config, move compile-time constants into runtime-configurable storage structs, and
   use a fixed MPC signing path so all users derive the same EVM wallet.                                                                                                                        
                                                                                                                                                                                                  
  Pallet changes                                                                                                                                                                                  

  pallet_signet
  - Removed initialize, Initialized, Admin storage
  - Added SignetConfigData<Balance> struct + set_config/pause/unpause extrinsics gated by UpdateOrigin
  - Moved MaxChainIdLength, MaxEvmDataLength, MaxSignatureDeposit from Config trait into struct; added MAX_CHAIN_ID_LENGTH = 128 pallet constant

  pallet_dispenser
  - Removed initialize, set_faucet_balance, FaucetBalanceWei storage
  - Moved MinimumRequestAmount, MaxDispenseAmount, DispenserFee, FaucetAddress, MinFaucetEthThreshold from Config trait into DispenserConfigData struct (also includes paused, faucet_balance_wei)
  - Added set_config/pause/unpause extrinsics; first set_config defaults to unpaused
  - Fixed SIGNING_PATH = b"dispenser" — single MPC-derived EVM wallet for all users

  Runtime changes

  - Removed obsolete parameter_types and simplified both pallet Config impls
  - Both use UpdateOrigin = EitherOf<EnsureRoot, TechCommitteeMajority>
  - Updated weight files for new extrinsics

  Script changes

  - tc-set-config.ts (new): Sets both pallet configs — auto-detects chopsticks (direct storage write) vs live network (TC proposal)
  - utils.ts: Added dev chain detection, retry with tip escalation + nonce polling + block scanning fallback, referendum reuse/idempotency, updated initializeVaultIfNeeded for new setConfig API
  - signet-client.ts/dispenser.test.ts: Updated for setConfig, fixed signing path, palletSS58Prefix0 for request IDs, block-scanning fallbacks, live chain compatibility


pallets/dispenser/src/lib.rs
```diff
@@ -363,19 +351,68 @@ pub mod pallet {
...
-			if DispenserConfig::<T>::get().is_none() {
-				DispenserConfig::<T>::put(DispenserConfigData { paused: true });
-			} else {
-				DispenserConfig::<T>::mutate_exists(|p| p.as_mut().unwrap().paused = true);
-			};
+			<T as pallet::Config>::UpdateOrigin::ensure_origin(origin)?;
+			DispenserConfig::<T>::mutate(|maybe| {
+				if let Some(cfg) = maybe.as_mut() {
+					cfg.paused = true;
+				}
```
dmoka: if DispenserConfig is None, we might wanna return `Error::<T>::NotConfigured`

otherwise we emit `Paused` event, which might be misleading

pallets/dispenser/src/lib.rs
```diff
@@ -363,19 +351,68 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			faucet_address: EvmAddress,
+			min_faucet_threshold: Balance,
+			min_request: Balance,
+			max_dispense: Balance,
+			dispenser_fee: Balance,
+			faucet_balance_wei: Balance,
+		) -> DispatchResult {
+			<T as pallet::Config>::UpdateOrigin::ensure_origin(origin)?;
+
```
dmoka, line 380: i would add some validations for input data to make sure we dont make mistake with setting config

pallets/dispenser/src/lib.rs
dmoka: is it safe to unwrap it to emptry string? should not we return error/none in such case?

pallets/dispenser/src/lib.rs
```diff
@@ -474,22 +481,8 @@ pub mod pallet {
 
 	impl<T: Config> Pallet<T> {
```
dmoka: we have dual impl block, we could merge the two into one.

pallets/dispenser/src/lib.rs
```diff
@@ -161,6 +150,14 @@ pub mod pallet {
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
+		/// Dispenser configuration has been set or updated.
+		ConfigUpdated {
+			faucet_address: EvmAddress,
+			min_faucet_threshold: Balance,
+			min_request: Balance,
+			max_dispense: Balance,
+			dispenser_fee: Balance,
```
dmoka, line 157: maybe we could add `faucet_balance_wei ` to this, as it is also part of config data

PR #1385, opened 2026-03-10
feat: dca monitor script

dmoka:
when we apply some new proposal or runtime upgrade, there is chance that we break existing and important treasury like the polkadot treasury DCA.

To prevent such issues, this PR adds a monitoring script that checks for dca failures.

tuned for chopstick (with block rolling included)


dmoka:
closed as old was corrupted with other commits https://github.com/galacticcouncil/hydration-node/pull/1386

Issue #1133, opened 2025-06-17
Fix EVM gas estimation


mrq1911:
- gas estimations is inconsistent with weight estimation
- gas estimation when calling proxy is very overestimated

mrq1911:
probably related to this, `estimateGas` returns insane value when the gas limit s not specified (None)

<img width="1527" height="1092" alt="Image" src="https://github.com/user-attachments/assets/a5e18c94-0cb8-4df7-8d15-70886cc20886" />

when there is some sufficiently high gas limit set it works like expected

mrq1911:
Another case: 

> Steps to reproduce "exceeds block gas limit" error on Talisman + Hydration

Environment:
- Wallet: Talisman browser extension
- Chain: Hydration (Chain ID: 222222)
- RPCs: https://rpc.hydradx.cloud, https://hydration.ibp.network, etc.

Contracts involved:
- DecentralPool (deposit contract): 0x207a626c07b73E76134177D1f44B0f32e94ADB5a
- Stablecoin (ERC20 token): 0x531a654d1696ED52e7275A8cede955E82620f99a (18 decimals)

Prerequisites:
- Install the Talisman browser extension
- Import or create an account that holds the stablecoin above on Hydration EVM
- Ensure the account has WETH for gas on Hydration EVM

Steps:
1. Approve the stablecoin for spending by the DecentralPool contract (this step works fine)
2. Call deposit(uint256 amount) on 0x207a626c07b73E76134177D1f44B0f32e94ADB5a with amount = 10000000000000000000 (10 tokens, 18 decimals)
3. When sending the transaction, provide an explicit gas parameter of 300000 (0x493e0)
4. Talisman will pop up asking for confirmation

Expected behavior:
- Talisman should use the provided gas limit of 300,000
- Transaction should be signed and broadcast with gas = 300,000
- Transaction should succeed on-chain

Actual behavior:
- Talisman ignores the explicit gas: 300000 parameter
- Talisman runs its own eth_estimateGas which returns ~46M (0x2bd9b93) due to broken gas estimation on the proxy contract
- Talisman buffers that to ~46.9M (0x2cba3ce) and signs an EIP-1559 transaction with that gas limit (also ignoring our type: "legacy" parameter)
- The node rejects with "exceeds block gas limit" because ~46.9M exceeds Hydration's per-transaction gas limit (~45M)
- Talisman retries sending the same signed transaction multiple times, all failing

[...]

PR #1387, opened 2026-03-11
fix: EVM gas estimation by enabling rpc-binary-search-estimate

iamyxsh:
## Summary

Closes https://github.com/galacticcouncil/hydration-node/issues/1133

- Enable the `rpc-binary-search-estimate` feature flag on `fc-rpc`, which makes `eth_estimateGas` perform a binary search instead of returning the inflated gas-forwarded amount
- Add a full end-to-end test environment (`scripts/gas-estimation-test/`) that reproduces the bug and verifies the fix

## Problem

Contracts that make multiple external calls (like DecentralPool's `deposit()`) trigger EIP-150's 63/64th gas rule. Without binary search, `eth_estimateGas` runs the call at the block gas limit and reports the total gas *forwarded* to subcalls — not what was actually consumed. This inflates estimates by ~183x (54M vs 297K actual), causing wallets like Talisman to either exceed the block gas limit or massively overpay for gas.

## Changes

**Runtime fix (1 line):**
- `Cargo.toml`: Add `features = ["rpc-binary-search-estimate"]` to the `fc-rpc` workspace dependency

**Test environment (`scripts/gas-estimation-test/`):**
- `DepositProxy.sol` — Solidity contract with 6 external calls per `deposit()`, mimicking DecentralPool
- `build-chainspec.sh` — Builds a patched raw chain spec with WETH registered, test accounts funded, and `IsTestnet=true` for fast governance
- `setup-chain.js` — Sets WETH XCM location + adds WETH as accepted fee currency via GeneralAdmin referendum
- `test-gas-estimation.js` — Deploys contracts, calls `eth_estimateGas`, executes the real tx, and compares results
- `zombienet.json` — Local parachain config (4 relay validators, 1 collator)
- Full step-by-step README

## Test results

| Metric | Without fix | With fix |
|--------|------------|----------|
| Actual gas used | 297,192 | 297,192 |
| `eth_estimateGas` (no cap) | 54,347,670 (183x) | ~300,000 (1.0x) |

## Test plan

- [ ] Run the reproduction steps in `scripts/gas-estimation-test/README.md`
- [ ] Verify `eth_estimateGas` returns a value close to actual gas usage (~1-2x)
- [ ] Verify wallets (Talisman) can submit transactions to proxy contracts without "exceeds block gas limit" errors

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -330,7 +330,7 @@ parachains-common = { git = "https://github.com/galacticcouncil/polkadot-sdk", b
 fc-consensus = { git = "https://github.com/moonbeam-foundation/frontier", branch = "moonbeam-polkadot-stable2503", default-features = false }
 fc-db = { git = "https://github.com/moonbeam-foundation/frontier", branch = "moonbeam-polkadot-stable2503", default-features = false }
 fc-mapping-sync = { git = "https://github.com/moonbeam-foundation/frontier", branch = "moonbeam-polkadot-stable2503", default-features = false }
-fc-rpc = { git = "https://github.com/moonbeam-foundation/frontier", branch = "moonbeam-polkadot-stable2503", default-features = false }
+fc-rpc = { git = "https://github.com/moonbeam-foundation/frontier", branch = "moonbeam-polkadot-stable2503", default-features = false, features = ["rpc-binary-search-estimate"] }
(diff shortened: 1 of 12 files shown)


mrq1911:
fixes #1133

PR #1388, opened 2026-03-11
feat: security_audit skill

cl0w5:


CLAUDE.md
```diff
@@ -0,0 +1,125 @@
...
+- **XCM** for cross-chain asset transfers — config in `runtime/hydradx/src/xcm/`.
+- **Circuit breaker** pallet for risk management — limits large trades/liquidity changes.
+
+## Testing guidelines
+
+- All features and bug fixes **must have tests**
+- Unit tests go in `pallets/<name>/src/tests/`
+- Integration tests in `integration-tests/` run against the full runtime
+- Mock runtimes: each pallet has its own `mock.rs`; `runtime-mock/` for shared mocking
+- Property testing available via `proptest`
```
dmoka, line 115: i would add one more list item about using `debug_assert!` at unexpected states, so our fuzzer can catch those.

CLAUDE.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+## Key patterns
+
+- **Weights are auto-generated** from benchmarks — never edit `weights.rs` files by hand. Use `scripts/benchmarking.sh` or add `[ignore benchmarks]` to PR title to skip.
+- **Math-heavy code** lives in `math/` crate with `fixed` and `rug` for arbitrary precision. Pallets call into math functions — keep math logic separate from storage/dispatch logic.
+- **ORML integration** for multi-currency support (orml-tokens, orml-currencies).
+- **Frontier/EVM** integration for Ethereum compatibility — EVM-related pallets include `dynamic-evm-fee`, `evm-accounts`, and precompiles.
+- **XCM** for cross-chain asset transfers — config in `runtime/hydradx/src/xcm/`.
+- **Circuit breaker** pallet for risk management — limits large trades/liquidity changes.
```
dmoka, line 107: I would add one more about using checked math wherever possible

```
Always use checked arithmetic methods (`checked_add`, `checked_sub`, `checked_mul`, `checked_div`, etc.) for all math operations. Never use bare operators (`+`, `-`, `*`, `/`). Always propagate the resulting `Option` with `?` — never `unwrap` or `expect` on math results.
```

or maybe it belongs to `## Code style` section

dmoka:
Cool stuff!

dmoka:
how about adding the claude skills into this repo too, instead of [separating it here ](https://github.com/galacticcouncil/hydration)

by doing so we can easily collect/add skills needed for hydration runtime.

Issue #1378, opened 2026-03-04
dispatch as emergency admin


mrq1911:
add function similar to dispatch as aave admin that will allow TC to act as some address, which can be then configured in smart contracts as emergency admin that could pause some market for example

dmoka:
# Spec: `dispatch_as_emergency_admin` for TC

## Motivation

The current path to interact with Aave contracts is `dispatch_as_aave_manager`, which requires an `EconomicParameters` referendum which might take several days. TC needs a faster path to react to emergencies.

## Design

Add a new `dispatch_as_emergency_admin` extrinsic to the dispatcher pallet. It follows the exact same pattern as the existing `dispatch_as_aave_manager`, but:

- **Origin:** TC majority (or Root), instead of `EconomicParameters`
- **Account:** A separate configurable emergency admin address, instead of the aave manager address

On the Aave side, this address will be registered as the emergency admin via `ACLManager.addEmergencyAdmin()`.

## Changes

### 1. Pallet config (`pallets/dispatcher/src/lib.rs`)

Add two new associated types to the `Config` trait:

```rust
type EmergencyAdminOrigin: EnsureOrigin<Self::RuntimeOrigin>;
type DefaultEmergencyAdminAccount: Get<Self::AccountId>;
```

### 2. Storage (`pallets/dispatcher/src/lib.rs`)

Add one new storage item:

```rust
#[pallet::storage]
#[pallet::getter(fn emergency_admin_account)]
pub type EmergencyAdminAccount<T: Config> =
    StorageValue<_, T::AccountId, ValueQuery, T::DefaultEmergencyAdminAccount>;
```

### 3. Events (`pallets/dispatcher/src/lib.rs`)

Add two new event variants:

```rust
EmergencyAdminCallDispatched {
    call_hash: T::Hash,
    result: DispatchResultWithPostInfo,
},
EmergencyAdminAccountNoted {
    account: T::AccountId,
},
```

### 4. Extrinsics (`pallets/dispatcher/src/lib.rs`)

**`dispatch_as_emergency_admin`**

Mirrors `dispatch_as_aave_manager`. TC (or Root) calls this; the pallet dispatches the inner call as the emergency admin address.

```rust
#[pallet::call_index(5)]
#[pallet::weight({
    let call_weight = call.get_dispatch_info().call_weight;
    let call_len = call.encoded_size() as u32;
    T::WeightInfo::dispatch_as_emergency_admin(call_len)
[...]

PR #1389, opened 2026-03-13
feat: dispatch as emergency admin

dmoka:
Spec: https://github.com/galacticcouncil/hydration-node/issues/1378#issuecomment-4048593791

TODO:
- rebench on ref machine


pallets/dispatcher/src/lib.rs
```diff
@@ -103,6 +105,11 @@ pub mod pallet {
 	#[pallet::getter(fn aave_manager_account)]
 	pub type AaveManagerAccount<T: Config> = StorageValue<_, T::AccountId, ValueQuery, T::DefaultAaveManagerAccount>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn emergency_admin_account)]
+	pub type EmergencyAdminAccount<T: Config> =
```
mrq1911: no need for this storage, the account should be hardcoded

pallets/dispatcher/src/lib.rs
```diff
@@ -328,6 +344,61 @@ pub mod pallet {
...
+			T::EmergencyAdminOrigin::ensure_origin(origin)?;
+
+			let call_hash = T::Hashing::hash_of(&call);
+			let call_len = call.encoded_size() as u32;
+
+			let (result, actual_weight) = Self::do_dispatch(
+				frame_system::Origin::<T>::Signed(EmergencyAdminAccount::<T>::get()).into(),
+				*call,
+			);
+			actual_weight.map(|w| w.saturating_add(T::WeightInfo::dispatch_as_emergency_admin(call_len)));
```
mrq1911: whats the purpose of this? you are not actually setting that weight? implementation should be the same as in  macro weight definition

PR #1391, opened 2026-03-18
feat: charge controller for EVM gas fees when using proxy

### LINKED ISSUE #1381: Handling of WETH for pureProxy accounts
### Problem

When using strategies, the frontend creates a new `pureProxy` for every position. To execute EVM actions on these proxy accounts, the controller calls `proxy.proxy(real=pureProxy, call=evm.call(...))`.

Currently:
- **Substrate extrinsic fee** → charged to the controller (correct)
- **EVM gas fee (WETH)** → charged to the pureProxy (broken)

Since pureProxy accounts are created with zero balance, EVM calls silently fail. The workaround of manually swapping and transferring WETH to each proxy is clunky and doesn't scale.

The root cause: by the time `TransferEvmFees::withdraw_fee()` executes, `pallet_proxy` has already swapped the origin to the pureProxy. The controller's identity is lost.

### Solution

Introduce a **thread-local fee payer override**, following the same pattern as `tx_fee_currency_override` in `pallet_transaction_multi_payment` — a thread-local that carries context through the dispatch pipeline.

We need a thread-local because `withdraw_fee()` is a trait method with a fixed signature (`fn withdraw_fee(who: &H160, fee: U256, ...)`) defined by Frontier's `OnChargeEVMTransaction`. We can't add a `fee_payer` parameter without forking the trait. The thread-local sidesteps this by carrying context out-of-band — exactly how `pallet_transaction_multi_payment` already handles fee currency overrides.

#### Components

1. **`EvmFeePayer` thread-local** — stores the real fee payer (`Option<AccountId>`) for the duration of the dispatch
2. **`SetEvmFeePayerExtension` (SignedExtension)** — inspects top-level calls in `pre_dispatch`:
   - If call matches `proxy.proxy(real, _, Box(evm.call(...)))` → stores the signer (controller) as fee payer
   - Clears in `post_dispatch`
3. **`WrapRunner::validate()`** — checks the controller's WETH balance instead of the pureProxy's when fee payer override is set
4. **`TransferEvmFees::withdraw_fee()`** — burns WETH from the controller instead of the pureProxy
5. **`TransferEvmFees::correct_and_deposit_fee()`** — refunds overpayment to the controller

#### Flow

```
Controller signs proxy.proxy(real=pureProxy, call=evm.call(...))
        │
        ▼
SignedExtension::pre_dispatch()
  → sees proxy wrapping an EVM call
  → writes controller's AccountId to thread-local
        │
        ▼
proxy.proxy() dispatches → origin becomes pureProxy
        │
        ▼
EVM executes → withdraw_fee() runs
  → reads thread-local → finds controller
  → charges controller instead of pureProxy
        │
        ▼
SignedExtension::post_dispatch()
  → clears the thread-local
```

### Testing

- [ ] Proxy EVM call → gas charged to controller, not pureProxy
- [ ] Direct EVM call → gas still charged to source (no regression)
- [ ] Fee refund goes to controller
- [ ] Controller with insufficient WETH → fails gracefully
- [ ] Benchmark weights updated for additional storage reads

Closes #1381

iamyxsh:
- When `proxy.proxy(real=pureProxy, call=evm.call(...))` is dispatched, EVM gas fees are now charged to the controller
  (transaction signer) instead of the pureProxy
  - Adds `SetEvmFeePayer` TransactionExtension that detects proxy-wrapping-EVM patterns and threads the controller's AccountId
   to the fee charging layer via a thread-local
  - Handles `proxy`, `proxy_announced`, and nested `utility.batch/batch_all/force_batch/as_derivative` patterns

  Closes #1381

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6378,7 +6378,7 @@ dependencies = [
 
 [[package]]
 name = "hydradx-runtime"
-version = "407.0.0"
+version = "408.0.0"
 dependencies = [
  "alloy-primitives 0.7.7",
  "alloy-sol-types 0.7.7",
@@ -10206,7 +10206,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-dispatcher"
-version = "1.7.1"
+version = "1.7.2"
 dependencies = [
  "frame-benchmarking",
  "frame-support",
--- a/pallets/dispatcher/Cargo.toml
+++ b/pallets/dispatcher/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-dispatcher"
-version = "1.7.1"
+version = "1.7.2"
 authors = ['GalacticCouncil']
 edition = "2021"
 license = "Apache-2.0"
--- a/runtime/hydradx/Cargo.toml
+++ b/runtime/hydradx/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hydradx-runtime"
-version = "407.0.0"
+version = "408.0.0"
 authors = ["GalacticCouncil"]
 edition = "2021"
 license = "Apache 2.0"
--- a/runtime/hydradx/src/lib.rs
+++ b/runtime/hydradx/src/lib.rs
@@ -129,7 +129,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("hydradx"),
 	impl_name: Cow::Borrowed("hydradx"),
 	authoring_version: 1,
-	spec_version: 407,
+	spec_version: 408,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,


runtime/hydradx/src/evm/evm_fee.rs
```diff
@@ -307,3 +385,281 @@ impl AccountFeeCurrency<AccountId> for FeeCurrencyOverrideOrDefault {
...
+		self,
+		_val: Self::Val,
+		origin: &crate::RuntimeOrigin,
+		call: &crate::RuntimeCall,
+		_info: &DispatchInfoOf<crate::RuntimeCall>,
+		_len: usize,
+	) -> Result<Self::Pre, TransactionValidityError> {
+		if is_proxy_wrapping_evm(call) {
+			if let Ok(signer) = frame_system::ensure_signed(origin.clone()) {
+				set_evm_fee_payer(signer.clone());
```
mrq1911: instead of extension just let proxy to set `set_evm_fee_payer`

traits/src/evm.rs
```diff
@@ -133,3 +133,17 @@ pub trait ExtraGasSupport {
...
 }
+
+/// Support for overriding the EVM fee payer.
+/// Used when a controller account should pay EVM gas fees
+/// on behalf of another account (e.g., a pureProxy).
+pub trait EvmFeePayerSupport {
+	type AccountId;
+
+	/// Set the fee payer override, returning the previous value (for nesting support).
+	fn set_fee_payer(payer: Self::AccountId) -> Option<Self::AccountId>;
```
mrq1911, line 143: having both set_fee_payer and restore fee payer seems redundant. restore fee payer should be achievable by set_fee_payer

PR #1394, opened 2026-03-20
feat: reverted H2O treasury mechanic

mrq1911:
reverted h2o treasury routing mechanism from omnipool sell_hub_asset and buy_asset_for_hub_asset — lrna now flows into the traded asset's
  hub_reserve as before instead of being redirected to treasury
  - removed HubDestination config type from omnipool pallet and all runtime/mock configurations
  - restored increase_hdx_subpool_hub_reserve oracle hook and fee sync for protocol fee routing
  - removed process_protocol_fee origin parameter dependency
  - deleted integration-tests/src/omnipool_h2o_routing.rs and all h2o-specific unit tests (treasury routing, oracle hooks, hub destination bypass)
  - reverted expected numeric values across integration tests (dca, oracle, dynamic_fees, referrals, omnipool_liquidity_mining, circuit_breaker)
  - fixed flaky liquidity_mining::invariants::calculate_user_reward proptest that could overflow on edge-case inputs
  - bumped versions of all changed crates and runtime spec_version

  reverted prs

  - #1340 (fix/route-h2o-trade-to-hdx-subpool)
  - #1367 (fix/treasury-h2o-trade) — only h2o-specific changes; global withdraw limit removal preserved

### DIFF
--- a/math/src/liquidity_mining/invariants.rs
+++ b/math/src/liquidity_mining/invariants.rs
@@ -1,4 +1,4 @@
-use crate::{assert_approx_eq, types::Balance};
+use crate::{assert_approx_eq, types::Balance, MathError};
 use proptest::prelude::*;
 use sp_arithmetic::traits::One;
 use sp_arithmetic::{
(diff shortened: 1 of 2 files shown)


math/src/liquidity_mining/invariants.rs
```diff
@@ -197,20 +197,32 @@ proptest! {
...
+			Some(v) => v,
+			None => return Ok(()),
+		};
+
+		let p = match accumulated_rpvs_now
+			.checked_sub(accumulated_rpvs)
+			.and_then(|v| v.checked_mul(valued_shares)) {
+			Some(v) => v,
+			None => return Ok(()),
+		};
```
martinfridrich: I don't think these should/can fail if `calculate_user_reward()` passes
I would let these fail as before, to be safe

math/src/liquidity_mining/invariants.rs
```diff
@@ -197,20 +197,32 @@ proptest! {
...
-			.checked_add(unclaimable_rewards).unwrap()
-			.checked_add(accumulated_claimed_rewards).unwrap();
-
-		let p = accumulated_rpvs_now
-			.checked_sub(accumulated_rpvs).unwrap()
-			.checked_mul(valued_shares).unwrap();
+			loyalty_multiplier);
+
+		// Skip inputs that cause overflow — not a bug, just out-of-range values
+		let (user_rewards, unclaimable_rewards) = match result {
```
martinfridrich, line 208: I would check overflow error explicitly.
I know now this calc. can't fail on anything else but just in case if we would add other errors in the future.

pallets/omnipool/src/lib.rs
```diff
@@ -2001,32 +1992,18 @@ impl<T: Config> Pallet<T> {
 		);
 
 		let (taken_fee, trade_fees) = Self::process_trade_fee(who, asset_out, state_changes.fee.asset_fee)?;
-		let mut state_changes = state_changes.account_for_fee_taken(taken_fee);
-
-		ensure!(
-			matches!(state_changes.asset.delta_hub_reserve, BalanceUpdate::Increase(_)),
-			Error::<T>::HubAssetUpdateError
-		);
```
enthusiastmartin, line 2009: perhaps - this we could keep. it does not not hurt, and ensures that we only increasing hub side.

pallets/omnipool/src/lib.rs
```diff
@@ -2140,33 +2113,18 @@ impl<T: Config> Pallet<T> {
 			Error::<T>::MaxInRatioExceeded
 		);
 
-		ensure!(
-			matches!(state_changes.asset.delta_hub_reserve, BalanceUpdate::Increase(_)),
-			Error::<T>::HubAssetUpdateError
-		);
```
enthusiastmartin, line 2146: same here. it is just a safety double check.  we can keep thi.s

Issue #1412, opened 2026-04-07
Bug: fix max slip fee cap


enthusiastmartin:
There seems to be an issue with applying max slip fee cap on the buy side.

Investigate, confirm and fix. 

Note that this does not cause any problems - users are protected by their slippages limits, and even if, to reach the slip fee so high that it would have to capped by the slip fee max - that requires a huge trade that would push the sllp based fee so high - at that point trader looses so much on slippage, that he does not care much about the slip fee anyway.

enthusiastmartin:
fixed in #1456

PR #1413, opened 2026-04-07
fix: apply max_slip_fee cap in buy-side inversion

F3Joule:
## Description
Fixes a bug where `max_slip_fee` cap was not applied during buy-side trade inversion in the
Omnipool math, breaking the invariant `delta_hub_reserve_in − total_protocol_fee = D_net`.

Changes:
- `calculate_buy_state_changes` — after `invert_buy_side_slip`, check if cap fired and override
  `d_gross` with the capped linear formula; same check after `invert_sell_side_fees` for the
  sell-side leg of buy inversion
- Add regression tests `buy_slip_fee_cap_invariant_holds` and `sell_slip_fee_cap_invariant_holds`

## Related Issue
#1412 

## Motivation and Context
The forward path (`calculate_slip_fee_amount`) correctly caps the slip fee; the inverse path did
not - so the two paths produced inconsistent results when the slip rate exceeded `max_slip_fee`.
Sell trades are unaffected (no inversion involved).

## How Has This Been Tested?
Two regression tests, one per inversion step, each verified to fail before the fix and pass after.

## Checklist:


math/src/omnipool/tests.rs
```diff
@@ -3350,3 +3350,127 @@ fn cross_validate_buy_dot_with_lrna_with_prior_delta() {
 		*r_fresh.asset.delta_hub_reserve
 	);
 }
+
```
dmoka, line 3353: i would add a test for the case when both cap is fired, so the trade size is relative big compared to both asset in and asset out.

math/src/omnipool/tests.rs
```diff
@@ -3350,3 +3350,127 @@ fn cross_validate_buy_dot_with_lrna_with_prior_delta() {
...
+	// large pool: ~0.02% hub flow → cap does not fire
+	let asset_out_state = AssetReserveState {
+		reserve: 100_000_000 * UNIT,
+		hub_reserve: 100_000_000 * UNIT,
+		shares: 100_000_000 * UNIT,
+		protocol_shares: 0u128,
+	};
+
+	let amount = 20_000 * UNIT;
+	let asset_fee = Permill::zero();
```
dmoka, line 3434: in both tests the fees are zero, I would add a test with some fees, but surely with protocol fee > 0

math/src/omnipool/math.rs
dmoka, line 239: dont we need to fix the bug here too?!

math/src/omnipool/math.rs
```diff
@@ -277,14 +277,58 @@ pub fn calculate_buy_state_changes(
 
 	// Step 2: Invert buy-side slip to find D_gross from D_net
 	let d_gross = if let Some(slip) = slip {
-		invert_buy_side_slip(d_net, slip.asset_out_hub_reserve, slip.asset_out_delta)?
+		let d_gross_uncapped = invert_buy_side_slip(d_net, slip.asset_out_hub_reserve, slip.asset_out_delta)?;
```
enthusiastmartin, line 297: why we suddenly add bunch of code here ? same for the next change? 

i havent thought it but i. believe it should be just possible to pass the max fee into the function and handle it there ?

enthusiastmartin (changes requested):

math/src/omnipool/math.rs
F3Joule: I intentionally kept the cap logic in `calculate_buy_state_changes` to keep `invert_*` functions as pure math without business constraints. Passing `max_slip_fee` into them would work but couples the inversion logic to a policy decision.

Happy to go either way though - from my side it was just a matter of preference

enthusiastmartin:
in favor of https://github.com/galacticcouncil/hydration-node/pull/1456

Issue #1341, opened 2026-01-23
Vote with staking does uncessary vote processing.


enthusiastmartin:
It seems each call to vote tries to process all votes, resulting in vote extrinsic tobe quite heavy.

That should not be necessary. Investigate the reasons why we do that and if it can be removed.

PR #1423, opened 2026-04-15
fix: optimize weight consumption on process_votes in conviction voting

F3Joule:
## Description
This PR introduces new `GetReferendumState` impl: `DirectReferendumStatus` that reads referenda state using direct access to `ReferendumInfoFor` instead of the `frame_support::traits::voting::Polling` abstraction, which implementation in `pallet_conviction_voting` caused extra writes to happen.

## Related Issue
Related to #1341. Not a direct fix

## How Has This Been Tested?
No existing regression tests failed. Weights updated as expected.


cl0w5 (approved):

martinfridrich (commented):
This PR doesn't fixes #1341 , we are still processing all votes on vote.
This PR improves benchmarks because `try_access_poll()` is doing unnecessary write so your improvement is ok but it's not fixing linked issue.

PR #1424, opened 2026-04-17
fix(stableswap)!: fix peg ordering by accepting and co-sorting asset-peg pairs

### LINKED ISSUE #1164: stableswap asset order dependency

khuzama98:
## Description

Fixes a peg-ordering bug in `create_pool_with_pegs` where passing assets and peg
sources in different orders caused silent misalignment after internal sorting.

**Root cause:** `do_create_pool` sorted assets internally, but the corresponding
`peg_source` array was not co-sorted — so `peg_source[i]` could end up paired with
the wrong asset after sort.

**Changes:**
- **Breaking:** `create_pool_with_pegs` now accepts `BoundedVec<(AssetId, PegSource), _>`
  instead of separate `assets` and `peg_source` parameters. Pairing is now explicit at
  the call site, making misalignment impossible.
- Assets and peg sources are co-sorted by asset ID before pool creation; `do_create_pool`
  no longer sorts internally (callers are now responsible for providing sorted assets).
- `create_pool` (non-peg variant) sorts assets at dispatch.
- Added `debug_assert!` in `get_target_pegs` verifying assets arrive pre-sorted.

## Related Issue

Fixes: #1164  

## Motivation and Context

A pool created via `create_pool_with_pegs` with assets provided in non-ascending order
would store a peg source at an index that no longer matched its intended asset after the
internal sort. This caused incorrect peg values to be applied to assets, a correctness
issue with direct impact on pool pricing.

## How Has This Been Tested?

- added peg-ordering unit tests to 3-asset pools with generic test helpers
  (`pallets/stableswap/src/tests/peg_ordering.rs`)
- Updated all affected unit test suites: `peg.rs`, `peg_one.rs`, `peg_ordering.rs`,
  `pegs_with_different_decimals.rs`, `update_peg_source.rs`, `update_max_peg_update.rs`,
  `remove_liquidity.rs`
- Added integration regression test in `integration-tests/src/stableswap.rs` that
  creates a pool with out-of-order assets and asserts correct peg assignment post-sort

## Checklist:


pallets/stableswap/src/lib.rs
```diff
@@ -2192,6 +2216,10 @@ impl<T: Config> Pallet<T> {
 			peg_sources.len(),
 			"Pool assets and peg sources must have the same length"
 		);
+		debug_assert!(
```
dmoka, line 2227: next to debug assert, shall we make it prod runtime failure too?

just like we did recently for trades https://github.com/galacticcouncil/hydration-node/pull/1429

runtime/hydradx/src/weights/pallet-stableswap.rs
```diff
@@ -0,0 +1,626 @@
+// This file is part of HydraDX.
```
dmoka: the name of this is off, it must be with underscore so pallet_stableswap.rs

so this should not be a new file, rather a modification of our existing weight files.

We possibly have a bug in the benchmark workflow file, could you please check it out that too and fix? thanks

pallets/stableswap/src/lib.rs
```diff
@@ -1110,7 +1117,6 @@ pub mod pallet {
 		/// - `ShareAssetInPoolAssets`: If the share asset is among the pool assets.
 		/// - `AssetNotRegistered`: If one or more assets are not registered in the AssetRegistry.
 		/// - `InvalidAmplification`: If the amplification parameter is invalid.
-		/// - `IncorrectInitialPegs`: If the initial pegs are incorrect.
```
dmoka, line 1114: what is hte reason this was removed? i still see `IncorrectInitialPegs` error returned when creating pool

PR #1427, opened 2026-04-21
fix: add fallback price for global rate limiter

dmoka:
external-asset withdraw/XCM returned FailedToConvertAsset whenever the oracle route was missing.

Adding onchain route is a quick fix, but better to add conversion using the fallback price if no oracle price can be retrieved


runtime/hydradx/src/circuit_breaker.rs
```diff
@@ -26,14 +27,17 @@ impl<ReferenceCurrencyId: Get<AssetId>> WithdrawCircuitBreaker<ReferenceCurrency
 			return Ok(amount);
 		}
 
-		let (converted, _) = ConvertBalance::<TenMinutesOraclePrice, XykPaymentAssetSupport, DotAssetId>::convert((
+		ConvertBalance::<TenMinutesOraclePrice, XykPaymentAssetSupport, DotAssetId>::convert((
```
F3Joule, line 30: `convert_to_hdx` is generic over `ReferenceCurrencyId`, but the fallback assumes ref currency = HDX (since `currency_price` always returns price relative to the native asset). Current callers all pass `NativeAssetId` so this is fine, but consider a `debug_assert!` or comment to guard against future misuse.

```
ConvertBalance::<TenMinutesOraclePrice, XykPaymentAssetSupport, DotAssetId>::convert((
```

F3Joule (approved):
lgtm

F3Joule (approved):

PR #1429, opened 2026-04-24
feat: ensure invariants on prod

### LINKED ISSUE #1426: enforce omnipool/stableswap invariants
Currently we enforce invariants only when built with test or try-runtime features

Consider enforcing these invariants in production too.

they need to be refactor to not panic! 

the trade should fail if the invariant check fails. 

it should still panic in test or try-runtime features, so the fuzzer can catch such problems.

dmoka:
Fixes https://github.com/galacticcouncil/hydration-node/issues/1426


pallets/omnipool/src/lib.rs
```diff
@@ -1122,6 +1126,9 @@ pub mod pallet {
 			let asset_in_state = Self::load_asset_state(asset_in)?;
 			let asset_out_state = Self::load_asset_state(asset_out)?;
 
+			let balance_one =
```
enthusiastmartin: better naming of this variable please. balance_once does not express anything

pallets/omnipool/src/lib.rs
```diff
@@ -2645,86 +2651,118 @@ impl<T: Config> Pallet<T> {
...
+			.checked_sub(asset_out_state.hub_reserve)
+			.ok_or(Error::<T>::InvariantError)?;
+		let hdx_id = T::HdxAssetId::get();
+		if hdx_id != asset_in && hdx_id != asset_out {
+			let hdx_state = Self::load_asset_state(hdx_id)?;
+			balance = balance
+				.checked_sub(hdx_state.hub_reserve)
+				.ok_or(Error::<T>::InvariantError)?;
+		}
+		Ok(balance)
```
enthusiastmartin: i would do something like this 

1. get hub reserves for all assets that you want to subtract
2. sum it up = sum_asset_hub-reserves
3. ensure(hub_supply >= hub_asset_hub_reserve, InvalidOmnipoolHubReserve);
4. return hub_supply.saturating_sub(sum_asseT_hub_reserve)

the importnat is the new error - should never happened! - but if happens we see this error clearly.

so also add debug_assert for this, so fuzzer can catch this.

enthusiastmartin (changes requested):

enthusiastmartin (approved):

PR #1436, opened 2026-05-01
feat: broadcast substrate transfers and swaps as evm logs

mrq1911:
## summary

broadcast substrate-side token movements and trades through the same evm tooling that already works against hydration's evm (`eth_getLogs`, `eth_getTransactionReceipt`, etherscan, the graph, dex aggregators).

tres pieces:
- inline emit from precompiles when an event-producing op runs inside a real eth tx — log lands on the real tx
- `pallet-synthetic-logs`: buffers `(bucket, emitter, log)` entries and on `on_finalize` builds one synthetic `pallet_ethereum::Transaction` per execution-context bucket (extrinsic / hook+origin), so eth-rpc serves substrate-origin events like any other tx
- bucket = `frame_system::Phase` + `pallet_broadcast::ExecutionContext` first frame; encoded into the synth tx's `nonce` so indexers can reverse it back to the substrate origin. canonical `tx.hash()` is used (no domain-separated digest) — frontier indexes by it

phase 1: orml-tokens transfers + native HDX transfers → erc-20 `Transfer` log
phase 2: `pallet_broadcast::Swapped3` → uniswap v2 `Swap` log on a stable per-pool address (=filler's evm address)

## modified deps

- polkadot-sdk → [`polkadot-stable2506-10-patch-hooks`](https://github.com/galacticcouncil/polkadot-sdk/tree/polkadot-stable2506-10-patch-hooks) — adds `pallet-balances` post-mutation hook so native HDX transfers/deposits/burns reach the synth-logs buffer
- open-runtime-module-library → [`polkadot-stable2506-patch-hooks`](https://github.com/galacticcouncil/open-runtime-module-library/tree/polkadot-stable2506-patch-hooks) — extends `orml-tokens` hooks to cover withdraw/reserve paths


pallets/synthetic-logs/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+			from: SENTINEL_ADDRESS,
+			to: Some(SENTINEL_ADDRESS),
+			contract_address: None,
+			logs: logs.clone(),
+			logs_bloom: bloom,
+		};
+
+		let receipt = Receipt::EIP1559(ethereum::EIP658ReceiptData {
+			status_code: 1,
+			used_gas: U256::zero(),
```
dmoka, line 208: what is the reason that the gas used is zero?

as far as i see this is cumulatitvely added in frontier pallet, so if we set zero, we alway set the cumulative used gas to zero

https://github.com/moonbeam-foundation/frontier/blob/beba745d94f4e368adee4a4cddc014155fdbb5c7/frame/ethereum/src/lib.rs#L481

pallets/synthetic-logs/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+	pub type Pending<T: Config> = StorageValue<_, Vec<(Bucket, H160, ethereum::Log)>, ValueQuery>;
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_finalize(_n: BlockNumberFor<T>) {
+			let drained = Pending::<T>::take();
+			if drained.is_empty() {
+				return;
+			}
+			Self::flush(drained);
```
dmoka, line 95: dont we need to account for this weight somewhere?

In other pallets like hsm we do such weight accounting in `on_initialize`

Or i just see that `Pending` storage is whitelisted, so maybe no need to account for it...

runtime/hydradx/src/evm/erc20_logs.rs
```diff
@@ -0,0 +1,214 @@
...
+	}
+}
+
+impl OnSlashReserved<AccountId, AssetId, Balance> for EmitErc20TransferLog {
+	fn on_slash_reserved(asset: AssetId, who: &AccountId, amount: Balance) {
+		if amount == 0 {
+			return;
+		}
+		let owner = evm_address_of(who);
+		push_transfer_log(asset, reserved_address_of(owner), H160::zero(), amount);
```
dmoka, line 107: slash hook receives the to-be-slashed amount and [not the clamped one](https://github.com/galacticcouncil/open-runtime-module-library/blob/7e26f9a8fc47dc6f515b328f39c5fa7fc520223b/tokens/src/lib.rs#L1207-L1218) so if we pass it on to the log, then the log might overstates the amount.

Here is a test to reproduce it:

```
#[test]
fn orml_tokens_on_slash_log_amount_matches_actually_slashed() {
	TestNet::reset();
	Hydra::execute_with(|| {
		let held = <Tokens as MultiCurrency<AccountId>>::free_balance(DAI, &ALICE.into());
		assert!(held > 0, "test net must seed alice with some DAI for this scenario");
		let overshoot = UNITS;
		let requested = held + overshoot;

		SyntheticLogsPending::<Runtime>::kill();

		let unslashed =
			<Tokens as orml_traits::currency::MultiCurrency<AccountId>>::slash(DAI, &ALICE.into(), requested);

		// Sanity: orml reports the un-slashable remainder, and Alice's balance is now 0.
		assert_eq!(
			unslashed, overshoot,
			"orml should report the un-slashable remainder"
		);
		let post_balance = <Tokens as MultiCurrency<AccountId>>::free_balance(DAI, &ALICE.into());
		assert_eq!(post_balance, 0, "alice should hold zero DAI after the partial slash");

		let asset_addr = HydraErc20Mapping::asset_address(DAI);
		let (_, _, log) = buffered_logs()
			.into_iter()
			.find(|(_, emitter, log)| *emitter == asset_addr && log.topics[2] == h160_to_h256(H160::zero()))
			.expect("synth log for DAI slash");

		let logged_amount = U256::from_big_endian(&log.data);
		assert_eq!(
			logged_amount,
			U256::from(held),
			"synth Transfer amount must equal what was actually slashed ({held}), not the requested amount ({requested})",
		);
	});
}
```

pallets/synthetic-logs/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+			Self::insert_synth_tx(bucket, logs, idx as u32);
+		}
+	}
+
+	fn insert_synth_tx(bucket: Bucket, logs: Vec<ethereum::Log>, group_index: u32) {
+		let chain_id = <T as Config>::ChainId::get();
+		let nonce = bucket_nonce(bucket);
+
+		let signature = ethereum::eip2930::TransactionSignature::new(false, SYNTH_SIG_RS, SYNTH_SIG_RS)
+			.expect("synthetic signature constants are within valid ECDSA range; qed");
```
dmoka, line 171: I would try to avoid using `expect` as it panics

pallets/synthetic-logs/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+			v.push((bucket, emitter, log));
+		});
+	}
+
+	// reads `frame_system::ExecutionPhase` directly so on_idle (where substrate
+	// has already set Phase=Finalization) buckets as Finalization too.
+	fn current_bucket() -> Bucket {
+		use frame_system::Phase;
+		let phase_key = frame_support::storage::storage_prefix(b"System", b"ExecutionPhase");
+		let phase: Phase = frame_support::storage::unhashed::get::<Phase>(&phase_key).unwrap_or_default();
```
dmoka, line 123: maybe using `defensive_unwrap_or_default()` ?

pallets/synthetic-logs/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+		// `value = group_index` so two synth txs sharing the same bucket-nonce in
+		// one block produce distinct envelope hashes (frontier indexes by hash).
+		let transaction = Transaction::EIP1559(EIP1559Transaction {
+			chain_id,
+			nonce: U256::from(nonce),
+			max_priority_fee_per_gas: U256::zero(),
+			max_fee_per_gas: U256::zero(),
+			gas_limit: U256::zero(),
+			action: TransactionAction::Call(SENTINEL_ADDRESS),
+			value: U256::from(group_index),
```
dmoka, line 182: it will produce synth txs with value 1,2,3,etc wei-s which might confuse indexers tracking native token flow?!

maybe `input` field is better to set uniqeness

pallets/synthetic-logs/src/lib.rs
mrq1911: Thats gas of the synthetic tx, which represents extrinsic which already paid weight.

pallets/synthetic-logs/src/lib.rs
mrq1911: I will rebench it, but there shouldn’t be any PoV  used.

pallets/synthetic-logs/src/lib.rs
mrq1911: It shouldn’t ever panic with those constants, but i will change it.

runtime/hydradx/src/evm/erc20_logs.rs
mrq1911: Nice, will fix

pallets/synthetic-logs/src/lib.rs
mrq1911: Good point

runtime/hydradx/src/evm/erc20_logs.rs
```diff
@@ -0,0 +1,214 @@
...
+		SyntheticLogs::<Runtime>::push(address, log);
+	}
+}
+
+impl OnTransfer<AccountId, AssetId, Balance> for EmitErc20TransferLog {
+	fn on_transfer(asset: AssetId, from: &AccountId, to: &AccountId, amount: Balance) -> DispatchResult {
+		if amount == 0 {
+			return Ok(());
+		}
+		push_transfer_log(asset, evm_address_of(from), evm_address_of(to), amount);
```
enthusiastmartin, line 47: just move the check of amount 0 into the push_transfer_log

there wont be necessary to check it everywhere before calling it.

mrq1911:
closed in favour of #1471

PR #1437, opened 2026-05-04
feat(gigahdx)!: implement GIGAHDX liquid staking token for HDX

### LINKED ISSUE #1362: GIGAHDX
## 1. Overview

GIGAHDX is a Liquid Staking Token (LST) for HDX designed to:

1. Provide automatic value accrual — holders benefit from appreciating exchange rate simply by holding GIGAHDX, with no manual claiming required
2. Enable HDX to be used as collateral in the Money Market while continuing to accrue value
3. Maintain voting capabilities for governance participation
4. Offer additional rewards for active governance participants through referenda voting incentives

---

## 2. Problem Statement

- Current HDX staking has low APR with diminishing incentive for long-term stakers
- Users at 100% claimable rewards have little reason to continue high-conviction participation
- Strong user demand to use HDX as collateral without missing DeFi opportunities
- Need to better align rewards with commitment level

---

## 3. Configurable Parameters

**Key feature:** After giga-unstake, there is a cooldown period (~222 days) before HDX can be claimed. This is a major differentiator from classic HDX staking which has no cooldown.

| Parameter                     | Description                                                           | Example/Suggested Value |
| ----------------------------- | --------------------------------------------------------------------- | ----------------------- |
| **pallet-gigahdx**            |                                                                       |                         |
| Cooldown period               | Lock duration after giga-unstake before HDX can be claimed            | ~222 days               |
| Min stake amount              | Minimum HDX to stake / minimum to hold after partial unstake          | 1,000                     |
| stHDX asset ID                | Asset ID for stHDX token                                              | 670                     |
| HDX asset ID                  | Asset ID for HDX token                                                | 0                       |
| GIGAHDX asset ID              | Asset ID for GIGAHDX token (MM aToken)                                | 67                     |
| **pallet-fee-processor**      |                                                                       |                         |
| HDX Staking Pot %             | Percentage of fees to classic HDX stakers                             | ~20%                    |
| stHDX Pot %                   | Percentage of fees to GIGAHDX value accrual                           | ~50%                    |
| GigaReward Pot %              | Percentage of fees to voting rewards                                  | ~30%                    |
| Referenda seeding % per track | How much of GigaReward pot seeds each referenda (per track)           | TBD                     |
| **Money Market**              |                                                                       |                         |
| LTV (Loan-to-Value)           | Maximum borrow ratio                                                  | ~40%                    |
| LT (Liquidation Threshold)    | Threshold for liquidation                                             | ~70%                    |
| Debt ceiling                  | Maximum total debt for GIGAHDX collateral                             | 1,000,000                     |
| Oracle EMA period             | EMA period for price oracle                                           | 1 day                     |

---

## 4. Architecture

Two new pallets will be introduced:

| Pallet | Responsibility |
|--------|----------------|
| `pallet-gigahdx` | GIGAHDX token mechanics, staking, unstaking, migration, reward claiming for governance participation |
| `pallet-fee-processor` | Fee accumulation, HDX buyback via omnipool, dynamic pot distribution |

Classic HDX staking remains available alongside GIGAHDX.

---

### 5.1 Tokens Overview

| Token | Description | Visibility | Requirements |
|-------|-------------|------------|--------------|
| HDX | Native token | Visible | — |
| stHDX | Intermediate staking token (received when staking HDX) | Mostly hidden from users | Transferable, non-tradable (no market/exchange listing) |
| GIGAHDX | Money Market token (received when supplying stHDX) | Visible | Transferable (unless locked), non-tradable, usable as collateral in AAVE v3 MM, usable for governance voting (with locking mechanism) |

Users interact with HDX and GIGAHDX on the frontend. stHDX is an intermediate token that exists for technical reasons but is abstracted away from the user experience.

**Note:** stHDX and GIGAHDX are interchangeable 1:1. The issuance of stHDX and GIGAHDX is always the same — supplying 100 stHDX to MM yields exactly 100 GIGAHDX, and withdrawing 100 GIGAHDX returns exactly 100 stHDX.

**Note on Value Accrual:** In isolation mode, users supply stHDX as collateral to borrow HOLLAR — nobody borrows stHDX itself. This means GIGAHDX rebasing from borrow interest is effectively zero. All value accrual comes from the stHDX pot (fee buybacks), not from Money Market interest.

### 5.2 GIGAHDX Token Implementation

GIGAHDX (the Money Market token) requires a custom ERC20 implementation that combines:

1. **AToken functionality** — standard AAVE aToken behavior for Money Market integration
2. **Locking mechanism** — ability to lock tokens for a specified period (for governance voting)

#### 5.2.1 Locking Mechanism Requirements

- Users can lock GIGAHDX for voting in governance
- Locked GIGAHDX cannot be transferred directly
- GIGAHDX can be withdrawn from MM via standard MM withdrawal, however locked GIGAHDX cannot be withdrawn
- The only way to withdraw locked GIGAHDX is via giga-unstake, which first removes any existing locks, then allows standard MM withdrawal to proceed
- Must track lock duration per account
- Lock should be compatible with conviction voting periods
- Must support forced lock removal for liquidation — when a position is liquidated, any existing locks must be removable so the liquidated amount can be transferred
- Must support forced lock removal on partial amounts — when user partially unstakes, the lock is removed only from the unstaked portion while preserving the lock on the remaining balance

### 5.3 Token Flow

From the user's perspective, the flow is simplified:

**Flow:**
```
1. Gigastake:    HDX → GIGAHDX
2. Borrow:       HOLLAR against GIGAHDX
3. Giga-unstake: GIGAHDX → HDX (after cooldown period)
```

Internally, gigastake handles: HDX → stHDX → supply to MM → GIGAHDX
Internally, giga-unstake handles: remove locks → withdraw from MM → stHDX → HDX

**Pros:**
- Standard AAVE v3 patterns preserved (no MM modification)
- Clean separation: stHDX mechanics vs AAVE mechanics
- User sees simplified view: HDX and GIGAHDX only

**Cons:**
- Edge case: user could theoretically end up holding only stHDX (which is hidden on frontend), however by design this should not be possible since gigastake always wraps to GIGAHDX and giga-unstake handles the full unwrap

---

## 6. GIGAHDX Value Mechanism

**Note:** This is technically the stHDX value mechanism, as stHDX is the underlying vault token that appreciates. Since stHDX and GIGAHDX are 1:1 interchangeable, the value accrual applies equally to GIGAHDX holders.

GIGAHDX operates as a vault token with appreciating exchange rate.

### 6.1 Exchange Rate

At launch, the ratio is 1:1. As rewards accrue to the pot, the HDX-per-GIGAHDX ratio increases. Later entrants receive fewer GIGAHDX per HDX deposited.

**Formulas:**

```
Exchange rate (HDX per GIGAHDX) = Total HDX in gigapot / Total stHDX supply

GIGAHDX received when staking = HDX deposited / Exchange rate

HDX received when unstaking = GIGAHDX burned × Exchange rate
```

**Note:** The exchange rate calculation uses stHDX supply internally (since stHDX and GIGAHDX are 1:1 interchangeable).

### 6.2 Example

1. User 1 deposits 1000 HDX → receives 1000 GIGAHDX
2. User 2 deposits 500 HDX → receives 500 GIGAHDX
3. 100 HDX accrues from fees into pot
4. Total pot: 1600 HDX, Total supply: 1500 GIGAHDX
5. Exchange rate: 1.067 HDX per GIGAHDX
6. User 3 deposits 500 HDX → receives 469 GIGAHDX
7. If User 3 withdraws immediately: receives 500 HDX (no profit from others' rewards)

---

#### 7.1.1 Gigastake

Converts HDX to GIGAHDX:
1. Transfers HDX from user to special account (gigapot)
2. Mints stHDX based on current stHDX:HDX exchange rate
3. Immediately supplies stHDX to Money Market as collateral
4. User receives GIGAHDX

User earns staking rewards as long as they hold GIGAHDX. The intermediate stHDX token is not visible to the user.

#### 7.1.2 Giga-unstake

Converts GIGAHDX to HDX:
1. Removes any existing governance locks on user's GIGAHDX (if present)
2. Withdraws from Money Market — user's GIGAHDX is returned as stHDX
3. Burns stHDX from user
4. Calculates HDX amount based on current exchange rate
5. Creates pending unlock position with cooldown period

**Cooldown calculation:**
```
Actual cooldown = max(Fixed cooldown period, Current GIGAHDX governance lock remaining)
```
This ensures users cannot bypass governance lock commitments by unstaking.

**Note:** Giga-unstake is the only way to withdraw locked GIGAHDX. Standard MM withdrawal will fail if GIGAHDX is locked. By removing locks first, giga-unstake enables the MM withdrawal to proceed.

Notes:
- Allows partial unstake, but must respect minimum staked amount
- User can have multiple concurrent unlock positions
- If user partially unstakes a locked amount, the lock must be preserved on the remaining GIGAHDX balance

#### 7.1.3 Claim Unlock

Claims HDX after cooldown period expires.

#### 7.1.4 Migrate

One-time migration from existing staked HDX to GIGAHDX directly:
1. Converts all staked HDX plus all unclaimed rewards (regardless of claim status)
2. Mints stHDX at current exchange rate
3. Immediately supplies stHDX to Money Market
4. User receives GIGAHDX

#### 7.1.5 Claim Rewards

Claims rewards from governance participation (referenda voting).

**Flow:**
1. Retrieves all pending reward entries for the user (recorded when votes were removed)
2. For each reward entry:
   - Transfers HDX from the referenda's reward pot
   - Performs internal "gigastake" — converts HDX to GIGAHDX
3. User receives GIGAHDX as reward
4. Claimed reward entries are cleared from storage

**Note:** Rewards become claimable only after the user removes their vote from a finished referenda (Approved or Rejected). See Section 10 for details on how rewards are calculated and recorded.

### 7.2 Configuration

| Parameter | Description |
|-----------|-------------|
| Cooldown period | Lock duration in blocks after giga-unstake |
| Min stake amount | Minimum HDX to stake / minimum GIGAHDX to hold after partial unstake |
| stHDX asset ID | Asset ID for stHDX token |
| HDX asset ID | Asset ID for HDX token |
| GIGAHDX asset ID | Asset ID for GIGAHDX token (Money Market aToken) |

---

### 8.1 Purpose

Consolidates fee handling currently spread across referral pallet:
1. Accumulate trading fees (various assets)
2. Buyback HDX via omnipool
3. Distribute to registered pots based on configured percentages

### 8.2 Note on Refactoring

This functionality is currently handled by `pallet-referrals`:
- Fee accumulation
- HDX buybacks (in `on_idle` hook)
- Distribution to staking pot (fixed, single pot)

This logic needs to be extracted from `pallet-referrals` into the new `pallet-fee-processor`. The referrals pallet should be refactored to delegate fee processing to this new pallet, retaining only referral-specific logic.

### 8.3 Pot System

Pots can be registered and updated dynamically via governance. Initial pots:

| Pot | Description |
|-----|-------------|
| HDX Staking Pot | Classic HDX stakers (existing) |
| stHDX Pot | Auto-compounds into stHDX value |
| stHDX Reward Pot | Active participation rewards (e.g., voting incentives) |
| Referral Pot | Referrals rewards |

### 8.4 Buyback Mechanism

- Runs in `on_idle` hook
- Converts accumulated fee assets → HDX via omnipool
- Distributes HDX to pots based on configured percentages
- Available for any component that needs asset exchange (not just GIGAHDX)

### 8.5 Referral changes

Referrals pallet has a system of tier, where each tier has configured percentages for trader,referrer and external.
External account is currently used for legacy staking.
With the refactored approach, referrals pallet is no longer responsible for fee processing, nor is for buyback. It simply receives a given portion of the trade fee which it can divided between trader, referrer and external account.
As legacy staking will have a dedicated fee receiver registered in the fee processing module, the external account will be changed to the Treasury account.

Example of TIER 1.
Currently
	Level::Tier1 => FeeDistribution {
				referrer: Permill::from_percent(6),
				trader: Permill::from_percent(4),
				external: Permill::from_percent(40),
			},

This will become  ( note that the percentage are only for demonstrating purposes )
	Level::Tier1 => FeeDistribution {
				referrer: Permill::from_percent(30),
				trader: Permill::from_percent(20),
				external: Permill::from_percent(50),
			},

This means, that given portion of the trade fee amount will be divided: 30% to referrer , 20% to trader and 50% goes to treasury.

---

### 9.1 Overview

Voting using `pallet-conviction-voting` is preserved. No custom changes to the conviction voting pallet are required.

**Key points:**
- GIGAHDX votes support **user-selected conviction** — users can choose their conviction level (1x to 6x) just like with HDX voting
- Higher conviction = longer lock period = higher governance rewards (see Section 10)
- Governance locks and staking cooldown are **combined** — giga-unstake cooldown is `max(fixed cooldown period, current governance lock)`

### 9.2 Current Implementation

The conviction voting pallet has a `Currency` config option:

```rust
type Currency: ReservableCurrency<Self::AccountId>
    + LockableCurrency<Self::AccountId, Moment = BlockNumberFor<Self, I>>
    + fungible::Inspect<Self::AccountId>;
```

Currently, this is configured to use an implementation from the balances pallet, which supports voting with the native HDX token only.

### 9.3 Proposed Approach: Custom Adapter

**Note:** This is a suggested approach. Further detailed analysis during implementation may reveal additional complexities or alternative solutions. The feasibility depends on whether all required trait methods can be properly delegated and coordinated between the two underlying implementations.

To extend voting support to HDX and GIGAHDX, we will create a special adapter that:

1. **Implements required traits** — `ReservableCurrency`, `LockableCurrency`, `fungible::Inspect`
2. **Returns combined balance** — When queried for user balance, returns sum of GIGAHDX + HDX
3. **Handles locking with priority** — When a lock is required:
   - First, consume GIGAHDX balance
   - Then, use HDX for the remainder (if any)

#### set_lock

When `set_lock` is called with amount X:
1. Determine how much GIGAHDX user has
2. Lock GIGAHDX first (call `set_lock` on GIGAHDX custom aToken implementation)
3. If X > GIGAHDX balance, lock remainder from HDX (call `set_lock` on balances pallet)

#### extend_lock / remove_lock

Same approach applies:
- Adapter coordinates between GIGAHDX token (custom aToken with locking support) and HDX (balances pallet)
- Operations are split proportionally based on what was originally locked

### 9.5 Example

User has 1000 GIGAHDX + 2000 HDX, votes with 2500:
- Adapter returns total balance: 3000
- Lock requested: 2500
- GIGAHDX locked: 1000 (full balance)
- HDX locked: 1500 (remainder)

### 9.6 Open Issue: Liquidation vs Voting Lock Conflict

There is a potential conflict when GIGAHDX is both:
- Used as collateral in the Money Market
- Locked for voting

**Scenario:**
1. User has GIGAHDX as collateral and borrows HOLLAR
2. User votes with GIGAHDX, locking it for a conviction voting period
3. User's position becomes undercollateralized (e.g., HDX price drops)
4. Liquidation is triggered — protocol attempts to seize GIGAHDX
5. **Problem:** The GIGAHDX is locked for voting and cannot be transferred

**Proposed solution:**
When liquidating a GIGAHDX position, any existing voting locks are removed before the liquidation proceeds. This will require custom liquidation logic specifically for the GIGAHDX asset that:
1. Checks for existing voting locks on the position - force removes any votes ( even a vote in ongoing referenda
2. Removes/clears the locks
3. Proceeds with standard liquidation flow

### 9.7 Voting Lock vs Giga-unstake

**Scenario:**
1. User has GIGAHDX
2. User votes with conviction, GIGAHDX gets locked for conviction voting period
3. User wants to giga-unstake (withdraw)

**Solution:**
The giga-unstake cooldown period is calculated as:
```
Actual cooldown = max(Fixed cooldown period, Current GIGAHDX governance lock remaining)
```

This means:
- If governance lock > fixed cooldown → cooldown extends to match governance lock
- If governance lock < fixed cooldown → standard fixed cooldown applies
- If no governance lock → standard fixed cooldown applies

**Example:**
- Fixed cooldown period: 222 days
- User voted with 6x conviction (lock: 192 days remaining)
- Actual cooldown: max(222, 192) = 222 days

**Example 2:**
- Fixed cooldown period: 222 days
- User voted with 6x conviction (lock: 250 days remaining)
- Actual cooldown: max(222, 250) = 250 days

**question:**
What happens if a referenda has not finished yet and the user has voted but wants to withdraw? 

This wont be allowed - user cannot have a vote in ongoing referenda if user wants to unstake. 

---

## 10. Referenda Participation Rewards

This section describes how active Referenda participants will be rewarded.

**Important:** 
- Rewards are distributed in **GIGAHDX**, not HDX
- Only votes made with **GIGAHDX** are eligible for rewards — votes made with HDX are not rewarded

### 10.1 Overview

Rewards for referenda participation are set up **after** a referenda finishes (Approved or Rejected) using a lazy approach. No temporary pots are created during the referenda.

**Note:** Cancelled referenda do not receive rewards — only Approved or Rejected referenda are eligible.

### 10.2 Vote Tracking

To determine reward eligibility, `pallet-gigahdx` must track GIGAHDX votes for each running referenda:
- When a user votes with GIGAHDX, record the vote (user, referenda ID, GIGAHDX amount, conviction)
- This tracking is separate from the conviction voting pallet's own tracking
- Only GIGAHDX votes are tracked — HDX votes are not recorded

### 10.3 Lazy Reward Allocation

Rewards are allocated lazily — triggered by the first vote removal after a referenda finishes (Approved or Rejected only).

**When the first vote is removed from a finished referenda:**
1. Take a configured portion from the GigaReward Pot
2. Transfer this amount to a temporary holding account for the specific referenda
3. Store the total reward amount and total GIGAHDX amount voted for this referenda in `pallet-gigahdx`

**On subsequent vote removals:**
- The reward pool for the referenda is already allocated
- No additional allocation from GigaReward Pot

**Cancelled referenda:**
- No rewards are allocated
- Users simply remove their votes without receiving any rewards

### 10.4 Reward Calculation on Vote Removal

When a user removes their vote from a finished referenda (Approved or Rejected):

1. Calculate user's share of rewards based on GIGAHDX amount voted **and conviction**:
```
User weighted vote = User GIGAHDX amount × User conviction multiplier
Total weighted votes = Sum of (GIGAHDX amount × conviction) for all voters

User reward = (User weighted vote / Total weighted votes) × Referenda reward pool
```

2. **Record** the calculated reward for the user (not transferred yet)
   - Store: user account, reward amount, referenda reward pot reference

**Example:**
- Referenda reward pool: 1000 HDX
- User A voted with 100 GIGAHDX at 1x conviction → weighted: 100
- User B voted with 100 GIGAHDX at 3x conviction → weighted: 300
- User C voted with 100 GIGAHDX at 6x conviction → weighted: 600
- Total weighted votes: 1000

Calculated rewards (recorded, not transferred):
- User A: (100 / 1000) × 1000 = 100 HDX
- User B: (300 / 1000) × 1000 = 300 HDX
- User C: (600 / 1000) × 1000 = 600 HDX

**Note:** Higher conviction = longer lock = higher reward share. Users must eventually remove their votes to have rewards calculated. This is enforced by the voting system itself — locked tokens cannot be used until votes are removed.

### 10.5 Claiming Rewards

To receive rewards, users must call `claim_rewards` from `pallet-gigahdx` (see Section 7.1.5).

**Claim flow:**
1. User calls `claim_rewards`
2. System retrieves all recorded rewards for the user (from various referenda)
3. For each reward entry:
   - Transfer HDX from the referenda's reward pot to user
   - Perform internal "gigastake" — convert HDX to GIGAHDX
4. User receives GIGAHDX as reward
5. Claimed reward entries are cleared from storage

### 10.6 Referenda Rewards Storage

The following information is stored in `pallet-gigahdx`:

**Per running referenda:**
- List of GIGAHDX voters (user, GIGAHDX amount voted, conviction)
- Total weighted votes (sum of amount × conviction)

**Per finished referenda (after first vote removal):**
- Total reward pool allocated
- Total weighted votes (snapshot at allocation time)
- Remaining reward balance
- Referenda reward pot account reference

**Per user (claimable rewards):**
- List of pending rewards: (referenda ID, reward amount, referenda reward pot reference)

### 10.7 Configuration

| Parameter | Description |
|-----------|-------------|
| Referenda reward percentage | Percentage of GigaReward Pot to allocate per finished referenda |

**Note:** The reward percentage may be configurable per track (e.g., root referenda get larger rewards than small treasury proposals).

---

### 11.1 Price Oracle

GIGAHDX price in the Money Market is determined as:

```
GIGAHDX/USD = EMA oracle of HDX/USD × GIGAHDX/HDX ratio
```

### 11.2 Isolation Mode

GIGAHDX will be configured as isolated collateral in AAVE v3:

| Parameter | Value |
|-----------|-------|
| Collateral | stHDX |
| Borrowable assets | HOLLAR only |
| LTV (Loan-to-Value) | ~40% (conservative) |
| LT (Liquidation Threshold) | ~50% |
| Debt ceiling | Start low, increase over time |

### 11.3 Liquidation

GIGAHDX liquidation uses a **Treasury-only** mechanism. External liquidators are blocked to prevent manipulation attacks (e.g., manipulating HDX price to liquidate stakers and extract HOLLAR).

**Why Treasury-only:**
- External liquidators would need to wait ~222 days to convert liquidated GIGAHDX to HDX
- Prevents price manipulation attacks targeting GIGAHDX holders
- Treasury can safely absorb the debt and collateral

**On-chain Liquidation Flow:**
1. Protocol flashmints HOLLAR to repay the user's debt
2. Gov locks are removed on the liquidated account
3. Liquidates position, receives GIGAHDX (doesn't unwrap)
4. Received GIGAHDX is transferred to dedicated treasury account
5. All needed HOLLAR is borrowed from regular treasury account against other collateral
6. Flashmint is repaid

**Note:** The simpler approach is to not use the liquidated GIGAHDX as treasury collateral at all — just borrow all HOLLAR against other treasury collateral. This is easier to calculate and execute, though it lowers debt ceiling somewhat.

**Debt Ceiling:** Should be set based on what treasury can safely cover in case of liquidations.

**Locked GIGAHDX:**
If the GIGAHDX being liquidated has existing governance voting locks, the liquidation process removes these locks before transferring the collateral (step 2 above).

See also Section 9.6 for discussion of the liquidation vs voting lock conflict scenario.

### 11.4 Future Features (Out of Scope for Initial Launch)

| Feature | Description |
|---------|-------------|
| Immortal loans | Pay charge after liquidation to recover position |
| Reduced HOLLAR borrow APY | Preferential rate for GIGAHDX holders |
| Self-repaying loans | HOLLAR revenue distributed as loan repayment |

---

## 12. Migration

Existing staked HDX holders can migrate to GIGAHDX directly:

1. All staked HDX plus all unclaimed rewards converted at current exchange rate
2. stHDX is minted and immediately supplied to MM
3. User receives GIGAHDX
4. One-time opt-in per account

---

## 13. Parallel Workstreams

The following workstreams can be developed:

| # | Workstream | Can Parallel? |
|---|------------|---------------|
| 1 | GIGAHDX token implementation (custom ERC20 + aToken + locking mechanism) | Yes |
| 2 | `pallet-gigahdx` | Yes |
| 3 | Referrals refactor + `pallet-fee-processor` | Yes |
| 4 | Voting integration + active votes rewards | Partially |
| 5 | Treasury-only liquidation mechanism | Yes |

Workstreams 1, 2, 3, and 5 can be developed fully in parallel. Workstream 4 can be started in parallel but will need integration with the other components.

khuzama98:
## Description

Implements GIGAHDX, a Liquid Staking Token (LST) for HDX that enables automatic
value accrual, Money Market collateral usage, and preserved governance participation
within a single staking position.

Three new pallets:
- `pallet-gigahdx` — stake/unstake/migrate/claim mechanics, exchange rate, cooldown
- `pallet-gigahdx-voting` — conviction voting adapter, governance reward distribution
- `pallet-fee-processor` — trade fee accumulation, HDX buyback via Omnipool, pot distribution

Supporting changes:
- Referrals fee routing refactored through `pallet-fee-processor`
- AAVE money market: GIGAHDX isolation-mode instance with HOLLAR-only borrowing and treasury liquidation
- EVM lock manager precompile for AAVE aToken ↔ GIGAHDX lock coordination
- External oracle: source registration moved to storage with runtime migration
- Collator rotation added; node switched to slot-based collation
- Polkadot SDK bumped to `polkadot-stable2506-10-patch`
- Cooldown period set to 100 days; undeciding timeout reduced from 14 to 2 days

## Related Issue

Closes: #1362

## Motivation and Context

Classic HDX staking requires manual reward claiming and cannot be used as Money
Market collateral. GIGAHDX solves both:

- **Automatic appreciation** — exchange rate rises passively as trade fees are bought
  back into the gigapot; no manual claiming needed
- **Collateral-ready** — GIGAHDX is accepted in the AAVE isolation-mode instance
  with HOLLAR as the sole borrowable asset
- **Governance preserved** — conviction voting (None–6x) is maintained via a voting
  adapter; unstake cooldown is unioned with governance locks

## How Has This Been Tested?

- Unit tests across both new pallets: stake/unstake (free + voted split), partial
  unstake, reward calculation, voting adapter, exchange rate, fee processing
- Integration tests against a Chopsticks snapshot: full stake/unstake flows,
  liquidation e2e, collator rotation, relay parent offset
- Regression tests for staking, referrals, oracle, and conviction voting verified passing

## Checklist:


integration-tests/snapshots/gigahdx/gigahdx2
dmoka, line 1: i think this is not used, so we can remove it

integration-tests/snapshots/gigahdx/gigahdx5_slim
dmoka, line 1: AFAIK, the latest greatest snapshot (created 29th April) is `gigahdx_slim_2lark`, could you please try to use that in files:

integration-tests/src/gigahdx_vote_scenarios.rs:39 

and 

integration-tests/src/gigahdx_vote_scenarios_extra.rs:35

So we will end up using only only one, and this gigahdx5_slim can be removed.

pallets/liquidation/src/lib.rs
```diff
@@ -139,18 +141,44 @@ pub mod pallet {
...
 
-	/// Borrowing market contract address
+	/// Borrowing market contract address (main Hydration Market pool)
 	#[pallet::storage]
 	#[pallet::getter(fn borrowing_contract)]
 	pub type BorrowingContract<T: Config> = StorageValue<_, EvmAddress, ValueQuery, DefaultBorrowingContract>;
 
+	#[pallet::type_value]
+	pub fn DefaultGigaHdxPoolContract() -> EvmAddress {
+		EvmAddress::from_slice(hex_literal::hex!("1b02E051683b5cfaC5929C25E84adb26ECf87B38").as_slice())
```
dmoka, line 174: what is the reason that we use the same address as default, like for DefaultBorrowingContract?

Since they are 2 separate AAVEs, i expect a different default contract for giga

pallets/liquidation/src/lib.rs
```diff
@@ -437,6 +490,92 @@ impl<T: Config> Pallet<T> {
...
+		if borrow_result.exit_reason != ExitReason::Succeed(ExitSucceed::Returned) {
+			log::info!(target: "liquidation",
+				"Treasury borrow failed. Reason: {:?}", borrow_result.value);
+			return Err(Error::<T>::BorrowFailed.into());
+		}
+
+		// Step 3: Treasury calls liquidationCall with receive_atoken=true
+		let treasury_gigahdx_before = <T as Config>::Currency::balance(collateral_asset, &treasury_evm_account);
+
+		//TODO: if verified, create a pallet constnat for this or user get_underlying asset from AaveTradeExecuto
```
dmoka, line 526: sthdx asset id is hadcoded on line 529, please get it from config (or we might already have a config for it somewhere else?)

pallets/fee-processor/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		/// Distribute HDX to all receiver pots according to their percentages.
+		fn distribute_to_pots(
+			source: &T::AccountId,
+			total: Balance,
+			destinations: Vec<(T::AccountId, Permill)>,
+		) -> DispatchResult {
+			for (dest, percentage) in destinations {
+				//TODO: why not used balance here ?
```
dmoka, line 279: todo

pallets/fee-processor/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+				.ref_time()
+				.checked_div(convert_weight.ref_time())
+				.unwrap_or(0)
+				.min(T::MaxConversionsPerBlock::get() as u64);
+
+			for asset_id in PendingConversions::<T>::iter_keys().take(max_conversions as usize) {
+				match Self::do_convert(asset_id) {
+					Ok(_) => {}
+					Err(e) => {
+						Self::deposit_event(Event::ConversionFailed { asset_id, reason: e });
```
dmoka, line 151: the thing is that if do_convert fails, for example in the end when we use `distribute_to_pots`, then we will leave some orphaned hdx in the pot as we don't roll back. 

Wondering if wrapping this in a transactional call would make sense or not, WDYT?

pallets/gigahdx/gigahdx-bugs-found.md
```diff
@@ -0,0 +1,75 @@
+# GigaHDX - Bugs Found During Testing
```
dmoka, line 1: please double check if all these bugs are resolved, then if so, we can remove this file

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,386 @@
...
+			1 => Permill::from_percent(8),  // whitelisted_caller
+			5 => Permill::from_percent(5),  // treasurer
+			_ => Permill::from_percent(3),  // default
+		}
+	}
+}
+
+// ---------------------------------------------------------------------------
+// pallet-gigahdx Config
+// ---------------------------------------------------------------------------
```
dmoka, line 186:
```suggestion
```

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,386 @@
...
+	type ForceRemoveVote = RuntimeForceRemoveVote;
+	type GigaRewardPotId = GigaRewardPotId;
+	type VoteLockingPeriod = VoteLockingPeriod;
+	type MaxVotes = GigaHdxMaxVotes;
+	type VotingWeightInfo = ();
+}
+
+// ---------------------------------------------------------------------------
+// pallet-fee-processor — FeeReceiver impls and Config
+// ---------------------------------------------------------------------------
```
dmoka, line 220:
```suggestion
```

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,386 @@
...
+	type MoneyMarket = AaveMoneyMarket;
+	type Hooks = GigaHdxVoting;
+	type PalletId = GigaHdxPalletId;
+	type HdxAssetId = NativeAssetId;
+	type StHdxAssetId = StHdxAssetId;
+	type GigaHdxAssetId = GigaHdxAssetIdConst;
+	type CooldownPeriod = GigaHdxCooldownPeriod;
+	type MinStake = GigaHdxMinStake;
+	type MaxUnstakePositions = GigaHdxMaxUnstakePositions;
+	type WeightInfo = ();
```
dmoka, line 200: this gotta be set

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,386 @@
...
+
+impl pallet_gigahdx_voting::Config for Runtime {
+	type NativeCurrency = Balances;
+	type Referenda = RuntimeReferendumInfo;
+	type TrackRewards = RuntimeTrackRewards;
+	type ForceRemoveVote = RuntimeForceRemoveVote;
+	type GigaRewardPotId = GigaRewardPotId;
+	type VoteLockingPeriod = VoteLockingPeriod;
+	type MaxVotes = GigaHdxMaxVotes;
+	type VotingWeightInfo = ();
```
dmoka, line 215: this gotta be set

pallets/gigahdx/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+				}
+				Ok(())
+			})
+		}
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Public functions (called by pallet-gigahdx-voting and runtime adapters)
+// ---------------------------------------------------------------------------
```
dmoka, line 431:
```suggestion
```

traits/src/gigahdx.rs
```diff
@@ -0,0 +1,318 @@
...
+			+ sp_arithmetic::traits::UniqueSaturatedInto<u32>
+			+ From<u32>,
+	> FeeReceiver<AccountId, Balance> for Tuple
+{
+	for_tuples!( where #(Tuple: FeeReceiver<AccountId, Balance, Error=DispatchError>)* );
+	type Error = DispatchError;
+
+	fn destination() -> AccountId {
+		// Not meaningful for tuple — use destinations() instead.
+		panic!("destination() should not be called on tuple")
```
dmoka, line 267: i think we should not panic in such.

we can surely add a debug_assert.

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,386 @@
...
+
+	fn withdraw(who: &AccountId, underlying_asset: AssetId, amount: Balance) -> Result<Balance, DispatchError> {
+		let _ = pallet_evm_accounts::Pallet::<Runtime>::bind_evm_address(RuntimeOrigin::signed(who.clone()));
+
+		let asset_evm = HydraErc20Mapping::asset_address(underlying_asset);
+		let pool = GigaHdxPoolContract::<Runtime>::get();
+
+		Aave::do_withdraw(pool, who, asset_evm, amount)?;
+
+		Ok(amount)
```
dmoka, line 68: here we return the amount passed as param but i wonder what happens we wihtdraw 1 less due to scaling index/rounding behavious we had in the past.

Maybe here we could calculate a balance delta and return that. And if that it not equal to passed amount we could either return error or/and do debug assert.

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,386 @@
...
+
+// ---------------------------------------------------------------------------
+// AaveMoneyMarket — supply/withdraw stHDX via real AAVE Pool contract
+// ---------------------------------------------------------------------------
+
+pub struct AaveMoneyMarket;
+
+impl hydradx_traits::gigahdx::MoneyMarketOperations<AccountId, AssetId, Balance> for AaveMoneyMarket {
+	fn supply(who: &AccountId, underlying_asset: AssetId, amount: Balance) -> Result<Balance, DispatchError> {
+		let _ = pallet_evm_accounts::Pallet::<Runtime>::bind_evm_address(RuntimeOrigin::signed(who.clone()));
```
dmoka, line 41: why is the error swallowed?

If there is a specific reason, I would comment that, otherwise propagating it to the caller.

The same hapens in withdraw function, below this.

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,386 @@
...
+// ---------------------------------------------------------------------------
+// RuntimeTrackRewards — per-track reward percentage
+// ---------------------------------------------------------------------------
+
+pub struct RuntimeTrackRewards;
+
+impl hydradx_traits::gigahdx::TrackRewardConfig for RuntimeTrackRewards {
+	fn reward_percentage(track_id: u16) -> Permill {
+		match track_id {
+			0 => Permill::from_percent(10), // root
```
dmoka, line 176: is there some constants somewhere for these track ids 0, 1, 5? otherwise they feel a bit like magic number

pallets/gigahdx-voting/src/lib.rs
```diff
@@ -0,0 +1,694 @@
...
+		Ok(())
+	}
+
+	fn additional_unstake_lock(who: &T::AccountId) -> BlockNumberFor<T> {
+		let current_block = frame_system::Pallet::<T>::block_number();
+		let mut max_remaining: BlockNumberFor<T> = Zero::zero();
+
+		for (_, vote) in GigaHdxVotes::<T>::iter_prefix(who) {
+			if vote.lock_expires_at > current_block {
+				let remaining = vote.lock_expires_at - current_block;
```
dmoka, line 669: i would use checked math

pallets/gigahdx-voting/src/lib.rs
```diff
@@ -0,0 +1,694 @@
...
+				}
+			}
+		}
+
+		// Priors carry the conviction window after `remove_vote`, so they
+		// must contribute to the unstake cooldown the same way active votes do.
+		for (_class, mut prior) in PriorLockSplit::<T>::iter_prefix(who) {
+			prior.rejig(current_block);
+			if prior.is_active() && prior.until > current_block {
+				let remaining = prior.until - current_block;
```
dmoka, line 681: i would use checked math

pallets/gigahdx-voting/src/adapter.rs
```diff
@@ -0,0 +1,563 @@
...
+	fn total_balance(who: &T::AccountId) -> Balance {
+		gigahdx_balance::<T>(who).saturating_add(hdx_balance::<T>(who))
+	}
+
+	fn balance(who: &T::AccountId) -> Balance {
+		<Self as fungible::Inspect<T::AccountId>>::total_balance(who)
+	}
+
+	fn reducible_balance(who: &T::AccountId, _preservation: Preservation, _force: Fortitude) -> Balance {
+		<Self as fungible::Inspect<T::AccountId>>::total_balance(who)
```
dmoka, line 231: are we sure that the reducible balance is equal to the total balance?

pallets/gigahdx/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+
+			// Notify hooks.
+			T::Hooks::on_stake(&who, hdx_amount, gigahdx_received)?;
+
+			Self::deposit_event(Event::Staked {
+				who,
+				hdx_amount,
+				st_hdx_minted: st_hdx_amount,
+				gigahdx_received,
+				exchange_rate: Self::exchange_rate(),
```
dmoka, line 246: this is the post stake exchange rate. This is what we/indexer want or the exchange rate happened before stake?

pallets/gigahdx/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+	/// so the recipient receives the full value of their reward.
+	pub fn stake_rewards(who: &T::AccountId, hdx_amount: Balance) -> Result<Balance, DispatchError> {
+		let total_st_hdx = Self::total_st_hdx_supply();
+		let total_hdx = Self::total_hdx();
+
+		// Exclude the reward HDX from total to get the pre-reward rate.
+		let st_hdx_amount = if total_st_hdx.is_zero() {
+			hdx_amount
+		} else {
+			let pre_reward_hdx = total_hdx.checked_sub(hdx_amount).ok_or(Error::<T>::Arithmetic)?;
```
dmoka, line 474: not sure if arithmetic error is the best way to represent this error, it is more about pot doesnt have enough hdx. Maybe a better error `InsufficientGigapotBalance` with proper desc would be better.

pallets/gigahdx-voting/src/adapter.rs
```diff
@@ -0,0 +1,563 @@
...
+
+	fn subsume(&mut self, other: Self) {
+		self.0 = self.0.saturating_add(other.0);
+	}
+
+	fn offset(self, other: Self::Opposite) -> frame_support::traits::SameOrOther<Self, Self::Opposite> {
+		use frame_support::traits::SameOrOther;
+		let (a, b) = (self.0, other.0);
+		if a >= b {
+			SameOrOther::Same(PositiveImbalance(a - b))
```
dmoka, line 127: pls use checked/saturating math

pallets/gigahdx-voting/src/adapter.rs
```diff
@@ -0,0 +1,563 @@
...
+		} else {
+			SameOrOther::Other(NegativeImbalance(b - a))
```
dmoka, line 129: pls use checked/saturating math

pallets/gigahdx-voting/src/types.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// A tracked GIGAHDX vote for a specific referendum.
+#[derive(Encode, Decode, DecodeWithMemTracking, Clone, Eq, PartialEq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct GigaHdxVote<BlockNumber> {
+	/// Combined committed amount (GIGAHDX + HDX side).
+	pub amount: Balance,
+	/// Conviction level chosen by the voter.
+	pub conviction: Conviction,
+	/// Block number when the vote was cast.
+	pub voted_at: BlockNumber,
```
dmoka, line 98: i dont see this field used anywhere, only set once, what is the purpose of it?

enthusiastmartin:
no longer relevant. it is being reworked

PR #1445, opened 2026-05-07
feat: Gigahdx 

enthusiastmartin:


runtime/hydradx/src/assets.rs
```diff
@@ -1868,6 +1872,70 @@ impl pallet_dispenser::Config for Runtime {
 	type BenchmarkHelper = DispenserBenchmarkHelper;
 }
 
+parameter_types! {
+	pub const GigaHdxLockId: frame_support::traits::LockIdentifier = *b"ghdxlock";
+	pub const GigaHdxPalletId: frame_support::PalletId = frame_support::PalletId(*b"gigahdx!");
+	// stHDX (id 670) MUST only be mintable / burnable by `pallet-gigahdx`.
```
dmoka: i dont see stHDX mintability enforced in runtime, maybe we could add a guard in FungibleCurrencies?

pallets/gigahdx/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		#[pallet::weight(T::WeightInfo::set_pool_contract())]
+		pub fn set_pool_contract(origin: OriginFor<T>, contract: EvmAddress) -> DispatchResult {
+			T::AuthorityOrigin::ensure_origin(origin)?;
+			ensure!(TotalLocked::<T>::get() == 0, Error::<T>::OutstandingStake);
+			GigaHdxPoolContract::<T>::put(contract);
+			Self::deposit_event(Event::PoolContractUpdated { contract });
+			Ok(())
+		}
+
+		/// Unstake `gigahdx_amount` of the caller's GIGAHDX. The MM burns the
```
dmoka: let's follow our convention for doc comments so:

description
params
emitted events

applicable for all other extrinsics

runtime/hydradx/src/assets.rs
```diff
@@ -1868,6 +1872,70 @@ impl pallet_dispenser::Config for Runtime {
...
+	#[cfg(not(feature = "runtime-benchmarks"))]
+	type MoneyMarket = crate::gigahdx::AaveMoneyMarket;
+	#[cfg(feature = "runtime-benchmarks")]
+	type MoneyMarket = crate::gigahdx::BenchmarkMoneyMarket;
+	type AuthorityOrigin = EnsureRoot<AccountId>;
+	type PalletId = GigaHdxPalletId;
+	type LockId = GigaHdxLockId;
+	type MinStake = GigaHdxMinStake;
+	type CooldownPeriod = GigaHdxCooldownPeriod;
+	type WeightInfo = ();
```
dmoka: weight needds to be set

pallets/gigahdx/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+			ensure!(gigahdx_amount <= stake.gigahdx, Error::<T>::InsufficientStake);
+
+			// Compute payout from PRE-unstake totals (helper reads live state, so
+			// it must run BEFORE any mint/burn/transfer below).
+			let payout = Self::calculate_hdx_amount_given_gigahdx(gigahdx_amount).map_err(|_| Error::<T>::Overflow)?;
+
+			// Pre-decrement `gigahdx` so `LockableAToken.burn`'s `freeBalance`
+			// check (via the lock-manager precompile) lets the burn through.
+			let new_gigahdx = stake.gigahdx.checked_sub(gigahdx_amount).ok_or(Error::<T>::Overflow)?;
+			Stakes::<T>::insert(
```
dmoka: is it necessary to insert StakeRecord here?

Because we do it also at line 428 too

pallets/gigahdx/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		/// Set the AAVE V3 Pool contract H160 used by the money-market adapter.
+		/// Gated by `AuthorityOrigin`. Refuses to swap the pool while users
+		/// hold outstanding stake — a swap mid-flight would route subsequent
+		/// `giga_unstake` calls to a pool that doesn't hold their atokens,
+		/// reverting the burn and leaving HDX permanently locked.
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::set_pool_contract())]
+		pub fn set_pool_contract(origin: OriginFor<T>, contract: EvmAddress) -> DispatchResult {
+			T::AuthorityOrigin::ensure_origin(origin)?;
+			ensure!(TotalLocked::<T>::get() == 0, Error::<T>::OutstandingStake);
```
dmoka: i think it would be better to check if total gigahdx is 0 or  no `Stakes` storage entry.

because there can be situation when user hdx is 0 but gigahdx is non-zero, see test `giga_unstake_should_handle_remaining_atokens_when_active_drained_by_yield`

so in this case user still holds gigahdx token issued by the current pool.

precompiles/lock-manager/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+{
+	/// Returns the locked GIGAHDX balance for `account`. Returns zero when
+	/// `token` is not the configured GIGAHDX aToken address.
+	#[precompile::public("getLockedBalance(address,address)")]
+	#[precompile::view]
+	fn get_locked_balance(handle: &mut impl PrecompileHandle, token: Address, account: Address) -> EvmResult<U256> {
+		// Blake2_128Concat key prefix (16) + AccountId (32) + StakeRecord (2 × u128 = 32) = 80 bytes
+		handle.record_db_read::<Runtime>(80)?;
+
+		if H160::from(token) != ExpectedToken::get() {
```
dmoka, line 62: small thing, but first we could do this check then the record_db_read read lone, so we optimize a bit

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,142 @@
...
+		// Pool.supply rounds the scaled balance down, so the actual aToken
+		// minted may be 1+ wei less than `amount`. Read the user's aToken
+		// balance before/after and return the delta — the pallet records that
+		// as `Stakes.gigahdx`, preserving the invariant
+		// `Stakes.gigahdx == aToken.balanceOf` that `LockableAToken.burn`'s
+		// `freeBalance = balanceOf - locked` check relies on.
+		let balance_before = Self::balance_of(who);
+
+		let supply_ctx = CallContext::new_call(pool, who_evm);
+		let mut data = Into::<u32>::into(AaveFunction::Supply).to_be_bytes().to_vec();
```
dmoka: to construct ABI call data, I would use our EvmDataWriter just like in other part of our runtime, so we follow convenction

runtime/hydradx/src/gigahdx.rs
```diff
@@ -0,0 +1,142 @@
...
+		let pool = Self::pool()?;
+
+		// Mirror the supply path — return the actual underlying received,
+		// not the requested amount, so callers can reconcile against AAVE's
+		// scaledBalance rounding. Symmetry with `supply` keeps round-trip
+		// accounting consistent across rate drift.
+		let balance_before = Self::balance_of(who);
+
+		let withdraw_ctx = CallContext::new_call(pool, who_evm);
+		let mut data = Into::<u32>::into(AaveFunction::Withdraw).to_be_bytes().to_vec();
```
dmoka: we could use EvmDataWriter here too

pallets/gigahdx/src/lib.rs
```diff
@@ -0,0 +1,559 @@
...
+
+#[cfg(test)]
+mod tests;
+
+#[cfg(feature = "runtime-benchmarks")]
+pub mod benchmarking;
+
+pub mod weights;
+
+/// Hook for benchmark setup — wires runtime-side helpers (asset registry
```
dmoka, line 46: there are a lot of unnecessary comments like these we could get rid of, pls let's clean those up

runtime/hydradx/src/assets.rs
enthusiastmartin: no need to do.

pallets/gigahdx/src/lib.rs
enthusiastmartin: yes

pallets/gigahdx-rewards/src/voting_hooks.rs
```diff
@@ -0,0 +1,215 @@
...
+use frame_support::traits::{Currency, ExistenceRequirement};
+use pallet_conviction_voting::{AccountVote, Status, VotingHooks};
+use primitives::Balance;
+use sp_std::marker::PhantomData;
+
+/// `VotingHooks` impl for this pallet. The runtime wires this through a
+/// tuple adapter alongside any other consumer (typically staking).
+pub struct VotingHooksImpl<T>(PhantomData<T>);
+
+impl<T: Config> VotingHooks<T::AccountId, ReferendumIndex, Balance> for VotingHooksImpl<T> {
```
dmoka, line 28: Since we added this, i think we should not forget to regenerate benchmark weights for pallet_conviction_voting.

pallets/gigahdx-rewards/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+			let who = ensure_signed(origin)?;
+			// Mirror `giga_stake`: a pre-existing HDX lock would shadow ours
+			// under FRAME's max-of-locks, leaving the staked HDX transferable.
+			ensure!(
+				<T as pallet_gigahdx::Config>::ExternalClaims::on(&who) == 0,
+				pallet_gigahdx::Error::<T>::BlockedByExternalLock
+			);
+
+			let total = PendingRewards::<T>::take(&who);
+			ensure!(total > 0, Error::<T>::NoPendingRewards);
```
dmoka, line 205: don't we need to check here if the total is bigger than MinStake?

Because we do the following in the beginning of giga_stake
```
ensure!(amount >= T::MinStake::get(), Error::<T>::BelowMinStake);
```

But we dont have such validation in `do_stake`, so basically we would allow here doing staking with below min stake.

Edit: I also found the same min check in `migrate`

pallets/gigahdx-rewards/src/voting_hooks.rs
```diff
@@ -0,0 +1,215 @@
...
+					}
+				}
+			});
+		}
+
+		if !matches!(status, Status::Completed) {
+			return;
+		}
+
+		let _ = maybe_allocate_and_record::<T>(who, ref_index, &record);
```
dmoka: here we silently swallow the error.

At least we could add a debug assert and log here so we ensure it never happens?

pallets/gigahdx-rewards/src/voting_hooks.rs
enthusiastmartin: this will be done at the very end

pallets/gigahdx-rewards/src/lib.rs
enthusiastmartin: By design. do_stake intentionally has no admission control (compounding already-earned rewards), so we don't enforce MinStake here. giga_stake/migrate keep the floor.

pallets/gigahdx-rewards/src/voting_hooks.rs
enthusiastmartin: Done — both sites now debug_assert! + log::error! on the Err arm instead of swallowing.

PR #1456, opened 2026-05-14
fix: omnipool max slip fee cap

enthusiastmartin:
Fixed the missing max cap on slip fees.


math/src/omnipool/math.rs
```diff
@@ -218,7 +218,7 @@ pub fn calculate_buy_for_hub_asset_state_changes(
 
 	// Invert buy-side slip to find how much hub asset the user must provide
 	let slip_buy_amount = if let Some(slip) = slip {
-		let d_gross = invert_buy_side_slip(d_net, slip.asset_hub_reserve, slip.asset_delta)?;
+		let d_gross = invert_buy_side_slip(d_net, slip.asset_hub_reserve, slip.asset_delta, slip.max_slip_fee)?;
```
dmoka, line 221: I think there is a missing assertion in our corresponding tests.

If I hardcode-pass either `Permill::one()` or `Permill::from_parts(2_000_000)` here instead of `slip.max_slip_fee`, no test fails.

math/src/omnipool/slip_fee.rs
dmoka: any reason we dont use `PERMILL_SCALE` constant here?

dmoka: any reason we dont se the `PERMILL_SCALE` constant here?

math/src/omnipool/tests.rs
```diff
@@ -3350,3 +3350,747 @@ fn cross_validate_buy_dot_with_lrna_with_prior_delta() {
...
+		hub_reserve: 100_000_000 * UNIT,
+		shares: 100_000_000 * UNIT,
+		protocol_shares: 0u128,
+	};
+	let slip = TradeSlipFees {
+		asset_in_hub_reserve: asset_in_state.hub_reserve,
+		asset_in_delta: SignedBalance::zero(),
+		asset_out_hub_reserve: asset_out_state.hub_reserve,
+		asset_out_delta: SignedBalance::zero(),
+		max_slip_fee: Permill::from_percent(1),
```
dmoka, line 3421: we could add some tests with `max_slip_fee` = `Permill::zero()` and `Permill::from_percent(99)`, so we see there is no unexpected behaviours.

math/src/omnipool/slip_fee.rs
dmoka: also, i would use checked math for consistency.

dmoka (approved):

PR #1460, opened 2026-05-20
feat: Gigahdx liquidations

enthusiastmartin:


pallets/gigahdx-rewards/src/traits.rs
```diff
@@ -22,3 +23,24 @@ pub trait TrackRewardTable<TrackId> {
...
+
+/// Conviction-voting interactions used by `pallet-liquidation` during a
+/// gigahdx liquidation. `force_release_vote_lock` exists because
+/// `try_remove_vote` does not resync the `pyconvot` lock — leaving the
+/// borrower's transfer freeze stale on the seize path.
+pub trait ClearConflictingVotes<AccountId> {
+	fn clear_conflicting_votes(who: &AccountId, max_remaining_hdx: Balance) -> Result<u32, DispatchError>;
+
+	/// Saturating reduction of the `pyconvot` lock by `amount`.
+	fn force_release_vote_lock(who: &AccountId, amount: Balance) -> Result<(), DispatchError>;
```
dmoka: i dont see this used in any prod code, is it intentional or we have some missing usage?

pallets/liquidation/src/lib.rs
```diff
@@ -234,9 +276,16 @@ pub mod pallet {
 		/// Emits `Liquidated` event when successful.
 		///
 		#[pallet::call_index(0)]
+		// Worst case is the gigahdx branch: up to 3 full-gas EVM calls
```
dmoka: not sure if this is the best way to handle the benchmark of worst case.

In the current liqudate benchmark test, we use collateral asset hdx.

I see 2 options:
- either we use gigahdx as collateral asset so we properly have the worst case here
- or we introduce a new benchmark test for gigahdx branch, then in the weight macro we switch to it if collateral asset is gigahdx.

PR #1464, opened 2026-05-25
Remove Helikon boot node

kukabi:
This PR removes the inactive Helikon boot node from the chainspec file at `/node/res/hydradx.json`.

Thanks.


enthusiastmartin:
thanks. removed as part of #1479

PR #1482, opened 2026-07-13
feat: improve stableswap pool share mint/burn process

enthusiastmartin:
This PR reworks stable pool shares mint and burn process. 

instead of relying on total issuance - stableswap tracks issuance of each pool separately.

The goal of this rework is to harden security of the stableswap implementation, despite the need of extra storage.


pallets/stableswap/src/lib.rs
```diff
@@ -1922,6 +1929,32 @@ impl<T: Config> Pallet<T> {
...
+		ShareIssuance::<T>::try_mutate(pool_id, |issuance| -> DispatchResult {
+			*issuance = issuance
+				.checked_sub(amount)
+				.ok_or(Error::<T>::InsufficientShareIssuance)?;
+			Ok(())
+		})?;
+		T::Currency::withdraw(pool_id, who, amount, ExistenceRequirement::AllowDeath)
+	}
+
+	fn debug_assert_issuance_in_sync(pool_id: T::AssetId) {
```
cl0w5: This will still trigger as a false positive (?) during OTC settlement:

1. OTC temporarily mints M LP tokens.
2. The token system’s total issuance increases by M.
3. StableSwap’s new internal ShareIssuance intentionally does not increase.
4. OTC routes the received assets through StableSwap to buy LP tokens.
5. StableSwap checks this debug assertion which fails because: total_issuance = ShareIssuance + M

Only later would OTC burn the temporary M tokens and bring both counters back into sync. The assertion runs before that cleanup can happen.

The panic from the debug assert will also prevent an end-to-end regression test

cl0w5 (commented):
Can you add an end-to-end regression / integration test covering the whole flow?

placeOrder → settleOtcOrder → router → StableSwap → OTC cleanup

pallets/stableswap/src/lib.rs
enthusiastmartin: i have prepare a change in the settlemtn to not use mint at all.

cl0w5 (approved):

pallets/stableswap/src/lib.rs
mrq1911: mby this should be real assert, not just debug

enthusiastmartin: done.

Issue #1495, opened 2026-07-20
Feature request: add an EVM `batch` precompile at 0x808 (one Call-Permit → a batch of sponsored EVM calls)


acyberduck:
## Summary
Proposes adding a standard EVM `batch` precompile at address `0x808`, of the same
lineage as Hydration's existing Call-Permit precompile (`0x080a`) — the one we already
use for sponsored, gasless transactions. It would let a single signed Call-Permit
execute a batch of EVM calls inside one frame, so a multi-step EVM flow can run gaslessly
for the user in a single sponsored transaction. No changes to any fee, signing, or
paymaster code.

**To be explicit: this is _not_ a flash-loan proposal** — no flash-borrowing, no receiver
contract, no credit delegation. It's simply a batch of the user's _own_ calls, run in one
sponsored frame.

## Motivation
We run a HOLLAR leverage vault on Hydration and use the signed `dispatch_permit` flow
(the Call-Permit precompile at `0x080a`) for gasless, sponsored user transactions — it
works great for **one** EVM call per permit.

Our vault's core action is a short loop (supply HOLLAR → borrow → supply → …). We'd like
the whole loop to execute in **one** sponsored transaction for a zero-balance user
account. Today that's the one thing we can't express:

- Sponsorship itself is **not** the blocker — the `EvmFeePayer` override already spans
  nested frames, and we verified (read-only mainnet dry-runs) that the **entire loop
  executes and commits atomically in one transaction** when run non-re-entrantly.
- The blocker is **EVM re-entrancy**: batching the loop's EVM calls at the substrate
  layer — `dispatch_permit(to=0x401 dispatch precompile, data=utility.batchAll([
  dispatcher.dispatchEvmCall(evm.call{source=user}), … ]))` — opens a fresh
  `pallet_evm::call` from *inside* the outer permit's EVM frame. That fails at the runner
  level (`EvmPermitRunnerError`, zero events), the expected result of starting a new EVM
  execution inside an executing one.

A `batch` precompile sidesteps this: it runs each subcall as an **internal EVM `CALL`**
(preserving `msg.sender = user`) inside the single sponsored frame — no re-entrancy,
`batchAll` atomicity, and the existing sponsorship covers the whole batch end-to-end.

## Proposed change
Register the standard `pallet-evm-precompile-batch` at the free address `0x808`, exactly
as the Call-Permit precompile is registered:

[...]

PR #1505, opened 2026-07-31
feat(meta-tx): sponsored meta transactions for native Substrate accounts [ignore benchmarks]

iamyxsh:
## What this is

A POC of **sponsored meta transactions for native Substrate accounts** — a user signs an intent
off-chain, anyone submits it and pays the fee, and the call executes under the *signer's* origin.

This is the Substrate counterpart of the EVM call-permit path. Opening it for feedback on
direction before hardening.

## Why only Substrate accounts

While scoping this I tested what already exists. **The EVM side already does most of what was
asked for:**

- A signer with 0 HDX / 0 WETH / 0 DAI is already fully sponsored end to end — the paymaster
  pays everything (measured: ~0.09 HDX).
- Batching already works under one signature — verified at 14 and at 50 legs.

So the remaining gap was native Substrate accounts only, and that is all this PR adds.

## Why it is small

On the EVM side "the relayer pays" needed the whole `EvmFeePayer` override, because EVM gas is
charged to `source`. On the Substrate side the relayer signs and submits the extrinsic, so
`ChargeTransactionPayment` already charges them. Nothing to build.

That leaves three jobs: verify a signature, dispatch under the signer's origin, bump a nonce.

## What is in the payload

`blake2_256` over `(tag, call, signer, nonce, deadline, genesis_hash, spec_version)`.

Genesis hash and spec version are the Substrate analogue of EIP-712's domain separator — without
them an intent replays across forks, and across runtime upgrades that change call encoding.

## Behaviour worth reviewing

- **The nonce is consumed even when the inner call fails.** A rejected intent must not stay
  replayable. The inner outcome is reported in the `Dispatched` event rather than as an
  extrinsic error. This mirrors the EVM permit path.
- **The runtime `CallFilter` applies** — verified by an integration test that pauses
  `Currencies::transfer` and confirms a signed intent for it does not execute.
- **Batching falls out for free** — `utility.batch_all` is just a valid inner call.

## The relayer

`scripts/meta-tx-relayer/` — an HTTP service that accepts intents, screens them against a
pallet allowlist, dry-runs, and submits. ERC-4337's bundler-plus-paymaster.

**It is deliberately not in the node.** The relayer holds a funded key and signs. RPC nodes are
public and replicated, so a spending key in one means every operator sponsors everyone. The
repo's own precedent agrees: `liquidation-worker-support` exposes an RPC that only serves data,
and the actor that signs lives outside. A node RPC remains possible if sponsorship ever needs
to be consensus-adjacent; it does not.

`npm run demo` generates a brand-new zero-balance account, signs an intent, and shows the
balance still zero afterwards.

## Tests

22 total, all passing.

- 16 unit (`pallets/meta-tx/src/tests.rs`): valid path and origin, wrong signer, call swapped
  after signing, deadline swapped after signing, replay, nonce ahead, expired, on-deadline
  boundary, nonce consumed on inner failure, `batch_all` all-or-nothing both ways, unsigned
  origin rejected, every payload field proven load-bearing, per-signer nonce independence.
- 6 integration (`integration-tests/src/meta_tx.rs`), through the **full fee pipeline** so the
  sponsorship claim is real: zero-balance signer fully sponsored, batching, replay, expiry,
  call filter respected, meta nonce independent of the system nonce.

## Known gaps — deliberate for a POC

- **Weights are a placeholder.** `weights.rs` is hand-set and needs `scripts/benchmarking.sh`
  before this carries value on mainnet. The benchmark currently measures the reject path.
- **Wallet UX.** There is no EIP-712 equivalent on Substrate, so a user signs an opaque 32-byte
  hash. Fine for a relayer-operated flow, not fine for arbitrary users. This may be the real
  blocker for production and is worth discussing.
- **No deferred execution** ("sign now, fire later"). Separate feature with its own storage,
  expiry and ordering questions. Not built pending confirmation it is wanted.
- **No call scoping beyond the existing `CallFilter`.** A signed call is a broad grant. A
  `ProxyType`-style filter is the obvious next step if that is a concern.
- **No revocation.** Bumping one's own meta-nonce would do it; cheap to add now, painful later.
- **Relayer is minimal** — in-memory rate limits, key from env, no auth. See its README.

## Questions

1. Is the Substrate gap the right target, given the EVM side already works?
2. Deferred execution — wanted, or out of scope?
3. Is opaque-hash signing acceptable, or does this need a readable-payload story first?

### DIFF
--- a/pallets/meta-tx/src/lib.rs
+++ b/pallets/meta-tx/src/lib.rs
@@ -175,7 +175,7 @@ pub mod pallet {
 				return Err(Self::rejected(Error::<T>::InvalidSignature, base_weight));
 			}
 
-			Self::execute(relayer, signer, call, nonce, next_nonce, base_weight)
+			Self::execute(relayer, signer, *call, nonce, next_nonce, base_weight)
 		}
 
 		/// Execute `call` under the account of an EVM address that authorised it off-chain,
@@ -200,6 +200,7 @@ pub mod pallet {
 			let call_weight = call.get_dispatch_info().call_weight;
 			T::WeightInfo::dispatch_evm_meta_tx().saturating_add(call_weight)
 		})]
+		#[allow(clippy::too_many_arguments)]
 		pub fn dispatch_evm_meta_tx(
 			origin: OriginFor<T>,
 			from: H160,
@@ -228,7 +229,7 @@ pub mod pallet {
 				return Err(Self::rejected(Error::<T>::InvalidEvmSignature, base_weight));
 			}
 
-			Self::execute(relayer, signer, call, nonce, next_nonce, base_weight)
+			Self::execute(relayer, signer, *call, nonce, next_nonce, base_weight)
 		}
 	}
 
@@ -263,7 +264,7 @@ pub mod pallet {
 		fn execute(
 			relayer: T::AccountId,
 			signer: T::AccountId,
-			call: Box<<T as Config>::RuntimeCall>,
+			call: <T as Config>::RuntimeCall,
 			nonce: u32,
 			next_nonce: u32,
 			weight: Weight,
--- a/pallets/meta-tx/src/lib.rs
+++ b/pallets/meta-tx/src/lib.rs
@@ -68,7 +68,7 @@ pub mod pallet {
 	use hydradx_traits::evm::{EvmFeePayerSupport, InspectEvmAccounts};
 	use sp_core::{H160, H256};
 	use sp_io::hashing::{blake2_256, keccak_256};
-	use sp_runtime::traits::{Dispatchable, IdentifyAccount, UniqueSaturatedInto, Verify, Zero};
+	use sp_runtime::traits::{Dispatchable, IdentifyAccount, Saturating, UniqueSaturatedInto, Verify, Zero};
 	use sp_std::boxed::Box;
 	use sp_std::vec::Vec;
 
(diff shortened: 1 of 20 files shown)


mrq1911 (changes requested):
make sure it works for evm calls and the inner permits (or forbid them)

mrq1911:
fixes #1495

pallets/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+			let relayer = ensure_signed(origin)?;
+
+			ensure!(
+				frame_system::Pallet::<T>::block_number() <= deadline,
+				Error::<T>::Expired
+			);
+			ensure!(Nonces::<T>::get(&signer) == nonce, Error::<T>::InvalidNonce);
+
+			let payload = Self::signing_payload(&call, &signer, nonce, deadline);
+			ensure!(signature.verify(&payload[..], &signer), Error::<T>::InvalidSignature);
```
enthusiastmartin: note that this early exit charges the relayer for the whole weight. 

if i keep sending permits that are heavy, with invalid signature - it just keeps paying the whole fee

pallets/meta-tx/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+
+			Nonces::<T>::insert(&signer, nonce.saturating_add(1));
```
enthusiastmartin: well - this is no good. 
far fetched as you'd need to send u32 txs, but in the end - you could replay the last tx easily. 

this silent saturation cause sthis 
1. it passes the ensure nonce == current nonce ( above )
2. sets the same value saturated here. 

so i see two issues, albeit theoretical. if you sign a permit with max nonce and max deadline - it can be replayed forever.

enthusiastmartin (changes requested):

enthusiastmartin:
no longer relevant

PR #1509, opened 2026-08-07
feat: increase pause max length

enthusiastmartin:


pallets/transaction-pause/src/mock.rs
```diff
@@ -42,6 +42,32 @@ mod transaction_pause {
 	pub use super::super::*;
 }
 
+/// Provides a call whose name is longer than any in the other mocked pallets, so that pausing can
+/// be exercised against long names.
+#[frame_support::pallet]
+pub mod long_named_pallet {
```
cl0w5, line 48: Why is this necessary? Can you provide some context?

cl0w5 (approved):

pallets/transaction-pause/src/mock.rs
enthusiastmartin: jsut mocked pallet for testing

PR #1510, opened 2026-08-07
feat: correct handling of deposit limits 

enthusiastmartin:
improve handling of deposit limits in certain situations

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10529,7 +10529,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-route-executor"
-version = "2.12.1"
+version = "2.12.2"
 dependencies = [
  "frame-benchmarking",
  "frame-support",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9801,7 +9801,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-hsm"
-version = "1.7.0"
+version = "1.7.1"
 dependencies = [
  "ethabi",
  "evm",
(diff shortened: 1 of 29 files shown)


cl0w5 (commented):
Mark sell context before transferring into the router — runtime/hydradx/src/assets.rs:618

InTradeContext now depends on Broadcast::get_swapper(). However, route_executor::do_sell transfers asset_in into the whitelisted router at pallets/route-executor/src/lib.rs:472 and only sets the swapper at line 481.

Consequently, the deposit-limit hook treats that initial transfer as outside a trade and may reserve funds on the router instead of returning DepositLimitExceededForWhitelistedAccount. If execution subsequently succeeds using the remaining reducible balance, the reserved input stays trapped in the router.

Set the swapper before the transfer, as the buy path already does, and add a sell test where the input transfer triggers deposit locking. The added integration test only covers deposits of the output asset after the context has been set.

cl0w5 (approved):
lgtm

enthusiastmartin:
> Mark sell context before transferring into the router — runtime/hydradx/src/assets.rs:618
> 
> InTradeContext now depends on Broadcast::get_swapper(). However, route_executor::do_sell transfers asset_in into the whitelisted router at pallets/route-executor/src/lib.rs:472 and only sets the swapper at line 481.
> 
> Consequently, the deposit-limit hook treats that initial transfer as outside a trade and may reserve funds on the router instead of returning DepositLimitExceededForWhitelistedAccount. If execution subsequently succeeds using the remaining reducible balance, the reserved input stays trapped in the router.
> 
> Set the swapper before the transfer, as the buy path already does, and add a sell test where the input transfer triggers deposit locking. The added integration test only covers deposits of the output asset after the context has been set.

corrected.

PR #1513, opened 2026-08-10
feat: buy dca no longer allowed

enthusiastmartin:
Do not allow creating buy DCAs.

Existing buys keeps executing until completed.

This is first step to completely remove buy possibilities.


pallets/dca/src/lib.rs
```diff
@@ -477,6 +479,10 @@ pub mod pallet {
 		/// Creates a new DCA (Dollar-Cost Averaging) schedule and plans the next execution
 		/// for the specified block.
 		///
+		/// Only `Order::Sell` schedules can be created. Buy orders are rejected with
+		/// `NoLongerSupported`; buy schedules already stored keep executing until they complete
+		/// or are terminated.
```
cl0w5, line 484: As we still support running buy execution i would prefer to keep the buy tests (you can inject them in storage for mock)

cl0w5 (approved):
lgtm

PR #1528, opened 2026-09-17
feat(coretime-alert): optional discord mention and --fake-alert drill

mrq1911:
## why

Alerts currently post an embed with no top-level `content`. **Discord does not send a notification for mentions inside an embed** — only message `content` does. So today a coretime renewal alert lands silently in the channel and is only seen by whoever happens to scroll past it. The renewal right is genuinely lost at the region deadline, so a silent alert is close to no alert.

There was also no way to see what a real alert looks like without waiting for an actual deadline — `--test` only proves the webhook is reachable.

## what

- **`ALERT_MENTION`** (new, optional env var) — posted as top-level `content` above the embed, so alerts actually ping. Takes `@here`, `<@&ROLE_ID>` or `<@USER_ID>`. Sets `allowed_mentions` explicitly rather than relying on the webhook default, so a role not flagged "mentionable" still pings. **Unset, the payload is byte-identical to today's.**
- **`--fake-alert`** (new flag) — posts one synthetic alert per configured chain (Hydration `URGENT`, Basilisk `WARNING`) through the real `buildEmbed` path. Each is titled `🧪 [DRILL]` with a "this is a test alert" line and a `synthetic drill` footer so it can't be mistaken for a live deadline. Contacts no RPC and does **not** touch the state file, so it cannot disturb the cooldown of a real standing alert.
- README: config-table row, the embed-doesn't-ping gotcha, ID formats, drill docs.
- `lark-stack.yml`: `ALERT_MENTION` wired up.

Both chains were already watched — Hydration (task 2034, Polkadot) and Basilisk (task 2090, Kusama); no change to coverage.

## testing

Dry-run verified with and without the var — unset produces exactly the current payload; set produces:

```json
"content": "<@...>",
"allowed_mentions": { "parse": ["everyone","roles","users"] }
```

Not yet run against the live webhook — that happens at deploy.

## deploy notes

Needs an image rebuild; `autoredeploy` on the lark `coretime-alert` stack picks up `:latest`, and `ALERT_MENTION` must be set on the service.

Unrelated, but worth flagging: `DISCORD_WEBHOOK_URL` is stored in plaintext in the stack env, so it surfaces in any Swarmpit service dump. The script already supports `DISCORD_WEBHOOK_URL_FILE` if we want to move it to a Swarm secret.


mrq1911:
drilled and deployed. `--fake-alert` posted both DRILL embeds through the live webhook with top-level `content` set, and the ping landed — so the mention path works end to end. lark now runs `sha256:31e5f36e` with `ALERT_MENTION` on the stack, re-confirmed from inside the running container.

basilisk was sitting at URGENT when I started (0/3 secured, 5.5d to region begin) and got renewed mid-deploy, so both chains are 3/3 now and no real alert is pending.

enthusiastmartin (approved):

PR #1535, opened 2026-09-21
feat: ICE - route provider without onchain routes

enthusiastmartin:
this pr introduces route provider that are determined dynamically, without relying on stored on chain routes.

note that this should be used carefully, as the computation can be heavy.

currently needed for ICE solution resolution.


cl0w5 (approved):
lgtm

